PR #120, opened 2015-01-09
Added Icarrere suggestion

EricOuellet2:
Added Icarrere suggestion: Call this.GetDimensions before renderer.ScaleTransform.


tebjan:
looks good, thank you!

PR #121, opened 2015-01-10
Parse Exponent stop-opacity

Stardraw:
Stop crash on null or empty images
Deal with multiple spaces in Matrix


tebjan:
good fix, thanks!

PR #122, opened 2015-01-14
Fix crash in GDI when empty rectangle

Stardraw:


tebjan:
looking good, thanks!

PR #128, opened 2015-02-11
Working on making this export valid SVG documents.

JoshMcCullough:
I noticed a lot of issues regarding the writing of SVG documents back out to XML. I loaded the SVG in the gist below and was using it to test. I fixed SOME of the issues but there are others for which I will create an issue.

Specifically this pull request fixes...
 - Some attributes were not camelCase - that should be fixed now.
 - Added docs to many enums.
 - Stopped output of null/empty attributes.
 - Fixed default values for some enums.
 - Simplified enum/default logic.
 - Marked some enums as flags.
 - Simplified element writing.
 - Allowed an element to determine whether or not it should be written (e.g. don't write empty text elements).

https://gist.github.com/JoshMcCullough/e01ab100147eba629875


tebjan:
hello james, thanks a lot for this big pull request. I have to find some time to look thru it. but have you tested it with the SvgW3CTestRunner app?

JoshMcCullough:
Who's James? :) I haven't tested it with that. In fact, I forgot that I ever did this even though it was recent! If I get some time, I can try to test it as you suggest.

tebjan:
just merged it and the test 'styling-css-04-f.svg' fails on serialization. please have a look with it with the W3CTestRunner application in the Svg solution. thanks!

tebjan:
also the two text-path test fail: 'text-path-01-b.svg' and 'text-path-02-b.svg'

tebjan:
reverting the merge now and waiting for the fixes.

JoshMcCullough:
Yeah, I did some testing last week and will make fixes where appropriate.

tebjan:
great, then just open a new pull request then. thanks!

Issue #124, opened 2015-01-21
ViewBox with negative values.


arthek89:
Hello, I do not know whether I understand SVG drawing and transformations well enough, but when I open a SVG file which has negative values (minX,minY) in its ViewBox, the image is not displayed properly - it looks as if it was moved in upper left direction ( I can see a quarter of the image).

At the same time the SVG file which is very small is not sharp enough.

Is it my fault or the library one ? 

I would be grateful for your piece of advice.

tebjan:
should be fixed now

PR #133, opened 2015-02-20
Scaling minx and miny of viewbox

kataklinger:
Should resolve issue #124


tebjan:
thanks a lot, will check it out an merge.

PR #140, opened 2015-03-23
Revert "Working on making this export valid SVG documents."

tebjan:
Reverts vvvv/SVG#128


JoshMcCullough:
Was there something wrong with my original pull request which caused it to be reverted?

JoshMcCullough:
NM I see the comments. Will investigate.

PR #147, opened 2015-04-03
Fixed attributes/enums.

JoshMcCullough:
I noticed a lot of issues regarding the writing of SVG documents back out to XML. I loaded the SVG in the gist below and was using it to test. I fixed SOME of the issues but there are others for which I will create an issue.

Specifically this pull request fixes...

- Some attributes were not camelCase - that should be fixed now.
- Added docs to many enums.
- Stopped output of null/empty attributes.
- Fixed default values for some enums.
- Simplified enum/default logic.
- Marked some enums as flags.
- Simplified element writing.
- Allowed an element to determine whether or not it should be written (e.g. don't write empty text elements).

https://gist.github.com/JoshMcCullough/e01ab100147eba629875


JoshMcCullough:
Note that this is 1 part of the larger pull request I sent before, but I fixed the issue with those tests you mentioned.

PR #148, opened 2015-04-03
Simplified element writing.

JoshMcCullough:
Simplified element writing (moved WriteElement(XmlTextWriter) out to SvgDocument as Write(XmlTextWriter); allowed elements to decide whether or not they should be written. E.g. text elements should not be written unless they have children.


JoshMcCullough:
This is another from the larger pull request.

tebjan:
hi josh, just to let you know, saw all your pull requests. work overload, coming back to it in a while.

JoshMcCullough:
No problem.

Issue #143, opened 2015-03-25
Clip-path is copied from parent to child which can hide elements if transformed


mkb137:
The clip-path is copied from parent to child.  This can break the SVG when there's transformations on the elements.  Compare this SVG in a normal SVG viewer and in the project's SVG viewer:

``` xml
<svg xmlns:xlink="http://www.w3.org/1999/xlink" height="400" width="1240" xmlns="http://www.w3.org/2000/svg" style="font-family:'lucida grande', 'lucida sans unicode', arial, helvetica, sans-serif;font-size:12px;" version="1.1">
    <defs>
        <clipPath id="highcharts-41">
            <rect height="241" width="1032" y="45" x="78"/>
        </clipPath>
    </defs>
    <rect class=" highcharts-background" fill="#fbfc92" strokeWidth="0" height="400" width="1240" y="0" x="0"/>
    <g clip-path="url(#highcharts-41)"  class="highcharts-annotations-group-0">
        <g transform="translate(719.1574467936198,67.23511904761904)">
            <g style="" transform="translate(0,0)">
                <text transform="translate(0,15)"  x="3">test</text>
            </g>
        </g>
    </g>
</svg>
```

I've implemented a fix in a fork: https://github.com/mkb137/SVG/commit/9a1fb1c54fb744a6e79942be9192e408fde5ee0e

mrbean-bremen:
Fixed with #150.

Issue #142, opened 2015-03-25
HSL color support


mkb137:
The library doesn't support hsl colors (e.g. <path stroke="hsl( 100, 10%, 60%)"/>).  I added an implementation in a fork, here: https://github.com/mkb137/SVG/commit/83bb847c17096fa5b7ab3228e2e428b3ddb161f5

tebjan:
thanks a lot, merged now!

PR #150, opened 2015-04-06
HSL support, fix for clip path (issues 142 and 143)

mkb137:
Fixes for issue #142 and issue #143.  This may also fix another user's issue #123.
The first isse adds support for HSL color (e.g. fill="hsl( 100, 10%, 20% )").  The second fixes how SVG would copy clip-paths from parent to child even when the child had other transforms making the clip path invalid.  There's no need for it that I can see and it has the effect of corrupting loaded SVG.


tebjan:
great, thanks a lot. currently work overload, coming back to it in a while.

dylanklebold:
Wow, thanks

PR #153, opened 2015-04-30
Fix for text-anchor on a text node containing tspans

Ellisthion:
There's currently a problem where the text-anchor attribute / style is ignored if the text element contains tspans.

    <text text-anchor="middle">
        <tspan>A</tspan>
        <tspan>B</tspan>
    <text>

This will render as if it doesn't have the text-anchor.

My pull request adds a step to SetPath which translates the paths of the tspans after they've been set (you need to wait 'til afterwards because you need to know the text bounds).

I had an issue where doing this left behind a black piece of text in the original position, so I've got a bit of a hack to solve this by wiping the text's path.

I'm not sure this is really the right solution to the problem but here it is anyway.

### DIFF
--- a/Source/Text/SvgTextBase.cs
+++ b/Source/Text/SvgTextBase.cs
@@ -399,9 +399,41 @@ namespace Svg
                 }
             }
 
-
             _path = path;
             this.IsPathDirty = false;
+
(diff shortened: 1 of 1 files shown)


Ellisthion:
This works but there's some issues. I need to think about it some more.

Issue #170, opened 2015-07-27
To draw image with given raster size


ubbn:
There is a need for a method Draw which can take raster image size (widh, height) as parameter. So image can be rendered in any given size rather than being ruled by svg size attributes, although it is still possible and okay.

Issue #169, opened 2015-07-27
SVGs are not scaleable.


Sonorpearl:
HeyHo guys,

I´m currently running in a problem, where my SVG I use in several projects,
are not scaled to full size/the size I set the SvgDoc/PictureBox.

Here is my code:
```
    Private Sub Button1_Click(sender As System.Object, e As System.EventArgs) Handles Button1.Click
        If (OpenFileDialog1.ShowDialog() = Windows.Forms.DialogResult.OK) Then
            svgDoc = SvgDocument.Open(OpenFileDialog1.FileName)

            'svgDoc.Transforms = New SvgTransformCollection()
            'svgDoc.Transforms.Add(New SvgScale(2, 2))
            svgDoc.X = 0
            svgDoc.Y = 0
            'svgDoc.Width = 250
            'svgDoc.Height = 250

            'svgDoc.AspectRatio = New SvgAspectRatio(SvgPreserveAspectRatio.xMaxYMax)

            svgDoc.Width = New SvgUnit(SvgUnitType.Percentage, 100)
            svgDoc.Height = New SvgUnit(SvgUnitType.Percentage, 100)
            'svgDoc.AspectRatio.Align = SvgPreserveAspectRatio.xMaxYMax
            Dim bm As Bitmap = New Bitmap(250, 250)
            PictureBox1.Image = svgDoc.Draw()
        End If
    End Sub
```

ubbn:
Similar issue faced with me. I opened new issue #170 . I hope I can pull new commit soon to fix it.

Sonorpearl:
@ubbn How far have you come? I see that you did quite a lot. I already checked it out. ^^

Do you know where the Problem lies?
I tried to follow up on the Width Settings within the Programm and found this out:
Somehow the Image gets a size of 151/171 (height/width).
The Scale Factor gets set to 2.4. Thats is correct for the width/height.

But the SVG Rectangle and other things in the SVG are just 60x60.
And the Rectangle and other SVG Properties now gets multiplied by 2.4.
60x2,4 = 142
And the actuall SVG Bitmap Image is 142x142 which is not correct. 
It took the Scale factor of 2.4 because of the Properties from the SVG: width="171.082px"  height="151.441px"
But it should have taken a Scale Factor from the rect/images and all other things in the SVG. 
Which would be 400/60 = 6.7.

What I believe woud be right:
[...]

PR #171, opened 2015-07-27
Added new Draw method

ubbn:
Added new Draw() method that can take raster size and render image in that given size. 
It fixes issues #170 and #169

### DIFF
--- a/Source/Basic Shapes/SvgCircle.cs
+++ b/Source/Basic Shapes/SvgCircle.cs
@@ -100,15 +100,24 @@ namespace Svg
         /// </summary>
         public override GraphicsPath Path(ISvgRenderer renderer)
         {
-            if (this._path == null || this.IsPathDirty)
+            if ((this._path == null || this.IsPathDirty)	&& base.StrokeWidth > 0)
             {
+							float halfStrokeWidth = base.StrokeWidth / 2;
+
+							// If it is to render, don't need to consider stroke width.
+							// i.e stroke width only to be considered when calculating boundary
+							if (renderer != null)
+							{
+								halfStrokeWidth = 0;
+								this.IsPathDirty = false;
+							}
+
                 _path = new GraphicsPath();
                 _path.StartFigure();
-                var center = this.Center.ToDeviceValue(renderer, this);
-                var radius = this.Radius.ToDeviceValue(renderer, UnitRenderingType.Other, this);
-                _path.AddEllipse(center.X - radius, center.Y - radius, 2 * radius, 2 * radius);
+								var center = this.Center.ToDeviceValue(renderer, this);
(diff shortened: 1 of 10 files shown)


Sonorpearl:
HeyHo Mate,

I just included your changes into my Local SVG Version.
And tried: "PictureBox1.Image = svgDoc.Draw(250, 250)"

And sadly nothing has changed. The Picture is still just 60x60, like pre-defined in the SVG and the Predefined Offset within the SVG is still in place.

ubbn:
Hey,

I tested your svg. It works fine by calling svgDoc.Draw(250, 250) which generates attached image in that size. One more thing, your image has colored rectangle size of 60x60. But your whole image is bigger than that. I am afraid you may confuse your whole image with colored rectangle. 


output

Sonorpearl:
Hey ubbn,

the thing is that I have hundreds of SVGs like this. And many SVGs on the Net are made in the same way. And the SVGs work in Environments like Adobe Air or HTML perfectly fine. They do get scaled there to 100% without any adjustments, just passing one width on the container and its done.
SVGs are always scaleable.

ubbn:
One thing I forgot to mention is that this new method Draw(width, height) can calculate scaled size even if you provide one of height or width  as 0. 

About your concern, those tools which you are using presumably do same thing (set height and width). So you can do simple implementation to Resize event of your container to re-render image with new size of your container using the new Draw method.

Sonorpearl:
Ahh okay, thats good to hear.

Do you have any idea on how to solve the Problem, I am facing with the SVG?
I mean how does it break up this Library but its easily done in HTML or Adobe Air?

And one suggestion I want to include is, that the Predefined X/Y of the SVG should be overrided with the svgDoc.X and Y Values. The same way you did with the Width. ^^ That would be very nice.

Thank you for your help and your time, so far.

ubbn:
Hi, I have an idea about how to solve this issue. Probably need to add new issue. But it will take some time to debug & test. 

About your suggestion, I guess once your first problem is fixed, it may not really needed. But I will consider. But I can't promise anything and date.

tebjan:
@ubbn thanks for your commits! have you tested your changes against the svg tester in the Svg.sln?
so that all passing test are still passing?

ubbn:
@tebjan I have tested my all commits on graphics of SVG3WCTestRunner as well as on my own graphics. If you provide instruction to do other detail tests, please let me know.

But about this fix to #171, it overwrites the original size of graphic when no size is given (which fulfills my requirement), but not the size given in < svg > tag which may needed for some other cases. For example in case of graphics of @Sonorpearl. I mean this my commit may not needed to generalize.

tebjan:
good, I'll merge it and check it out...

Issue #59, opened 2014-04-07
SvgDocument.Open fails returning null sometimes when used by multiple threads


kimmccoy:
I am using Svg in a Asp.net Mvc app. Png images are generated via this library from SVG markup. It's a web app, so multiple threads hit the image generation code at the same time.
Sometimes, SvgDocument.Open will return a null even though the svg is valid. Other times it's fine. So not thread safe is my guess.

The problem is in SvgElementFactory.SetPropertyValue where more than one thread may attempt to set the same property value at the same time, attempting to add to _propertyDescriptors just after another thread has already added the dictionary key.

Easy fix is to add [MethodImpl(MethodImplOptions.Synchronized)] on the method.

SvgElementFactory.AvailableElements is also prone to this some problem.

Note: I don't have time to submit a nicely tested fix. My hack is to lock in my own app to avoid the issue. Hope this saves someone time. I will submit a patch if the lock becomes a performance issue for me.

Threetwosevensixseven:
+1
Same problem here, in a MVC app.

My workaround is:

```
private static object _lock = new object();
...
SvgDocument doc = null;
lock (_lock)
{
     doc = SvgDocument.Open(file);
}
```

Threetwosevensixseven:
I logged the first chance exceptions, and there's a whole bunch of stuff getting wadded up, culminating in the NullReferenceException:

```
System.NotSupportedException: Must call AddTerm AddSeparator in that order
   at ExCSS.TermList.AddTerm(Term term) in c:\Dev\vvvv\SVG\Source\External\ExCSS\Model\Values\TermList.cs:line 35
===========

System.NotSupportedException: Must call AddTerm AddSeparator in that order
   at ExCSS.TermList.AddTerm(Term term) in c:\Dev\vvvv\SVG\Source\External\ExCSS\Model\Values\TermList.cs:line 35
===========

System.InvalidOperationException: The CreateDocument method can only be used to parse root <svg> elements.
   at Svg.SvgElementFactory.CreateDocument[T](XmlReader reader) in c:\Dev\vvvv\SVG\Source\SvgElementFactory.cs:line 59
===========

System.InvalidOperationException: The CreateDocument method can only be used to parse root <svg> elements.
   at Svg.SvgElementFactory.CreateDocument[T](XmlReader reader) in c:\Dev\vvvv\SVG\Source\SvgElementFactory.cs:line 59
===========

[...]

PR #186, opened 2015-08-26
Fixed problems when running in multiple threads

pulsatrixbv:
Fixed problems when running in multiple threads. The use of a static matrix and the use of StringFormat.GenericTypographic seemed to cause problems when used in multi-threading. This is resolved by making the matrix non static and use a copy of the StringFormat.

Also created a new exception for memory overflows. When using multithreading to render a large amount of big images the bitmap creation can fail (due to memory allocation) throwing an error. This is now caught and a specific exception is thrown.

This will fix issue: #59

### DIFF
--- a/Source/Exceptions/SvgMemoryException.cs
+++ b/Source/Exceptions/SvgMemoryException.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace Svg.Exceptions
+{
+	[Serializable]
+	public class SvgMemoryException : Exception
+	{
+		public SvgMemoryException() {}
+		public SvgMemoryException(string message) : base(message) {}
+		public SvgMemoryException(string message, Exception inner) : base(message, inner) {}
+
+		protected SvgMemoryException(
+			SerializationInfo info,
+			StreamingContext context) : base(info, context) {}
+	}
+}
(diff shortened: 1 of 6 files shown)


abnersouza:
Thanks very much @pulsatrixbv not I'm not having problem anymore.

tebjan:
thanks a lot and good findings!

JohnMcAvinue:
Thanks very much for taking the time to fix this and for sharing it!

JohnMcAvinue:
I'm getting OOM exceptions when generating 5 images. Any idea why this might be?

abnersouza:
@JohnMcAvinue What is the size of the images? Because I did some tests with more than 100 images, .Also @tebjan told that via email:  I ran into memory issues when processing a lot of images at once (20 concurrent drawings of 5000x5000 pixels). In that case you can limit the maximum number of threads (I use parallel foreach so I put a max number of threads in the execution of the parallel).

JohnMcAvinue:
It turns out it's for a specific image which is 1700px X 400px. The chart looks fine though and isn't complex at all. Any ideas what it may be?

pulsatrixbv:
@JohnMcAvinue the OOM exceptions most often occur in the .Draw and/or the allocation of the bitmap and not within the rendering. Possibly your are hitting a part that I have not yet tested.

In my own branch I have a testcase with an example SVG to test the capabilities of a machine, perhaps you can run it on your own machine (to see where the limit is approx.). You might also try to replace the example SVG with your own and test the limits on that one. Some SVG's create a lot of data on the bitmap (esp. when using gradients, so maybe that is the problem) or render a big bitmap to start with.

If you attach an example SVG I can look into it and see if maybe the component is "leaking" on other places.

PR #187, opened 2015-08-30
W3CTestRunner: Added the code to infer the right directory separator for the OS

Emasoft:
W3CTestRunner: Added the code to infer the right directory separator for the OS. Now doesn't crash anymore on Mac OSX.


tebjan:
this PR also includes a change of behavior in error handling, please make two separate PRs from it... thanks!

PR #188, opened 2015-09-04
Fix #185: Multiline text is not rendered. 

### LINKED ISSUE #185: Multiline text is not rendered
Multiline texts in SVG is notated as below according to http://www.w3.org/TR/SVG/text.html#TSpanElement

```xml
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink">
<defs>
<symbol id="Lines" overflow="visible">
  <text x="100"  y="60" style="font-family: Eras Demi ITC; font-size: 50px">
    <tspan x="100" y="160">Line01</tspan>
    <tspan x="100" y="220">Line02</tspan>
  </text>
</symbol>
</defs>
<use xlink:href="#Lines"/>
</svg>
```

But text with this notation can not be rendered.

ubbn:
GetPaths method made recursive call to children of only group element. So, it was not calculating path for tspan element which is nested/child element to text element (non-group). This is why multi-line texts were not rendered. Now GetPaths will call non-group elements' child if there is any. And texts are rendered correctly

### DIFF
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -789,6 +789,10 @@ namespace Svg
         			{
                         var childPath = ((SvgVisualElement)child).Path(renderer);
         				
+      					// Non-group element can have child element which we have to consider. i.e tspan in text element
+      					if (child.Children.Count > 0)
+    				  		childPath.AddPath(GetPaths(child, renderer), false);
+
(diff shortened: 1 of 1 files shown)


tebjan:
thanks, nice little fix!

PR #198, opened 2015-09-18
Fix #176: Blurry text when overlapped with colored background

### LINKED ISSUE #176: Blurry text when overlapped with colored background
When text is overlapped on image, that text is rendered blurry. But on transparent image, rendered normal. See difference in the below images

<b>Image with opaque white background</b>
text_on_image_opaque

<b>Image with transparent background</b>
text_on_image_transparent

ubbn:
CompositingMode specifies how the source colors are combined with the background colors for rendering composite graphics. Default value is SourceOver which blends foreground and background colors. This affects SVG Texts to render blurry when it is overlapped with colored background.

To avoid it, we should set it to SourceCopy which overwrites background color during Text rendering.

Read about CompositingMode: https://msdn.microsoft.com/en-us/library/system.drawing.drawing2d.compositingmode(v=vs.110).aspx

### DIFF
--- a/Samples/SVGViewer/DebugRenderer.cs
+++ b/Samples/SVGViewer/DebugRenderer.cs
@@ -93,7 +93,11 @@ namespace SVGViewer
             _transform.Translate(dx, dy, order);
         }
 
-
+        public CompositingMode CompositingMode
+        {
+          get { return System.Drawing.Drawing2D.CompositingMode.SourceOver; /* default value */ }
+          set { /* Do Nothing */ }
+        }
 
         public SmoothingMode SmoothingMode
         {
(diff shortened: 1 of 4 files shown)


tebjan:
hi, many people seem to have a problem with this commit, could you test it again and maybe make a fix for it? see https://github.com/vvvv/SVG/issues/237
in 2.2 your fix is in, before it was not in the nuget...

ubbn:
I will check and come back soon

tebjan:
would be a big help, thank you!

ubbn:
hi, you can revert this change. But original issue remain unsolved. As seems more tricky, I will take some time or probably someone else will fix meantime.
Thanks

FLoibl:
Issue #253 is referring to the same problem imho.

mrbean-bremen:
As this looks as if it will not be fixed soon, could you please revert the change for the time being as asked by @ubbn above? This would at least fix the mentioned regressions.
Thanks

tebjan:
sure, good point...

tebjan:
see: https://github.com/vvvv/SVG/pull/267

mrbean-bremen:
Thanks! That was fast!

tebjan:
no problem, was just two clicks. nuget update will also come soon...

PR #209, opened 2015-12-09
Split by comma, remove space in the end

ifcwebapp:
This is the fix of the ending space in the text coordinates as well as stroke-dasharray delimiter (should be comma). This fix would allow to render text and dashed paths using [Magick.NET](https://magick.codeplex.com/) while converting to other formats.


tebjan:
hi, i think the space separation is correct by the official svg documentation, but your method is more correct, i would just exchange comma by space.
http://www.w3.org/TR/SVG/coords.html

tebjan:
this is obsolete now... thanks anyway.

Issue #212, opened 2016-01-02
When rendering as image, markers are rendered off their offsets.


podostro:
When rendering to image with `SvgDocument.Draw()` method, the markers on the lines are renderer off their offsets. I'm using nuGet package v2.1.0

Here is an example of the output Image inside a Picturebox:
![image](https://cloud.githubusercontent.com/assets/2327913/12075280/ef71c528-b159-11e5-8ef5-c003936c24a2.png)

Here is the output from `SvgDocument.GetXML()` using the same document

	<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
	<svg x="0" y="0" width="100%" height="100%" overflow="inherit" viewBox="0, 0, 265, 265" preserveAspectRatio="xMidYMid" font-size="0" id="svgmap" xml:space="default" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1">
	  <defs id="defsmap" xml:space="default">
		<marker refX="3" refY="2" orient="auto" overflow="hidden" markerWidth="5" markerHeight="5" id="markerArrowLightBlue" xml:space="default">
		  <path d="M0 0 L4 2 L0 4 L1 2 z" id="pathMarkerArrowLightBlue" xml:space="default" style="fill:#ADD8E6;" />
		</marker>
		<marker refX="3" refY="2" orient="auto" overflow="hidden" markerWidth="5" markerHeight="5" id="markerArrowLightCoral" xml:space="default">
		  <path d="M0 0 L4 2 L0 4 L1 2 z" id="pathMarkerArrowLightCoral" xml:space="default" style="fill:#F08080;" />
		</marker>
	  </defs>
	  <g id="gmap" xml:space="default">
		<rect x="80" y="10" width="35" height="35" rx="7" ry="7" stroke="#ADD8E6" id="rectRoomX1Y0" xml:space="default" style="fill:#ADD8E6;" />
		<line x1="78" y1="47" x2="49" y2="76" marker-end="url(#markerArrowLightBlue)" stroke="#ADD8E6" stroke-width="3" id="lineLinkedRoomX0Y1" xml:space="default" />
[...]

Issue #210, opened 2015-12-09
Drawing Windows metafile generates an GDI exception


aconys:
Drawing in image works fine but in Metafile GDI generate an exception :
No problem in IE & Chrome the rendering is slow but fine

Exception :

System.Runtime.InteropServices.ExternalException (0x80004005): A generic error occurred in GDI+.
   at System.Drawing.Graphics.CheckErrorStatus(Int32 status)
   at System.Drawing.Graphics.FillPath(Brush brush, GraphicsPath path)
   at Svg.SvgRenderer.FillPath(Brush brush, GraphicsPath path) in c:\Dev\Plot3D\SVG-master\Source\Rendering\SvgRenderer.cs:line 59
   at Svg.SvgVisualElement.RenderFill(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\Basic Shapes\SvgVisualElement.cs:line 186
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer, Boolean renderFilter) in c:\Dev\Plot3D\SVG-master\Source\Basic Shapes\SvgVisualElement.cs:line 152
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\Basic Shapes\SvgVisualElement.cs:line 109
   at Svg.SvgElement.RenderChildren(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\SvgElement.cs:line 726
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer, Boolean renderFilter) in c:\Dev\Plot3D\SVG-master\Source\Basic Shapes\SvgVisualElement.cs:line 163
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\Basic Shapes\SvgVisualElement.cs:line 109
   at Svg.SvgElement.RenderChildren(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\SvgElement.cs:line 726
   at Svg.SvgElement.Render(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\SvgElement.cs:line 714
   at Svg.SvgFragment.Render(ISvgRenderer renderer) in c:\Dev\Plot3D\SVG-master\Source\Document Structure\SvgFragment.cs:line 180
[...]

Issue #204, opened 2015-11-09
Any way to load custom fonts when calling SvgDocument.Draw?


mmmeff:
We need to load custom fonts for rendering images, but I can't find any way to pass a System.Drawing PrivateFontCollection into the rendering - it looks like there's a hardcoded reference to System.Drawing.FontFamily.Families here, blocking out the ability to load custom fonts from disk/memory.

Given our deployment scenario, we cannot simply install the font into Windows to add the custom font to the drawing context and it seems like a PrivateFontCollection is the only exposed interface that would allow us to do so.

Is there some other way I'm missing? Or will this require a refactoring of SvgElementStyle's ValidateFontFamily method? https://github.com/vvvv/SVG/blob/44173b47e29429f98db60413f5058b016d1024f4/Source/SvgElementStyle.cs#L386

tebjan:
hope this is closed now with the mentioned pull request. otherwise please reopen it.

Issue #203, opened 2015-10-18
bug with Thread.CurrentThread.CurrentCulture


DevilPF:
In SvgPathBuilder.cs, SvgExtentions.cs, SvgDocument.cs and SvgColourConverter.cs you use construction when you change CultureInfo of current thread, do some work and return it back ... but unfortunately it's not even in try-finally clausule, so on many occasions it is not returned!
In SvgColourConverter.cs/ConvertFrom it is even skipped because of the return before - so the calling thread ends up with changed CultureInfo. I'm not even sure why do you need to change it this way - all converts have CultureInfo parameter to override that of the calling thread.

tebjan:
looks like an error, could you provide a pull request with the fixes?

DevilPF:
Hi, I'm quite new to GIT and not sure how to make branches, pull requests and all that stuff. It will be faster if I'll just give you the new versions: https://dl.dropboxusercontent.com/u/591980/Source.zip

tebjan:
hope this is closed now with the mentioned pull request. otherwise please reopen it.

PR #213, opened 2016-01-05
CurrentCulture Fix; SVG Viewer NullReferenceException fix; Little Maker rendering fix; Some optimization; Add "shape-rendering" support; Add "customer/ private fonts" support

HeinrichAD:
I started to use this SVG Rendering Engine and some thinks get my attention. I am sorry if this is too much for one pull request.

Fixes:
CurrentCulture Fix:
Now the application make sure to set back the old culture even an error occurred.
resole #152 [SVG load corrupts thread current culture]
resole #200 [Culture possibly not restored due to exception]
resole #203 [bug with Thread.CurrentThread.CurrentCulture]

SVG Viewer NullReferenceException fix:
A NullReferenceException occurred when the SVG XML will be loaded over the textbox. The reason is that in this case the svgDoc.BaseUri is null.

Maker rendering does not get right position fix:
This changes are based on issue #212 from mater.
NOTE: This fix is based on the issue. Please notice that the position finding of marker elements with the property "SvgMarkerUnits.StrokeWidth" is not really developed yet. [Nevertheless this should be work for the most cases.]
resole #212 [When rendering as image, markers are rendered off their offsets.]

Changes:
- Many other properties were unknowingly overloaded. I changed them to override.
- Some methods or structures need the [CLSCompliant(false)] Attribute. For example the Enums.TryParse extension. This extension has the where condition IConvertible. IConvertible is [CLSCompliant(false)] and that the reason why it is correct to mark the Enums.TryParse extension also as [CLSCompliant(false)].
- Property "float Opacity" in SvgElement was unknowingly overloaded in SvgGradientStop as string [stop-opacity]. In the end this property was only used as float. I changed the SvgGradientStop.Opacity property from string to float and to override.
- Some speed optimization. Note: Some optimization are based on C#6.0 but this is not new. Previously C#6.0 was already used. Note only that C#6.0 is only the default since Visual Studio 2015. If somebody used Visual Studio 2013 or older this person have to install the compiler nuget package [more see: https://github.com/dotnet/roslyn#download-c-and-visual-basic] and do some changes.
- I liked the method in ifcwebapp Pull Request #209 [Split by comma, remove space in the end]. I changed comma to space separation based on documentations [for more see the comments].

Adds:
Add "shape-rendering" Attribute basic support.
resole #161 [shape-rendering attribute not supported]

Add customer/ private fonts support
resole #204 [Any way to load custom fonts when calling SvgDocument.Draw?]

Add Unit Tests:
I added a SVG test helper class to avoid dupplicate code.

Unit Test metafile rendering:
Create metafile rendering Unit Test based on Issue #210 from master.
resole #210 [Drawing Windows metafile generates an GDI exception]

Unit Test to new feature "customer/ private fonts"

Unit Test to "maker-end" attribute based on issue #212.

Note & TODO: The already existing MultiThreadingTest referenced to a SVG test file which is not available. The creator should an this SVG to the resources.


tebjan:
thats truly a lot of fixes at once... but it looks well documented. i'll merge it as is. but please make one pull request per topic in the future. sometimes we have to revert a pull request it it creates problem, in this case all of the above would be reverted...

PR #229, opened 2016-04-30
changed rounding behavior of float to int color conversion from floor() to round()

popekim:
changed rounding behavior of float to int color conversion from floor() to round()

This is huge especially for the low values because human eyes distinguish darker colors much better.


tebjan:
excellent, thanks!

PR #232, opened 2016-05-22
Feature/external dependencies

dittodhole:
The inlined dependencies to ExCSS and Fizzler are now [true NuGet references](https://github.com/dittodhole/dotnet-System.Svg/tree/feature/external-dependencies/Source/packages.config).
This comes with one big disadvantage though: as these external dependencies are not signed, I had to disable signing for [Svg.csproj](https://github.com/dittodhole/dotnet-System.Svg/tree/feature/external-dependencies/Source/Svg.csproj) and [Svg.UnitTests.csproj](https://github.com/dittodhole/dotnet-System.Svg/tree/feature/external-dependencies/Tests/Svg.UnitTests/Svg.UnitTests.csproj) as well.

Additionally the whole project was upgraded to .NET 4.0, as the external dependencies are only available for .NET 4.0 - this is bearable imho


tebjan:
hello, sorry due to repeated user requests the library has to stay on .NET 3.5, also signing is important for many companies to protect their code. as much is i want to upgrade and profit from the new version i cannot accept the pull request at this time.

dittodhole:
regarding the .NET upgrade: one could also duplicate the project for Net40 and add these files with `target="lib\net40"` set.

see https://docs.nuget.org/create/nuspec-reference#metadata-section for more information

tebjan:
thanks for the insight, i'll consider releasing the nuget with two .net versions.

PR #226, opened 2016-04-19
Issue 225 - Image with large embedded image fails to render

AnkHansen:
- Changed SvgImage.Href property from Uri type to String to avoid UriFormatException if Uri contains too much data
- Changed GetImage to receive a string instead of Uri object.
  - Created a temp Uri object in method with substring of allowed length and using this to determine Uri characteristics

PR #233, opened 2016-05-22
fixed SvgImage.GetImage

dittodhole:
Fixing PR #226, which introduced a problem with data-uris shorter than 65519 characters. I've included a test.


tebjan:
excellent, thanks!

PR #234, opened 2016-05-22
made SvgImage.GetImage public

dittodhole:
I will use this for rendering Svg to other languages that do not rely on Image as the output.


tebjan:
looks good!

dittodhole:
Just in case you are curious regarding _other languages_: https://github.com/dittodhole/dotnet-Svg.Contrib.Render

PR #240, opened 2016-06-28
Fix stack overflow exception for em font-size value type

niksan321:
Hello, I found a critical bug, if svg attribute font-size set to enter values measurement units "em", then we get stackoverflow exception. Try to draw this svg -

```
<svg>
<text font-size = "2em"> Some Text </text>
</svg>
```


tebjan:
great, thanks!

PR #251, opened 2016-09-14
SvgPointCollection ToString method fix

sovietmagic:
**SvgPointCollection.ToString** now produce valid svg coordinate string. This fixes an issue when polylines and polygons (when converted to svg document file) were not rendering correctly due to their "points" attribute value was incorrectly formatted.

Now
- Each value pair is separated by whitespace and their x and y values - by comma. 
- No unit type is written to output string


tebjan:
great, thanks!

PR #255, opened 2016-11-03
Nested svg areas are misplaced or invisible 

frohlfs:
If the root svg uses transformations ( e.g. defines a viewBox and
different width and height) the nested svg areas are misplaced or
invisible.


tebjan:
Great, thanks!

PR #258, opened 2016-11-14
Copy the nodes of a SvgElement when creating a deep copy.

marcostroppel:
I wrote a deep copy of an SvgDocument to XML but the text element was missing completely. The problem was, that the SvgElement.Nodes were not copied in the deep copy. I added a method ISvgNode.DeepCopy() which will create a deep copy of these nodes. These deep copies of the nodes are now also added to the nodes of a SvgElement in the SvgElement.DeepCopy.
I added a test with a SvgText element where the whole text element was not written to XML of a deep copy of the SvgDocument without this bugfix.

### DIFF
--- a/Source/SvgContentNode.cs
+++ b/Source/SvgContentNode.cs
@@ -1,12 +1,17 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-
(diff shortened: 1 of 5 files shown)


tebjan:
great, thanks!

PR #261, opened 2016-11-18
Circle, ellipse, polygon and rectangle render correctly when their stroke-width is set to zero

### LINKED ISSUE #256: Polygon not rendered if "stroke-width" = 0
The following svg renders correctly in my browser, but produces nothing in this library. There seems to be a problem when the stroke-width is explicitly set to 0.

<svg id="svg2" version="1.1" preserveAspectRatio="xMidYMid meet" viewBox="0 0 200 200">
  <polygon fill="Red" stroke="Black" stroke-width="0" points="0,0 200,0 0,200 200,200" />
</svg>

sovietmagic:
I've fixed strange behaviour when circle, ellipse, polygon and rectangle will not render if their StrokeWidth is less or equal zero. I removed this check from their Path overrides and moved it to their base class's (SvgVisualElement) RenderStroke method.

- Now stroke won't be rendered in any ancestor of SvgVisualElement that has StrokeWidth  <= 0.
- Circle, ellipse, polygon and rectangle will render only their fill if StrokeWidth  <= 0.

fixes #256

### DIFF
--- a/Source/Basic Shapes/SvgCircle.cs
+++ b/Source/Basic Shapes/SvgCircle.cs
@@ -89,7 +89,7 @@ namespace Svg
         /// </summary>
         public override GraphicsPath Path(ISvgRenderer renderer)
         {
-            if ((this._path == null || this.IsPathDirty)	&& base.StrokeWidth > 0)
+            if (this._path == null || this.IsPathDirty)
             {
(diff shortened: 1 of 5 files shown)


tebjan:
looks very good, thanks!

PR #269, opened 2017-01-06
Update SvgDocument.cs

avogelba:
1. Added XML Header, this is conform according to https://www.w3.org/TR/SVG/struct.html
2. Added support for removing Byte Order Mark (BOM), this needed in some cases, eg. if the SVG is used as SSI (Server Side Include) in IIS (Internet Information Server).


tebjan:
thanks, looks very good.

Issue #281, opened 2017-02-23
Bounds in path based elements do not consider transformations


mrbean-bremen:
For path-based elements (paths, lines, circles, rectangles, polygons) the Bounds property does not consider transformations. I think this is not intended - it is especially a problem if getting the bound of an `SvgGroup `with elements which have different transformations.

This can be fixed by using:
``` C#
public override RectangleF Bounds
{
    get
    {
        var path = this.Path(null);
        if (Transforms != null)
        {
            path = (GraphicsPath)path.Clone();
            path.Transform(Transforms.GetMatrix());
        }
        return path.GetBounds();
    }
}
```
instead of:
``` C#
public override RectangleF Bounds
{
    get { return this.Path(null).GetBounds();
}
```
in the respective elements. I can create a respective PR if this fix makes sense.

PR #282, opened 2017-02-25
Add transform to bounds calculation in path based objects

mrbean-bremen:
- added bounds tests for some elements
see #281


tebjan:
great, thanks a lot!

PR #285, opened 2017-03-02
Added array caching to AddToPath for a significant performance increase.

LightningDragon:
Every time you index graphicsPath.PathPoints or graphicsPath.PathTypes the accessor rebuilds the arrays and makes dll calls, graphicsPath.PointCount also makes dll calls. this is fixed by simply assigning the arrays to local variables and using them from there. Using this [SVG](https://upload.wikimedia.org/wikipedia/commons/e/e8/Dualshock_4_Layout.svg) as an example it reduces the render time from 12-13 seconds to 1-2 seconds.


tebjan:
nice find, than ks a lot!

PR #286, opened 2017-03-02
Do not create a clip region for view box

### LINKED ISSUE #279: View box always has clip box
If I have a document with an embedded SVG fragment which has a view box. This fragment is always clipped if shown with SVG lib. If I look at the same image in a browser (checked with Firefox, Chrome and IE), the fragment is only clipped if the overflow is not set to visible.

The clip rect is created in `SvgViewBox.AddViewBoxTransform()`. I'm not sure what the correct behavior should be, but I guess that the clip rect shall only be created if the overflow for the element is set accordingly, maybe by adding an additional property to `SvgViewBox`.

EDIT: At least for `SvgFragment`, it works correctly for me if I just remove the clip region creation in `SvgViewBox`, e.g. a clip region is only created if `Overflow` is set to hidden. I'm not sure about how overflow handling is done for markers and patterns, though.

mrbean-bremen:
- fixes #279
- overflow for markers and images does not work as before (not implemented), so this change should have any other impact


tebjan:
hello, i dont really get this sentence: "overflow for markers and images does not work as before (not implemented), so this change should have any other impact"

can you rephrase it?

mrbean-bremen:
Ok, what I meant was that the overflow attribute is also valid for markers and patterns (not images, this was incorrect), but has no effect in the SVG library rendering (neither before nor after my change). As far as I can see, the respective functionality is not implemented. I checked this to avoid any regressions.
I  hope this is more clear...

mrbean-bremen:
Also,  I just noticed that more changes got into this PR than intended - I had some trouble with the GitHub GUI and messed it up. I will resubmit the PR tomorrow.

tebjan:
alright, thanks!

PR #288, opened 2017-03-03
Do not create a clip region for view box

### LINKED ISSUE #279: View box always has clip box

mrbean-bremen:
- a view box does not need a clip region, this breaks the overflow
handling
- overflow handling:  handle overflow attribut "Inherit" as Visible and
"Scroll" as Hidden as per W3C documentation
- fixes #279


mrbean-bremen:
This is the re-commit of the PR I created and deleted yesterday. I added fixed handling of "inherit" and "scroll" overflow values for SVG fragment.

PR #292, opened 2017-03-10
Use Hidden as default SvgOverflow value

mrbean-bremen:
- changes behavior to be as described in this file and in the W3C test
masking-path-03-b
- fixes that test
- this has been a regression after removing the unconditional clip box
for every view box, (see #279)


tebjan:
great, thanks for the fix!

PR #295, opened 2017-03-12
Fixing Tiger Test

Shkyrockett:
This reverts the change that broke the Tiger test.


mrbean-bremen:
@Shkyrocket: Is this a replacement for my PR? Shall I close it?

mrbean-bremen:
@Shkyrockett :
On a second look, your revert also reverts the very nice performance improvement for path rendering by @LightningDragon. While the call to `PointCount` seems unavoidable, I would keep copying the path points (see #294).

tebjan:
after merging #294 tiger.png seems to work. can you confirm? if yes i'll close this issue.

Shkyrockett:
Yep, the check for whether there are any points seems to mitigate the access violation.

PR #296, opened 2017-03-12
Reverted unneeded part of previous fix

mrbean-bremen:
- caused regression


mrbean-bremen:
Sorry, committed too much last time.

PR #297, opened 2017-03-13
Found better solution to SvgClosePathSegment tested with Tiger.

LightningDragon:
PathData doesn't throw an exception when PointCount is 0 and gets the points and types at the same time. tested it with Tiger and everything seems to be working.


LightningDragon:
Oh and PathData only makes one dll call so it's even faster than the other caching method.

tebjan:
looking good, thanks!

Issue #243, opened 2016-07-01
SVG is not correctly interpreted, e.g. US-Flag's stars are distributed over the whole graphic


cbries:
It looks like that the "path" is not correctly interpreted:
`<path d="M0,450H7410m0,600H0m0,600H7410m0,600H0m0,600H7410m0,600H0" stroke="#fff" stroke-width="300"/>`  
The resulting "Star" is distributed over the whole graphic.

Example application to reproduce this issue (svg is uncluded):
[WpfApplication1.zip](https://github.com/vvvv/SVG/files/343116/WpfApplication1.zip)

By the way, the svg is copied from Wikipedia:
https://en.wikipedia.org/wiki/United_States#/media/File:Flag_of_the_United_States.svg

![followupissue](https://cloud.githubusercontent.com/assets/5418062/16516633/51ffb238-3f79-11e6-9a5f-efa1f0455f59.PNG)

ThisIsRobokitty:
Can confirm, I have the same issue with the most recent source.

mrbean-bremen:
Duplicate of #202, both have been fixed some time ago (test is present).

mrbean-bremen:
Closing as fixed.

PR #300, opened 2017-03-19
Fixed inheritance of  text-anchor and baseline-shift attributes

mrbean-bremen:
- use GetInheritedAttribute() to correctly handle "Inherit" value
- fixes inheritance of these attributes from groups
- force redraw of SvgUse children to ensure correct style inheritance
- fixes alignment in types-basic-01-f.svg and text-text-12-t.svg tests


mrbean-bremen:
Note: the US flag test (first commit) is unrelated to the actual fix. I just added it after I broke it accidentally, to avoid regressions of #243.

PR #305, opened 2017-03-25
Add a text copy of the Ms-PL and update Markdown formatting

rogersachan:
In this PR:

* I added a text copy of the MS-PL, allowing detection by Github.
* I formatted and linted the readme, ensuring that it is clean and readable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,25 +1,22 @@
-SVG.NET
-=======
+# SVG.NET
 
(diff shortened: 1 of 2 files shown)


rogersachan:
@tebjan GitHub now also gives info based on license, making this possible (visible in my PR.)

image

PR #308, opened 2017-03-28
Fix scaling of Draw(width, height)

alex-sherman:
For whatever reason, these lines have been commented and uncommented several times. I believe it is correct to NOT call scale, because it appears to have been set somewhere else already. An easy way to observe this is to simply change the SVGViewer sample to call `svgDoc.Draw(width, height)` and measure the output. It is clear that the scaling happens twice, changing the width and height creates a non-linear scaling response.

Here you can see the original response to calling Draw(64, 64)
before

And here is the response with the pull request
after


tebjan:
thanks for digging into it.

PR #309, opened 2017-03-28
Removed Default Parameters to restore full .NET 3.5 support. Fix scaling

Fr33dan:
I filed an issue but no one responded so I took it upon myself to fix the problem.

I removed any default parameters used and replaced them with overloaded methods that call the functional version. For the interfaces where I couldn't write a method body I used added extension methods instead. 

The extension method class was misnamed so I fixed that too.

Once I got the code to compile I discovered that scaling did not work on a draw to bitmap and fixed that as well.

### DIFF
--- a/Source/DataTypes/SvgUnit.cs
+++ b/Source/DataTypes/SvgUnit.cs
@@ -181,7 +181,7 @@ namespace Svg
         {
             if (owner == null) return null;
             var visual = owner.Parents.OfType<SvgVisualElement>().FirstOrDefault();
-            return visual?.GetFont(renderer);
+            return visual == null ? null : visual.GetFont(renderer);
         }
 
         /// <summary>
--- a/Source/External/ExCSS/IToString.cs
+++ b/Source/External/ExCSS/IToString.cs
@@ -2,6 +2,6 @@
 {
     public interface IToString
     {
-        string ToString(bool friendlyFormat, int indentation = 0);
+        string ToString(bool friendlyFormat, int indentation);
     }
 }
--- a/Source/External/ExCSS/Model/MediaTypeList.cs
+++ b/Source/External/ExCSS/Model/MediaTypeList.cs
@@ -75,10 +75,10 @@ namespace ExCSS
 
         public override string ToString()
         {
-            return ToString(false);
+            return ToString(false, 0);
         }
 
-        public string ToString(bool friendlyFormat, int indentation = 0)
+        public string ToString(bool friendlyFormat, int indentation)
         {
             return friendlyFormat
                 ? string.Join(", ", _media.ToArray())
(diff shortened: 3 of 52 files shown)


Source/DataTypes/SvgViewBox.cs
```diff
@@ -156,10 +156,10 @@ public void AddViewBoxTransform(SvgAspectRatio aspectRatio, ISvgRenderer rendere
                     fScaleX = Math.Min(fScaleX, fScaleY);
                     fScaleY = Math.Min(fScaleX, fScaleY);
                 }
-                float fViewMidX = (this.Width / 2) * fScaleX;
-                float fViewMidY = (this.Height / 2) * fScaleY;
-                float fMidX = width / 2;
-                float fMidY = height / 2;
+                float fViewMidX = (this.Width / 2);
+                float fViewMidY = (this.Height / 2);
+                float fMidX = width / 2 * fScaleX;
```
tebjan: thanks for the pull request, but what are these changes? they don't seem to be related to .NET 3.5. there was recently a pull request regarding the same problem. please remove this file from the pull request, update your repo to the latest state, test again and make a separate PR for this issue.

Fr33dan:
I made those changes because as I said, when I got it compiled scaling was still wrong. At the time given no response after 4 days and several other ignored issues I thought the project was abandoned. I just needed it to work so I made it work.

It was only when I went to make a new fork to share the changes I had already made that I noticed there were new commits since I snagged the code and the project was still alive. I had to fight with git to include the changes made on those commits. I went ahead and I tried them and they did not fix the scaling issue. 

If I only change the SvgViewBox the changes I made to SvgElement.Draw for that issue will still be included in this pull request. I'm not sure how to split two changes made on one file like that but I'll see what I can do when I get to my computer.

tebjan:
>  I'm not sure how to split two changes made on one file like that but I'll see what I can do when I get to my computer.

you can do this by creating two branches and make a pull request from each. the two branches must start from the current official code base and only contain the relevant commits.

Fr33dan:
Thank you for the guidance on that, it would have taken me a long time to figure that out. I'm not very good with git. I think I've cleaned up this request so that now it is only for the removal of default parameters.

Fr33dan:
The dust on this one should be settled now. Came back to this one as it's the only one I can compile after setting up the other two branches/pull requests. 

It is the current version I'm using but since I don't want to push the scaling updates on top of it my code is not fully functional. Since then and now I moved away from the Draw(Bitmap) and use Draw(Graphics) instead but when my viewport gets bigger my image gets smaller.

Fr33dan:
The attempts to do multiple things with this single merge made this a mess.

I also now understand it was not necessary in the first place. The presence of default parameters doesn't make it incompatible with .NET 3.5, but only with VS2008. So while I couldn't work with it in VS2008 Standard, I could have used the Community version of the latest editions to make any changes/contributions I need to this open source project which is allowed under the Community license.

I could have then compiled that to a DLL and referenced that in VS2008 for my enterprise application that I cannot develop in the community edition.

PR #310, opened 2017-03-29
Fixed made to make view box scaling and draw to bitmap work as it should.

Fr33dan:


tebjan:
> view box scaling and draw to bitmap work as it should.

do you have test cases for that? how many different ones and do they all work?

Fr33dan:
I only have two test cases. The both work as expected. I even went a changed the the preserveAspectRatio value of them after getting it to work and realizing that I wanted xMinYMin and not xMidYMid.

mrbean-bremen:
@Fr33dan : I just tested this PR, and it seems to break a couple of tests:
- coords-viewattr-01-b.svg
- coords-viewattr-03-b.svg
- __issue-279-01.svg

You can check these using the test runner application.
Also, if you have you own test images that cover functionality not covered by the existing W3C tests, it makes sense to add them together with the png. You can create the png with the SVGViewer application.

Fr33dan:
I am seeing that problem as well. I have been playing with it for a couple of hours and can't find a solution that works for my usage and gets those tests working. 

> Also, if you have you own test images that cover functionality not covered by the existing W3C tests, it makes sense to add them together with the png. You can create the png with the SVGViewer application.

I was easily able to add my image to the test program but it doesn't do any good since the problem only occurs when I'm drawing the image at a size that isn't the original image size. 

At this point if my usage has brought a problem with the library to light or if I'm using it incorrectly. Either way I'm done fighting with it today. I'll be back on it tomorrow.

Fr33dan:
Okay well the problem was a combination my SVG files not being matching what I had in my head and an me incorrectly using the library.

Once I figured out what I was doing wrong I did have to make some changes to the library to be able to get the functionality I want. That is being able to draw my scaled image directly to a graphics object.

My primary goal is to be able to print the image with the other data in a report. The new draw method works great for rendering on the screen graphics object in OnPaint (I don't understand why everyone is drawing to a PNG to then draw the PNG, seems inefficient). When I try to render to the graphics object in my PrintPage event however it always draws it at at the origin. I haven't given up yet but I've hit a wall for the moment.

mrbean-bremen:
Ok, this version looks certainly better :)
A hint  (as you are new to git) - try `git rebase -i` to squash your commits (you have to use `git push --force` afterwards, which is OK as long as nobody has been using your code).
As to rendering - I am rendering to a memory stream to draw a bitmap, no PNG needed there, though I'm quite new to .NET myself, so I'm not the best advisor. Good luck with your project!

Fr33dan:
This pull request is a mess and no longer does what it originally say it does. I rebased from the latest version and made my changes their so I'm closing this request.

PR #311, opened 2017-03-29
Fix extensions

Fr33dan:
The extension methods for the library were split between two files, one of which had an typo in the name.


mrbean-bremen:
The moved `Extensions.cs` file is missing in this PR - please fix this, or create a new one. 
Also, it would be nice if no unrelated changes (in this case added references) are included.

mrbean-bremen:
Ok, given that this is quite old and not working as is, I'm closing this now.
Feel free to add a new PR.

Issue #224, opened 2016-04-03
Exception: "message":"An error has occurred.","exceptionMessage": "ColorBlend object that was set is not valid.Position's last element must be equal to 1.0.  ColorBlend objects must be constructed with the same number of positions and color values.  Positions must be between 0.0 and 1.0, 1.0 indicating the last element in the array."


bromose:
Hi. I got:
The following exception:
{"message":"An error has occurred.","exceptionMessage":
"ColorBlend object that was set is not valid.Position's last element must be equal to 1.0. 
ColorBlend objects must be constructed with the same number of positions and color values. 
Positions must be between 0.0 and 1.0, 1.0 indicating the last element in the array."
,"exceptionType":"System.ArgumentException","stackTrace":" 
at System.Drawing.Drawing2D.LinearGradientBrush._SetInterpolationColors(ColorBlend blend)
at Svg.SvgLinearGradientServer.GetBrush(SvgVisualElement renderingElement, ISvgRenderer renderer, Single opacity, Boolean forStroke)
Svg.SvgVisualElement.RenderFill(ISvgRenderer renderer)
at Svg.SvgVisualElement.Render(ISvgRenderer renderer, Boolean renderFilter)
at Svg.SvgElement.RenderChildren(ISvgRenderer renderer)
at Svg.SvgVisualElement.Render(ISvgRenderer renderer, Boolean renderFilter)
at Svg.SvgElement.RenderChildren(ISvgRenderer renderer)

With help from :stackoverflow.com/questions/25268107/svg-renderer-throws-color-blend-object-not-valid-with-telerik-radhtmlchart

I was able to make a fix in "SvgGradientServer", protected ColorBlend GetColorBlend(ISvgRenderer renderer, float opacity, bool radial):

`                position =
                    radial
                        ? 1 -
                          (currentStop.Offset.ToDeviceValue(renderer, UnitRenderingType.Horizontal, this)/boundWidth)
                        : (currentStop.Offset.ToDeviceValue(renderer, UnitRenderingType.Horizontal, this)/boundWidth);

                **position = (float) Math.Round(position, 1, MidpointRounding.AwayFromZero);**`

Feel free to use my "hack" in your great projekt.

Best regards,
Kaj Bromose

Steven0621:
the problem still exist! 
who knows how to solve?
thanks!

bansan85:
[...]

PR #314, opened 2017-04-20
Fix rendering error on 64 bit IIS, ColorBlend position must be betwee…

epuig-ssinergie:
#224


tebjan:
thanks!

sir-wilhelm:
@tebjan are there any plans to include this fix in a nuget package?

PR #316, opened 2017-05-10
Fix Visual Studio 2013

ststeiger:
In SVG/DataTypes/SvgUnit.cs 

Function private IFontDefn GetFont(ISvgRenderer renderer, SvgElement owner) 
contains a 
`return visual?.GetFont(renderer);`
this break compilation in Visual Studio 2013. 

Fix with equivalent code 
`return (visual != null) ? visual.GetFont(renderer) : null;`
so that it compiles fine in Visual Studio 2013


mrbean-bremen:
@ststeiger - there are a lot of changes here, looks like only the first commit belongs to this PR.
Can you check please?

mrbean-bremen:
Ok, I added your fix from the first commit manually - closing this PR now.
Hint for future PRs - better use a separate branch for PRs instead of master.

PR #326, opened 2017-07-20
Netstandard20 support

mmosca:
Hello,

I have updated the project to use Visual Studio 2017 15.3 preview and made it target multiple .Net framework version.

.Net 3.5 as the original
.Net 4.0 since it seems to work better with the .Net SDK command line tooling (no visual studio required)
.Netstandard2.0 to support .Net Core 2.0+ apps.

The .Netstandard 2.0 version uses a port of System.Drawing callsed ZKWeb.System.Drawing.

Most of the code changes have been to add conditional compilation using statements and clear up namespace conflicts, since ZKWeb uses System.DrawingCore, instead of System.Drawing.

Nuget package generation is now integrated with Visual studio, the old .nuproj project is no longer needed.


calico-crusade:
There is currently System.Drawing.Common and System.Drawing.Primitives in the works from Microsoft. They are currently beta packages. I think it would be a decent idea to move to those implementations instead of ZKWeb.System.Drawing.

8VAid8:
Should code like this work for that branch?
> SvgDocument SvgDocument = SvgDocument.Open(filePath);
> SvgDocument.Draw(imgWidth, imgHeight)

P.S. I have tried it for my core version with CoreCompat.System.Drawing.v2, but i got only a black shapes instead of proper svg.

AtomCrafty:
@8VAid8 In case you're still interested in this: I had a similar issue, #365 fixed it for me.

mrbean-bremen:
FYI: As discussed with @tebjan, this will be merged somewhere after the next nuget release.

codeunifier:
These changes were life-saving. Glad to hear that it'll be implemented in upcoming releases!

tebjan:
lets make a netstandard-2.0 branch for this in the repo, so the issues can be resolved and tested there.

tebjan:
and it should probably switch to that nuget instead of ZKWeb.System.Drawing: https://www.nuget.org/packages/System.Drawing.Common

mrbean-bremen:
> lets make a netstandard-2.0 branch for this in the repo

good idea. Should we resolve the conflicts first?

mrbean-bremen:
I rebased with trunk and created a [new branch](https://github.com/vvvv/SVG/tree/netstandard2) in the main repo, but there are still some problems:
- `ColorConverter` seems not be available in netstandard2, or I'm doing something wrong
- the unit tests complain about the missing 4.0 target, but this may depend on the installed test framework
- the AppVeyor config has to be changed to use VS 2017

Generally, this will work only from VS 2017, as far as I understand - is this correct, and if yes, is this what we want in master?
I may need some help here, will look further tomorrow if I find the time...

Noah1911:
Looks like the ColorConverter and KnownColor is being added in **netstandard-2.1** according to this catalog https://apisof.net/catalog/System.Drawing.ColorConverter and pull request https://github.com/dotnet/standard/pull/803 which targets **netstandard-2.1** milestone.

tebjan:
can we have a local copy of the class in this repo for the time being? registering it should be easy...

mrbean-bremen:
Sounds reasonable. I may have a go at the weekend (don't have much time right now), or maybe someone else wants to try it...

8VAid8:
> @8VAid8 In case you're still interested in this: I had a similar issue, #365 fixed it for me.

Thanks, but i've solved the issue already. I used that branch - https://github.com/FastReports/SVG. It works fine with core and netstandard 2.0.

mrbean-bremen:
@8VAid8 - I checked the branch you referred to, and it basically replaced the svg project with a new one, which has an additional `netstandard2` target, and used `System.Drawing.Common` and `CoreCompat.System.Drawing.v2`. I also saw that it uses a VS2015 solution, so VS2017 is obviously not needed for this.
In the current branch, only `System.Drawing.Common` is used - `CoreCompat.System.Drawing.v2` seems to provide mono support, so maybe we shall add this, too? As I wrote before, I know basically nothing about `netstandard2` (having used only .NET 3/4, and only under Windows) and want to be sure to not break any compatibility by merging the PR.

mrbean-bremen:
Some status update - I merged all commits together (in the [other branch](https://github.com/vvvv/SVG/tree/netstandard2)) and partly cleaned them up to remove unneeded changes.
I'm probably going to recreate the branch from scratch with as few changes as possible, but that may take some time (I'm a bit short on time at the moment). 
I will leave this PR open as a place for related discussion.

mrbean-bremen:
I'm going to close this now, as it is hopelessly out-of-date, and I think we can have another go at Netstandard compatibility after .NET Core support has been stabilized. 
As far as I can see, this will be mostly a case of adding Netstandard to the project files and handle the missing ColorConverter (or using a newer version of Netstandard).

PR #351, opened 2018-01-12
Fixed bug causing all glyphs in an embedded SvgFont to be drawn.

hrclausandersen:
When embedding fonts in an .svg file the glyph paths are all rendered when the document is rendered. To avoid this an empty override of the Render method is added to the SvgFont class.

### DIFF
--- a/Source/Text/SvgFont.cs
+++ b/Source/Text/SvgFont.cs
@@ -51,5 +51,8 @@ namespace Svg
         {
             return base.DeepCopy<SvgFont>();
         }
-    }
+
+        protected override void Render(ISvgRenderer renderer)
+        { }
+	}
 }


Source/Text/SvgFont.cs
```diff
@@ -51,5 +51,8 @@ public override SvgElement DeepCopy()
         {
             return base.DeepCopy<SvgFont>();
         }
-    }
+
+        protected override void Render(ISvgRenderer renderer)
+        { }
+	}
```
mrbean-bremen: looks like some TAB/SPACE mixup here

hrclausandersen:
I have everything set up for tabs so my changes would probably have introduced some tabs in the source. Sorry for not being consistent with the original style.

mrbean-bremen:
No problem - there is no style guideline, and the tabs and spaces are mixed up anyway in existing code.
I may make a separate PR for this later (as it plays havoc with different viewers), but not before most of the PRs are merged, to avoid conflicts.

PR #354, opened 2018-02-06
Compatibility with SVG generated by Highcharts in IE

lpa-vadim-zaslavsky:
This pull request fixes the following problems in the library when it encounters snippets like

```
<text style="font: bold 15px/normal externalfont; text-transform: uppercase;">Text</text>
```

- font shorthand is parsed incorrectly;
- font weight and variant are not being mapped to the enumerations;
- text is not rendered uppercase.

### DIFF
--- a/Source/DataTypes/SvgTextTransformation.cs
+++ b/Source/DataTypes/SvgTextTransformation.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.ComponentModel;
+
+namespace Svg
+{
+    /// <summary>This property describes transformations that are added to the text of an element.</summary>
+    [TypeConverter(typeof(SvgTextTransformationConverter))]
+    [Flags]
+    public enum SvgTextTransformation
+    {
+        /// <summary>The value is inherited from the parent element.</summary>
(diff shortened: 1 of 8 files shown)


mrbean-bremen:
@lpa-vadim-zaslavsky - this looks good to me! Do you have a text image with the corresponding png, that you can check-in to get this tested?

lpa-vadim-zaslavsky:
@mrbean-bremen, an example would be the following SVG:
``` xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 50" width="100" height="50">
  <text x="20" y="20" style="font: italic bold 15px/normal &quot;Fira Code&quot;; text-transform: uppercase;">Abc</text>
  <text x="20" y="40" style="font-weight: bold; font-style: italic; font-size: 15px; font-family: &quot;Fira Code&quot;; text-transform: uppercase;">Def</text>
</svg>
```

v2.3.0 produces this output (first line is missing, no uppercase transformation applied):
output_2 3 0

v2.4.0 produces this output (same as what the browsers render):
output_2 4 0

Thanks for merging this!

mrbean-bremen:
Thanks - I added this as a regression test.

PR #356, opened 2018-02-22
Text positioning when printing - vertical graphics context resolution

mul82:
Hi,

Vertical graphics context resolution has been used in an inconsistent way. Only in Gdi/SvgFontDefn.cs/Ascent method, the DpiY property of graphics context is used to get the font baseline. This leads to wrong positioned text when printing.

Many thanks & Regards,
Matthias


tebjan:
great, thanks! looking into it after i merged the other  PRs.

PR #386, opened 2018-12-27
Fix just Compiler Warning (level 3) CS0659.

H1Gdev:
`SVG\Source\SvgAttributeAttribute.cs(13,18,13,39): warning CS0659: 'SvgAttributeAttribute' overrides Object.Equals(object o) but does not override Object.GetHashCode()`

https://docs.microsoft.com/dotnet/csharp/misc/cs0659

ref. #365


mrbean-bremen:
Thanks!

PR #394, opened 2018-12-30
Add additional "Draw" (to Graphics) without optional parameter

heindlalex:
the recently introduced optional size parameter is causing incompatibilities with our C++ application.
@mrbean-bremen - I would appreciate if we can add this additional, so we can stay compatible without any change in the calling application.

FYI - I merged the master back to my forked master, let my know if this works without the need to create a new branch again, at least it shows only one change, but still more than one commit.

Thx.


Source/SvgDocument.cs
```diff
@@ -438,6 +438,16 @@ public void Draw(ISvgRenderer renderer)
             this.Render(renderer);
         }
 
+        /// <summary>
+        /// Renders the <see cref="SvgDocument"/> to the specified <see cref="Graphics"/>.
+        /// </summary>
+        /// <param name="graphics">The <see cref="Graphics"/> to be rendered to.</param>
+        /// <exception cref="ArgumentNullException">The <paramref name="graphics"/> parameter cannot be <c>null</c>.</exception>
+        public void Draw(Graphics graphics)
```
mrbean-bremen: Hm, shouldn't you remove the default value from the `size` parameter in the other method in this case?

mrbean-bremen:
This branch has conflicts and cannot be merged as is.
I suggest you create a new branch from original master and cherry-pick the relevant commit, or just save the changes and copy them. As I wrote, using your master branch for PRs is not very convenient. 
It may take some time to get used to the git/GitHub workflow (it took for me), but it avoids these problems. I don't know how used you are to git - in case you are new to it, there is the book Pro Git (available online for free ), that has helped me a lot.

Source/SvgDocument.cs
heindlalex: You're right. I should remove the default value, but keep the null conditional operator ("?") operator, which I believe was done to  have "null" as default for this parameter (no conversions in SizeF).
In case someone actually uses the optional parameter with "null", to stay compatible.

heindlalex:
I created a new clean pull request #395 from a new branch. I hope that works. Thx.

PR #414, opened 2019-02-01
use element applies referenced viewBox scaling. FontManager created.

peracto:
Before I begin, apologies, this is my first ever pull request ...

1) The 'use' element. 

When referencing a element that contains a viewBox attribute - the result should be scaled using the width/height defined on the use element and the width/height of the viewbox.

I created a test case ... "__issue-323-02-01" - but realise that I've not followed any proper naming convention.

2) Fonts ... My use case required the use of adhoc fonts. The existing method SVGElement.PrivateFonts - partially solved the issue, but not completely ... Introducing the SvgFontManager. 

When a font is requested in the render process, if the font is not found as an embedded SvgFont, it defers to the SvgFontManager.FindFont method. The SvgFontManager.FindFont searches a dictionary of fonts (pre loaded with the system fonts).  If a font can't be found and a callback has been provided - then the callback should perform any validation and return a font (or null if not found/error). When the SvgFontManager can't identify the font, we then fallback to the SVGElement.PrivateFont method.

Example usage.

            SvgFontManager.FontLoaderCallback = name =>
            {
                var ff = Path.Combine(Config.FontsFolder, name + ".ttf");
                if (!File.Exists(ff))
                {
                    Log.Error($"Font request ({name}) does not exist at {ff}");
                    return null;
                }

                var rc = SvgFontManager.AddFontFile(ff);
                if (rc == null) Log.Error($"Font request ({name}) font not loaded. {ff}");
                return rc;
            };


mrbean-bremen:
> this is my first ever pull request

Congratulations, and thanks :) 
> I created a test case ... "__issue-323-02-01" - but realise that I've not followed any proper naming convention.

I think it would be better to rename it to `__pull_request-414-01` to relate it to the correct PR, but other than that, it looks good to me! I will have a closer look tomorrow, but so far it looks sensible.

Source/Document Structure/SvgUse.cs
```diff
@@ -131,6 +167,19 @@ protected override void Render(ISvgRenderer renderer)
...
+                    {
+                        var viewBox = element.Attributes.GetAttribute<SvgViewBox>("viewBox");
+                        if (viewBox!=SvgViewBox.Empty && Math.Abs(ew - viewBox.Width) > float.Epsilon && Math.Abs(eh - viewBox.Height) > float.Epsilon)
+                        {
+                            var sw = ew / viewBox.Width;
+                            var sh = eh / viewBox.Height;
+                            renderer.ScaleTransform(sw, sh, MatrixOrder.Prepend);
+                        }
+                    }
+
```
mrbean-bremen: Maybe you could use `ViewBox.AddViewBoxTransform` here instead? Not sure about the aspect ratio, though.

Source/SvgFontManager.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Text;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Svg
+{
+    public static class SvgFontManager
```
mrbean-bremen: Could you please add a description what this does and how it is used (along the lines you put in the comments in this PR already)?

Tests/Svg.UnitTests/PassingTests.csv
```diff
@@ -187,6 +187,7 @@ __issue-239-01
 __issue-242-01
 __issue-279-01
 __issue-323-02
+__issue-323-02-01
```
mrbean-bremen: As already mentioned - please rename to __pull_request-414-01

Source/SvgFontManager.cs
peracto: complete.

Tests/Svg.UnitTests/PassingTests.csv
peracto: complete.

Source/Document Structure/SvgUse.cs
peracto: I can see the intention of the `ViewBox.AddViewBoxTransform`. If the referenced element was an `SvgFragment`, then I could inject the Width/Height from the `SvgUse` element into it. The problem is that not all of the elements that support viewBox are `SvgFragments`.

I could investigate a more robust approach - but given that noone else has raised an issue - and that the solution works for my specific use-case - could we merge as-is?

mrbean-bremen: That's ok - it was only a suggestion, I'm fine with the current state.

Issue #399, opened 2019-01-06
System.NotSupportedException with specific svg file


mrbean-bremen:
This is a spin-off of #59, as this part is not related to the original issue.

@dillanmann commented on 6 Sep 2017:
The version I am using is version 2.2.1.39233, the latest version available via NuGet.

The code that is throwing the error is:

```
        public static void RenderSvgStringToPng(XDocument xmlDoc, string outputFile)
        {
            var innerXml = xmlDoc.ToString();
            using (var xmlStream = new MemoryStream(Encoding.Default.GetBytes(innerXml)))
            {
                xmlStream.Position = 0;
                var doc = SvgDocument.Open<SvgDocument>(xmlStream); // <---- throwing here
                using (var image = doc.Draw())
                    image.Save(outputFile, ImageFormat.Png);
            }
        }
```

Note that it throws the error at SvgDocument.Open regardless of whether I just give it the plain SVG string (from `xmlDoc.ToString()` using `SvgDocument.FromSvg`) or whether I use the memory stream way.

The error I get is:

```
An exception of type 'System.NotSupportedException' occurred in Svg.dll but was not handled in user code

Additional information: Must call AddTerm AddSeparator in that order
```

I noticed some first chance exceptions logged:

```
A first chance exception of type 'System.ArgumentNullException' occurred in Svg.dll
iisexpress.exe Warning: 0 : Attribute 'd' cannot be set - type 'Svg.Pathing.SvgPathSegmentList' cannot convert from string ''.
A first chance exception of type 'System.NotSupportedException' occurred in Svg.dll
```

This is all the detail I could find, let me know anything else you need! :) 

@gvheertum commented on 6 Sep 2017:
@dillanmann no worries :) Thanks for the details. 

From what I can see it fails on something *in* the SVG rather than crashing on threading issues. Nevertheless, if there is something making the SVG invalid, it should fail with more grace than a nullpointerexceptions. From what I see now (after a quick glance at the problem) is it tries to make a path, but the property is empty (iisexpress.exe Warning: 0 : Attribute 'd' cannot be set - type 'Svg.Pathing.SvgPathSegmentList' cannot convert from string ''.).

I'm not at the machine where I can test this, but I'll try to look into this when I am getting back at my PC and see if I can reproduce this and try to pinpoint what exactly goes wrong.

@gvheertum commented on 6 Sep 2017:
Checking the SVG source I noticed:
```html
<path class="plot-now" d="" style="opacity: 0;"></path>
```
[...]

PR #418, opened 2019-02-19
#399 lexer error isolation

gvheertum:
#399 Testability for the hex and !imporant combination issue in the lexer

Isolated the !important issue when working with hex values into a testcase. The specific case still fails but is now at least isolated into a test and can be further investigated and fixed.

Also fixed an issue in the HEX parser where hexvalues with a of char count of < 3, > 3  && < 6 and > 6 were ignored, these now yield an exception (the incorrect values caused some issues along the way, so having them fail is much cleaner in my opinion)


Source/External/ExCSS/Parser.Blocks.cs
```diff
@@ -464,23 +464,23 @@ private bool ParseValuePool(Block token)
 
         private bool ParseHexValue(Block token)
         {
-            switch (token.GrammarSegment)
+			switch (token.GrammarSegment)
             {
```
mrbean-bremen: I would prefer to use spaces everywhere instead of tabs (e.g adapt the editor settings) - tabs tend to introduce formatting issues due to different tab size. Also, it introduces unintented changes like this one.
Currently, there is still a mix of tabs and spaces, but mostly spaces are used (I think tabs have been introduced by some PRs).

Source/External/ExCSS/Parser.Blocks.cs
```diff
@@ -554,7 +554,7 @@ private bool ParseSingleValueHexColor(string color)
 
             if(HtmlColor.TryFromHex(color, out htmlColor))
                 return AddTerm(htmlColor);
-            return false;
+			throw new ArgumentOutOfRangeException(nameof(color), $"{color} is not a valid hex color");
         }
```
mrbean-bremen: if we raise here, we have to handle the excption in the caller code, and use the default color instead. Or we may do the conversion here, if we have enough information (not sure, I haven't checked the code), and issue a warning instead of the exception.

Tests/Svg.UnitTests/LexerIssueTests.cs
```diff
@@ -0,0 +1,92 @@
...
+		{
+			var str = GetResourceXmlDocAsString(GetFullResourceString(ResourceStringLexerTemplate));
+			str = str.Replace(LexerTemplateReplaceToken, testContent);
+			var xml = GetXMLDocFromString(str);
+			var doc = OpenSvg(xml);
+			return doc;
+		}
+
+	}
+}
```
mrbean-bremen: This is nice! Maybe we can disable the tests as long as the bug is not fixed and merge it in - this makes it easier for someone else to have a go at it.

Source/External/ExCSS/Parser.Blocks.cs
mrbean-bremen: Hm, I thought I had already explained why I don't want to have the exception here, but can't find it now. Just in case it got lost, or I actually didn't click the Comment button: 
We usually try to behave like browsers, and browsers in this case (e.g. incorrect color description) just ignore the color and use the default. I recently fixed a bug along these lines. So we just haveto make sure that the default color is used, if the provided color is invalid.

Tests/Svg.UnitTests/SvgTestHelper.cs
```diff
@@ -231,8 +264,9 @@ protected virtual void CheckSvg(SvgDocument svgDoc, Image img)
             using (var ms = new MemoryStream())
             {
                 img.Save(ms, ImageFormat.Png);
+				img.Save($"d:\\temp\\{Guid.NewGuid().ToString()}.png");
```
mrbean-bremen: This line was probably for debugging - it leads to a few failing tests because of missing drive.
If removing it, all tests, but the one that should fail pass.

Source/External/ExCSS/Parser.Blocks.cs
gvheertum: Fixed, my default was set to tabs and I in this version I did not have the whitespace characters visible, so I didn't notice. Is there a way to set this up on a project/solution level?

Source/External/ExCSS/Parser.Blocks.cs
gvheertum: I reverted the change and changed my unit test, however if I feed an incorrect hex value the lexer goes haywire somewhere else (test: Lexer_FileWithInvalidHex_ColorTagIsIgnored)

```System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index```

I think that's why I introduced the "nicer" exception, since the lexer did crash anyway :D, so why not fail with a nice and descriptive exception. For now I flagged the test inconclusive to pass the build.

Tests/Svg.UnitTests/SvgTestHelper.cs
gvheertum: Whoops, good catch. I did that to see the output results of some of the tests and did not spot that debug code. Removed it from the code.

Tests/Svg.UnitTests/LexerIssueTests.cs
gvheertum: Flagged them as inconlusive for now

gvheertum:
@mrbean-bremen Thanks for the feedback. Pushed some updates, as far as I can see, the new commits should fix the issues addressed by you.

gvheertum:
> Looks good now, thanks - I will merge now!
:)

I'm still a bit puzzled about what happens in test Lexer_FileWithInvalidHex_ColorTagIsIgnored by the way, since the parser seems to really dislike invalid Hex codes, so maybe we should find a solution for that (perhaps the default color), because returning false (as done in the old code and after the revert), will put the parser somehow in an invalid state causing a crash later-on.

(but maybe this is something for a different ticket)

mrbean-bremen:
Yes - I agree. I don't know the lexer code, so I'm sure about a solution yet, but may have a look later this week. I recently had a first glance at the parser, to fix another problem, but haven't seen the lexer so far at all...
Well, with your tests in place this may be easier now. 
If we won't handle the invalid color problem in this issue, we may write another issue later.

PR #423, opened 2019-02-21
Added handling of invalid color as default color

mrbean-bremen:
- adapted test to check for default value
- see #399

### DIFF
--- a/Source/External/ExCSS/Parser.Blocks.cs
+++ b/Source/External/ExCSS/Parser.Blocks.cs
@@ -482,8 +482,8 @@ namespace Svg.ExCSS
                     break;
             }
 
-            ParseSingleValueHexColor(_buffer.ToString());
             SetParsingContext(ParsingContext.InSingleValue);
+            ParseSingleValueHexColor(_buffer.ToString());
             return ParseSingleValue(token);
         }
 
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
@gvheertum - can you please review this? This does not fix the actual bug, but the related one with invalid color.

gvheertum:
@mrbean-bremen looks good! Also like what you did with the test to expand the checks performed (not only testing if no error occurs, but also checking the working). The change to the template is also a good one, I never really looked into it in-depth since I was looking for a crash/non-crash scenario and never did a cleaning of the file :)

mrbean-bremen:
Ok, thanks - here comes the second commit :)
This one is smaller...

mrbean-bremen:
@gvheertum - if you are ok with the second commit (which will fix your initial issue) I will merge this.
This is a very small fix, but it took me an hour to get there...

gvheertum:
> Ok, thanks - here comes the second commit :)
> This one is smaller...

Wow... Sometimes it's just "that simple", switching two lines ;) ... Sweet! I kept landing in the Hex parsing, but never was able to pin-point the exact problem. 

Great find! Glad you were able to fix it!

mrbean-bremen:
Ok, thanks - it was easier having your tests in place :) Merging this then..

PR #425, opened 2019-02-26
Took SvgDocument exclusion out of ApplyRecursive and optmized travers…

kbiel:
There appears to be no reason to limit `ApplyRecursive` when an `SvgDocument` is the `SvgElement` parameter. The one place it is used already checks and does not call `ApplyRecursive` on an `SvgDocument`. The same question applies for `ApplyRecursiveDepthFirst` as it is only called from `SvgElementCollection` and an `SvgDocument` is never a member of an instance of `SvgElementCollection`.

I also optimized the traversal for speed and to remove the possibility of a stack overflow.

### DIFF
--- a/Source/SvgExtentions.cs
+++ b/Source/SvgExtentions.cs
@@ -78,28 +78,52 @@ namespace Svg
 
         public static void ApplyRecursive(this SvgElement elem, Action<SvgElement> action)
         {
-            action(elem);
-
-            if (!(elem is SvgDocument)) //don't apply action to subtree of documents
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
Hm, looks ok for me at first glance, but I don't know that code, so maybe I'm overloking something.
@tebjan - you added the respective change in [this commit](https://github.com/vvvv/SVG/commit/2bfed054464c08b7a3cc28442a51fe5294632709), do you remember why it was made?

PR #428, opened 2019-03-06
Fix: character corruption.

H1Gdev:
- use FromSvg method.

### Sample SVG
```svg
<?xml version="1.0" encoding="shift_jis"?>
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0" y="0" width="80" height="400" xml:space="preserve">
<text x="10" y="30" fill="#000" font-size="14">あABC</text>
</svg>
```

### Character corruption
issue_output

### DIFF
--- a/Samples/SVGViewer/SvgViewer.cs
+++ b/Samples/SVGViewer/SvgViewer.cs
@@ -44,11 +44,8 @@ namespace SVGViewer
         {
             try
             {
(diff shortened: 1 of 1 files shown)


H1Gdev:
Thanks for advice me:thumbsup:
It's just test.

mrbean-bremen:
That wasn't an advice - more of a reminder to myself for the future ;)

PR #438, opened 2019-04-23
Fix #437 'Doesn't render stroke of rect, if fill opacity is 0' issue.

### LINKED ISSUE #437: Doesn't render stroke of rect, if fill opacity is 0
I realized, that this SVG lib doesn't render the border of rectangles, when the fill opacity is set to 0:
```svg
<rect id="SvgjsRect1008" width="313" height="448" stroke-opacity="1" stroke="#ff0d28" stroke-width="3" fill-opacity="1" fill="rgba(0, 0, 0, 0)" x="360" y="48" />
```

This example will render the stroke in the same opacity as the fill color. So it's not rendered at all in this example. This, as far as I know, doesn't correspond to the SVG standard.

H1Gdev:


NMGo:
I think this PR does not fully address the underlying issue #437. Please refer to my comment in that issue.

mrbean-bremen:
@H1Gdev - thanks for the PR! I'm a bit short on time today, so I will probably do the review tomorrow. Maybe you can have a look at the other mentioned problem and check if it is related...

H1Gdev:
@NMGo 

Thanks for your report !
I check your new issue.

PR #433, opened 2019-04-10
Fix: opacity attribute.

H1Gdev:
### Sample SVG
```svg
<?xml version="1.0" encoding="utf-8"?>
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" width="320" height="320">

<!-- rect -->
<rect x="40"  y="40" width="40" height="40" style="opacity:0.5;fill:red;stroke:black;" />
<rect x="100" y="40" width="40" height="40" style="opacity:0.5;fill:red;stroke:black;stroke-width:10;" />
<rect x="160" y="40" width="40" height="40" style="opacity:0.5;fill:red;stroke:black;stroke-width:10;stroke-opacity:0.8;" />
<rect x="220" y="40" width="40" height="40" style="opacity:0.5;fill:red;stroke:black;stroke-width:10;stroke-opacity:0.8;fill-opacity:0.8;" />

<!-- circle -->
<circle cx="60"  cy="120" r="20" style="opacity:0.5;fill:red;stroke:black;" />
<circle cx="120" cy="120" r="20" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;" />
<circle cx="180" cy="120" r="20" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;stroke-opacity:0.8;" />
<circle cx="240" cy="120" r="20" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;stroke-opacity:0.8;fill-opacity:0.8;" />

<!-- ellipse -->
<ellipse cx="60"  cy="180" rx="16" ry="20" style="opacity:0.5;fill:red;stroke:black;" />
<ellipse cx="120" cy="180" rx="16" ry="20" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;" />
<ellipse cx="180" cy="180" rx="16" ry="20" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;stroke-opacity:0.8;" />
<ellipse cx="240" cy="180" rx="16" ry="20" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;stroke-opacity:0.8;fill-opacity:0.8;" />

<!-- polygon -->
<polygon points="50,220  70,220  50,250  70,240"  style="opacity:0.5;fill:red;stroke:black;" />
<polygon points="110,220 130,220 110,250 130,240" style="opacity:0.5;fill:red;stroke:black;stroke-width:8;" />
[...]

PR #443, opened 2019-04-26
Fix: Dispose Graphics and Image in SvgRenderer.

H1Gdev:
Dispose Graphics and/or Image in SvgRenderer if needed.

### DIFF
--- a/Source/Rendering/SvgRenderer.cs
+++ b/Source/Rendering/SvgRenderer.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Drawing;
 using System.Drawing.Drawing2D;
 using System.Drawing.Imaging;
@@ -12,10 +10,13 @@ namespace Svg
     /// <summary>
     /// Convenience wrapper around a graphics object
     /// </summary>
-    public sealed class SvgRenderer : IDisposable, IGraphicsProvider, ISvgRenderer
+    public sealed class SvgRenderer : ISvgRenderer, IGraphicsProvider
     {
-        private Graphics _innerGraphics;
-        private Stack<ISvgBoundable> _boundables = new Stack<ISvgBoundable>();
+        private readonly Graphics _innerGraphics;
+        private readonly bool _disposable;
+        private readonly Image _image;
+
+        private readonly Stack<ISvgBoundable> _boundables = new Stack<ISvgBoundable>();
 
         public void SetBoundable(ISvgBoundable boundable)
         {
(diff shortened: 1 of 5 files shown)


Source/SvgDocument.cs
```diff
@@ -545,21 +539,35 @@ public virtual Bitmap Draw(int rasterWidth, int rasterHeight)
...
+            this.RasterizeDimensions(ref bitmapSize, rasterWidth, rasterHeight);
 
             if (bitmapSize.Width == 0 || bitmapSize.Height == 0)
                 return null;
 
-            var bitmap = new Bitmap((int)Math.Round(bitmapSize.Width), (int)Math.Round(bitmapSize.Height));
+            Bitmap bitmap = null;
             try
             {
-                var renderer = SvgRenderer.FromImage(bitmap);
```
H1Gdev: Do not Dispose at latest version.

Source/Rendering/SvgRenderer.cs
```diff
@@ -139,13 +152,14 @@ public static ISvgRenderer FromImage(Image image)
         /// <param name="graphics">The <see cref="Graphics"/> to create the renderer from.</param>
         public static ISvgRenderer FromGraphics(Graphics graphics)
         {
-            return new SvgRenderer(graphics);
+            return new SvgRenderer(graphics, false);
```
H1Gdev: Do not Dispose.

Source/SvgDocument.cs
```diff
@@ -461,46 +466,43 @@ public void Draw(Graphics graphics, SizeF? size)
                 throw new ArgumentNullException("graphics");
```
H1Gdev: What is this method used for ?
Not `SizeF` but `RectangleF` ?

mrbean-bremen: see #337

mrbean-bremen:
@H1Gdev - is this ready for merge? Can you describe the problem you solved?

H1Gdev:
@mrbean-bremen 

I analyzed source codes with focus on SvgRenderer at after #433.

I fixed it because it is an instances of IDisposable but not call Dispose method.
I checked that called Dispose when call SvgDocument.Draw().

Source/SvgDocument.cs
H1Gdev: @mrbean-bremen 

Thanks!!

I test following source code.
```cs
var imageSize = svgDoc.GetDimensions();

// same size
using (var image = new Bitmap((int)imageSize.Width, (int)imageSize.Height))
using (var graphics = Graphics.FromImage(image))
{
    var size = new SizeF(imageSize.Width, imageSize.Height);
    svgDoc.Draw(graphics, size);
}

// size up
using (var image = new Bitmap((int)imageSize.Width, (int)imageSize.Height))
using (var graphics = Graphics.FromImage(image))
{
    var size = new SizeF(imageSize.Width * 2, imageSize.Height * 2); 
    svgDoc.Draw(graphics, size);
}

// size down
using (var image = new Bitmap((int)imageSize.Width, (int)imageSize.Height))
using (var graphics = Graphics.FromImage(image))
{
    var size = new SizeF(imageSize.Width / 2, imageSize.Height / 2);
    svgDoc.Draw(graphics, size);
}
```
The results were all the same.

mrbean-bremen: ok, that doesn't seem right...

PR #444, opened 2019-04-29
Fix: boundary with corner and stroke.

H1Gdev:
### Sample SVG
```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" width="400" height="180">
  <rect x="50" y="20" rx="20" ry="20" width="150" height="150"
  style="fill:red;stroke:black;stroke-width:4;opacity:0.5" />
</svg>
```

### Result
output

### Fixed this issue
output

Not correct boundary if corner is present.


Source/Basic Shapes/SvgRectangle.cs
```diff
@@ -123,7 +123,7 @@ public override GraphicsPath Path(ISvgRenderer renderer)
                 }
 
                 // If the corners aren't to be rounded just create a rectangle
-                if (CornerRadiusX.Value == 0.0f && CornerRadiusY.Value == 0.0f)
+                if (renderer == null || (CornerRadiusX.Value == 0.0f && CornerRadiusY.Value == 0.0f))
```
mrbean-bremen: nice catch!

Issue #427, opened 2019-03-03
Svg on opening loads nothing in .NET Core 2.2


thedonciuxx:
Hi, I am using the .NET Core 2.2 and am using the latest NuGet version of the SVG Package (2.4.2),

I am getting no errors when running the code like I saw on other issue reports, but for me the svg's are "empty". I'm trying to load / open an svg and drawing it, but it draws nothing.

I modified my code to save each loaded svg to a png image, this is the result: (The color is black, but that's a windows explorer thing, they're actually empty, but with defined size)
![explorer_2019-03-03_11-55-44](https://user-images.githubusercontent.com/13348029/53695913-0babb000-3dca-11e9-84ad-4e98c4a001de.png)

mrbean-bremen:
.NET Core is not yet supported in 2.4.2 (as you probably know, as you have commented under the [respective PR](https://github.com/vvvv/SVG/pull/415)), but hopefully will be in soon.

thedonciuxx:
Ah, so you've seen my comment... Still it's been a really long time since the original author of the pr made any changes, I really hope the conflicts get resolved and bugs get fixed soon to merge! 😄

mrbean-bremen:
Yeah, I hope so, too... maybe I will resolve the conflicts myself, but I really would like to get a response from the author first.
This issue is basically a duplicate of #400 and #248, hopefully to be resolved soon.

thedonciuxx:
I don't know if I should open a different issue or update this one, but the build which should support the .NET Core 2.2 is still in need of bug fixing...

I ran the published linux app on a Ubuntu emulator (VirtualBox to be specific) and got this error:
```
System.NullReferenceException: Object reference not set to an instance of an object.
at Svg.SvgDocument.Open[T](String path, Dictionary`2 entities)
```

I know that it loads the file, because one time I forgot to add it and it said that file wasn't found (or something like that can't remember exactly)

mrbean-bremen:
Well, this is work in progress, and l haven't done anything apart from conflict resolving and getting the AppVeyor tests running again. I planned to have a closer look this weekend, but I caught a cold and am not really up to it at the moment.
Also, the CI tests are run under Windows only. It would certainly make sense to add Travis.CI tests for Linux and Mac OS. An, help is welcome...

mrbean-bremen:
So I mostly got rid of the cold and may have another look at this somewhere in the next days, but I know nothing about using .NET Core under Linux. Can you give me some hints about your usage?
[...]

PR #450, opened 2019-05-09
Stub for non-Windows DPI; possible fix for #427

chucker:


mrbean-bremen:
Ok, I cleaned up the commits a bit and will merge this in a moment.

mrbean-bremen:
You may need to re-branch from master now, as I force-pushed to your branch - sorry for that...

mrbean-bremen:
And thanks a lot for the fix - any help is welcome! 
Specifically, any help to test/check the stuff under Linux/MacOS would help, for example, setting up CI tests for Linux. Generally, any help to reduce the number of open issues, to add documentation, or to improve platform support is appreciated.

chucker:
Anything to help move .NET Standard support forward I benefit from, so… :-)

I’ll have to read up on how the AppVeyor stuff works.

mrbean-bremen:
> I’ll have to read up on how the AppVeyor stuff works. I assume I could theoretically make pull requests to that config to add Linux CI builds?

Yes, of course! You can setup AppVeyor for your own repo, so you can test it in your own branch.
I haven't used AppVeyor for Linux so far (used Travis.CI for that, but that have been Python projects), but there seems to be [Linux support](https://www.appveyor.com/docs/getting-started-with-appveyor-for-linux/) specifically for .NETCore.

arendvw:
Just tested this build on debian buster. I still needed `libgdiplus` installed from the mono repositories for this to work. Is this to be expected?

mrbean-bremen:
Ok, good to know that!

gvheertum:
> Just tested this build on debian buster. I still needed [`libgdiplus`](https://github.com/mono/libgdiplus) installed from the mono repositories for this to work; but it did work. Is this to be expected?
> 
> **Update:** It seems this is required for [System.Drawing under linux](https://github.com/dotnet/core/issues/2746), and is not a dependency of this project.

MacOs has a similar issue, when running the tests an error is thrown regarding the libgdiplus libraries not available. For MacOs there is a possibility to stub the libgdiplus by including the `runtime.osx.10.10-x64.CoreCompat.System.Drawing` Nuget (see: https://github.com/CoreCompat/libgdiplus-packaging).

I'm not sure if there is an option to have specific Nuget settings for specific configurations (since this is most likely not to work on Windows or Linux), but to get the package up and running on MacOs I think having the CoreCompat package might be the most usable option.

mrbean-bremen:
Thanks for digging into this! I'm not aware of a way for OS-specific Nuget settings, but we need at least the instructions how to install the package under Linux and MacOS somewhere with the package - probably with the package description.
@tebjan - what do you think?

chucker:
Haven't tried, but something like this might work:

```xml
  <PropertyGroup Condition="'$(RuntimeIdentifier)'=='osx.10.10-x64'">
    <PackageReference>runtime.osx.10.10-x64.CoreCompat.System.Drawing</PackageReference >
  </PropertyGroup>
```

It _might_ also be that you don't have to do any of that and can just add the reference regardless (it might just be a noop on other platforms).

mrbean-bremen:
> It might also be that you don't have to do any of that and can just add the reference regardless (it might just be a noop on other platforms).

Definitely not - just tried it and got it working only after manually deleting all object files and restarting the solution. 
Anyway, we still would need documentation for linux.

gvheertum:
> > It might also be that you don't have to do any of that and can just add the reference regardless (it might just be a noop on other platforms).
> 
> Definitely not - just tried it and got it working only after manually deleting all object files and restarting the solution.
> Anyway, we still would need documentation for linux.

Yeah, I'm also not sure about having this kind of stuff included by default. This might be very breaking. I've been playing a bit with the settings, but haven't got it to work yet (it seems to not pickup the RuntimeIdentifier somehow).

From what I've read until now, the MacOs package needs to be included in the SVG library and needs to be taken all the way up to the executable/library consuming the SVG component (in that way it differs from the Linux version which seems to install the gdi libs in your global scope). MacOs with the CoreCompat package will put the stubs together with the project and apparently it will "auto resolve" to the CoreCompat implementation for Mac instead of letting the framework figure it out (which will result in errors on the Mac).

I know that you can setup NuGet packages to have different content/deploys for different frameworks and versions of the framework, but I'm not sure if you can do platform specific stuff. If that's not possible the users somehow need to get the gdi stuff global on their system or build their own binaries (with additional instructions).

One of the options (which would be more complex) is to put a proxy between everything gdi related coming out of the SVG (for example, I've seen it break on Matrix) and check if the GDI is available, if not, load the alternative route (for Mac and Linux), but I'm afraid that might be quite a big change.

mrbean-bremen:
I think for the time being some simple instructions will do (like the ones in your link - I copied them over to the [NetCore 2.2 PR discussion](https://github.com/vvvv/SVG/pull/495#issuecomment-505591582)). Anything more complicated may be added later, though I think your last option (proxying GDI access) would be a bit of overkill.

Issue #452, opened 2019-05-10
SvgFontManager fails to initialize if multiple fonts of the same name exist


chucker:
(This _may_ be a platform-specific quirk in .NET Standard. It's not clear to me from the description of https://docs.microsoft.com/en-us/dotnet/api/system.drawing.fontfamily.families?view=netframework-4.8 what happens if two fonts of the same name exist — it could be that this isn't possible in Windows but is in macOS.)

The type initializer in `SvgFontManager` creates a dictionary by font name. If names aren't unique, this will throw:

```
System.ArgumentException: "An item with the same key has already been added. Key: hiragino sans gb"
  at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)\n   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)\n   at System.Linq.Enumerable.ToDictionary[TSource,TKey](TSource[] source, Func`2 keySelector, IEqualityComparer`1 comparer)\n   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector, IEqualityComparer`1 comparer)\n   at System.Linq.Enumerable.ToDictionary[TSource,TKey](IEnumerable`1 source, Func`2 keySelector)\n   at Svg.SvgFontManager..cctor() in /Users/chucker/Repositories/SVG-448-netcore/Source/SvgFontManager.cs:22
```

This breaks the test `LargeEmbeddedImageTest` and possibly others on my machine.

mrbean-bremen:
Reading up about font names, it looks as if is allowed to have several fonts with the same name as long as they have different weight and/or style. 
As far as I understand the code, it doesn't matter if there is more than one font for a name, as the font name is only used to lookup the font family, so maybe your workaround is already sufficient to resolve this issue.

mrbean-bremen:
Actually, this has been introduced with the recent introduction of the font manager - before that `FirstOrDefault` has been used. So this is probably a regression...

mrbean-bremen:
@chucker - I think this can be closed. As I wrote, I do not consider your fix a workaround, as it shouldn't matter for the usage if there is more than one font for the name.

PR #453, opened 2019-05-10
Fix non-unique font names

chucker:
See #452.

This is only a workaround. However, given that `FindFont` compares by name, we probably need to do something like this. (We might be able to better filter/order which fonts are "valid" or most likely to be the one referenced in the SVG?)


chucker:
Hmmm, maybe I should have made the pull request to `netcore` instead. Is anything in particular left to do for the `netcore` branch to be closed?

Source/Document Structure/SvgGroup.cs
```diff
@@ -56,7 +56,7 @@ protected override void Render(ISvgRenderer renderer)
         }
 
         /// <summary>
-        /// Gets the <see cref="GraphicsPath"/> for this element.
+        /// Gets the <see cref="System.Drawing.Drawing2D.GraphicsPath"/> for this element.
```
H1Gdev: Is this change unnecessary ?

chucker:
(Will do when I get around to it; maybe this weekend)

mrbean-bremen:
Sure, thanks!

mrbean-bremen:
@chucker - if you don't mind, I will just incorporate your fix manually to have it in for the next release, as it is actually a small change.

chucker:
Sure, that’s fine!

mrbean-bremen:
Thanks!

PR #454, opened 2019-05-11
Supports various formats of URL string.

H1Gdev:
- Absolute path.

```svg
filter="url(file:///absolute/path/to/definitions.svg#filter1)"
```

```svg
filter="url(https://absolute/path/to/definitions.svg#filter1)"
```

- Enclose in **'** or **"**.

```svg
filter="url('./definitions.svg#filter1')"
```

```svg
filter='url("./definitions.svg#filter1")'
```

### DIFF
--- a/Source/SvgElementIdManager.cs
+++ b/Source/SvgElementIdManager.cs
@@ -1,10 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Text.RegularExpressions;
 using System.Net;
-using System.IO;
+using System.Text.RegularExpressions;
 
 namespace Svg
 {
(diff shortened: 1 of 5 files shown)


Source/SvgElementIdManager.cs
```diff
@@ -43,31 +34,59 @@ public virtual SvgElement GetElementById(string id)
...
+            var urlString = GetUrlString(uri.ToString());
+
+            if (!urlString.StartsWith("#"))
             {
-                var fullUri = new Uri(this._document.BaseUri, uri);
-                var hash = fullUri.OriginalString.Substring(fullUri.OriginalString.LastIndexOf('#'));
-                SvgDocument doc;
-                switch (fullUri.Scheme.ToLowerInvariant())
+                var index = urlString.LastIndexOf('#');
+                var fragment = urlString.Substring(index);
```
mrbean-bremen: Is it guaranteed at this point that urlString contains "#"? Otherwise, it would throw here.

H1Gdev: Thank you for your review.

There are various formats of id in GetElementById methods, so same as old code.

mrbean-bremen: Ok, I see now that this has not changed - it the URL was in wrong format (e.g. missing #), it would have crashed before, too.

mrbean-bremen:
Thanks for the contribution!

PR #458, opened 2019-05-18
Auto-start-reverse support

francois-dorin:
A simple modification to allow markers with 'auto-start-reverse' orientation (See [W3C "marker" element specification](https://www.w3.org/TR/svg-markers/#MarkerElement) for more informations).

I realized an implementation which do not break compatibility with existing codes. To use it, I introduced a 2 parameters constructor to `SvgOrient `class:
`public SvgOrient(bool isAuto, bool isAutoStartReverse)`

The first boolean allow to enable 'auto' mode, and the second to determine if it is 'auto' (false) or 'auto-start-reverse' (true).

`isAutoStartReverse `has no effect if `isAuto == false`.

Example of usage : 
```CSharp
SvgMarker svgMarker = new SvgMarker()
{
   ID = "arrow",
   ViewBox = new SvgViewBox(0, 0, 10, 10),
   RefX = 10,
   RefY = 5,
   MarkerWidth = 6,
   MarkerHeight = 6,
   Orient = new SvgOrient(true, true)
};
```


Source/DataTypes/SvgOrient.cs
```diff
@@ -82,7 +100,7 @@ public override string ToString()
             string type = string.Empty;
 
 			if (this.IsAuto)
-				return "auto";
+				return this.IsAutoStartReverse ? "auto-start-reverse" : "auto";
```
mrbean-bremen: This is actually not in the SVG 1.1 spec, which is used in the lib until now (and the support is not complete by far). I think it is a good idea to add later additions (e.g. SVG 2) to the library, though it would be nice to have an example image for a regression test. Our current W3C test images are for SVG 1.1.
@tebjan - what do you think about SVG 2 support? Maybe we shall a SVG 2 test suite?

Source/DataTypes/SvgOrient.cs
```diff
@@ -11,11 +11,13 @@ namespace Svg
...
     {
         private bool _isAuto = true;
+        private bool _isAutoStartReverse = false;
         private float _angle;
   
 		public SvgOrient()
 		{
 			IsAuto = false;
+            IsAutoStartReverse = false;
             Angle = 0;
```
mrbean-bremen: Hm, I was sure I had already commented here - the mentiuoned formatting issue is the space/tab mix here (which has been there before, so you didn't introduce it), so it would be nice if you could just replace the tabs with spaces in this file.

francois-dorin: Thank you for your feedback and for accepting my contribution :)
I will fix this little formatting issue. Svg/png images will also be added to the test suite.
And I will complete the implementation to support both rendering and loading correctly this new value (my first proposal only take into account my use case for generating svg image).

Source/DataTypes/SvgOrient.cs
```diff
@@ -11,23 +11,31 @@ namespace Svg
...
+            IsAuto = isAuto;
+            IsAutoStartReverse = isAutoStartReverse;
+        }
+
         public SvgOrient(float angle)
-		{
-			Angle = angle;
-		}
+        {
+        Angle = angle;
```
francois-dorin: A remaining formatting issue . I will correct it rapidly.

Source/DataTypes/SvgOrientConverter.cs
```diff
@@ -22,6 +22,8 @@ public override object ConvertFrom(ITypeDescriptorContext context, System.Global
 			{
 				case "auto":
 					return (new SvgOrient(true));
+                case "auto-start-reverse":
```
francois-dorin: Another formatting issue, but in this case, it's because I used spaces and not tabs!

Source/DataTypes/SvgOrient.cs
francois-dorin: I did not find any information about the fact that the lib support only the SVG 1.1 spec. Maybe it can be added in the README ?

mrbean-bremen: Good point. I guess that at the time the library was created by Microsoft, this has been the latest spec, but a mention in the README would certainly help.

mrbean-bremen:
Looks like you are done here - thanks! Is that correct?

francois-dorin:
> Looks like you are done here - thanks! Is that correct?

It is correct ;)

mrbean-bremen:
Thanks for the contribution :)

Issue #220, opened 2016-03-02
Image tag with data:image/svg not implemented


Cian-Chambliss:
for image tags of the form

<image xlink:href="data:image/svg+xml,..>

and

<image xlink:href="data:image/svg+xml;base64,..>

an exception is thrown by image.FromString(), which is expecting a bitmap image, not vector graphics.

Fixed this by changing the GetImage() method in SvgImage.cs,  first test the mime type portion of the message for image/svg, then get the SVG string & create an SVG Document.

```
    protected object GetImage(Uri uri)
    {
        try
        {
            // handle data/uri embedded images (http://en.wikipedia.org/wiki/Data_URI_scheme)
            if (uri.IsAbsoluteUri && uri.Scheme == "data")
            {
                string uriString = uri.OriginalString;
                int dataIdx = uriString.IndexOf(",") + 1;
                if (dataIdx <= 0 || dataIdx + 1 > uriString.Length)
                    throw new Exception("Invalid data URI");
                string imageType = uriString.Substring(0, dataIdx); // Check for SVG embedded in the data...
                if (imageType.Contains("image/svg"))
                    {
                        string embeddedSVG;
                        if (imageType.Contains("base64") )
                        {
                            embeddedSVG = System.Text.Encoding.UTF8.GetString( Convert.FromBase64String(uriString.Substring(dataIdx)) );
                        }
                        else
                        {
                            embeddedSVG = uriString.Substring(dataIdx);
                        }
                        using (var stream = new MemoryStream(System.Text.UTF8Encoding.Default.GetBytes(embeddedSVG)))
                        {
                            var doc = SvgDocument.Open<SvgDocument>(stream);
                            doc.BaseUri = uri;
                            return doc;
                        }
                    }
                else
                    {
                        // we're assuming base64, as ascii encoding would be *highly* unsusual for images
                        // also assuming it's png or jpeg mimetype
                        byte[] imageBytes = Convert.FromBase64String(uriString.Substring(dataIdx));
                        using (var stream = new MemoryStream(imageBytes))
                        {
                            return Image.FromStream(stream);
                        }
                    }
            }

            if (!uri.IsAbsoluteUri)
            {
                uri = new Uri(OwnerDocument.BaseUri, uri);
            }

            // should work with http: and file: protocol urls
            var httpRequest = WebRequest.Create(uri);

            using (WebResponse webResponse = httpRequest.GetResponse())
            {
                using (var stream = webResponse.GetResponseStream())
                {
[...]

Issue #71, opened 2014-05-28
data-uri schemes do not support SVG data-uris


rwdalpe:
Commit https://github.com/planetclegg/SVG/commit/6f53360e477d3b9b72f09e2b862ca1baedf29520 added basic support for data-uri schemes for the image tag. It is also possible for an SVG to be embedded via data-uri as well.

For example, shown both encoded and not encoded (obviously only the encoded version would be valid)

```
xlink:href='data:image/svg+xml;charset=utf-8,&lt;?xml version='1.0' encoding='UTF-8' ?&gt;&lt;svg ...

xlink:href='data:image/svg+xml;charset=utf-8,<?xml version='1.0' encoding='UTF-8' ?><svg ...
```

This should be functionally the same as specifying a path to an SVG document containing that content.

mrbean-bremen:
@rwdalpe - this is quite old, did you end up adding that support?

rwdalpe:
Unfortunately we wound up not having time to implement this. Sorry for
leaving this thread hanging for so long.

mrbean-bremen:
Thanks - just cleaning up old issues. I haven't really looked into this one, so I leave it open for now.

H1Gdev:
Unescaped **<**, **>**, etc. not allowed in attributes values.

Error on Firefox, Chrome, Edge.

PR #462, opened 2019-05-20
Support for embedded SVG in data URI.

H1Gdev:
ref. #71, and #220.


Source/Basic Shapes/SvgImage.cs
```diff
@@ -240,78 +224,50 @@ protected override void Render(ISvgRenderer renderer)
...
-            {
-                //Uri MaxLength is 65519 (https://msdn.microsoft.com/en-us/library/z6c2z492.aspx)
-                safeUriString = uriString.Substring(0, 65519);
-            }
-            else
-            {
-                safeUriString = uriString;
-            }
+            // Uri MaxLength is 65519 (https://msdn.microsoft.com/en-us/library/z6c2z492.aspx)
+            // main use is data URI scheme, but very long to use it as data URI scheme.
```
mrbean-bremen: I overlooked this one - could you please rephrase this in the other PR? I'm not sure I understand the meaning, but I guess it means that a very long URI may happen if using the data URI schema?

H1Gdev: @mrbean-bremen 

Thanks for your review.
That's right.

PR #464, opened 2019-05-21
Update SVG Viewer.

H1Gdev:
- Unsubscribe from TextChanged event so that load SVG once.
- Small refactoring.

### DIFF
--- a/Samples/SVGViewer/SvgViewer.cs
+++ b/Samples/SVGViewer/SvgViewer.cs
@@ -22,12 +22,20 @@ namespace SVGViewer
                     var svgDoc = SvgDocument.Open(openSvgFile.FileName);
                     RenderSvg(svgDoc);
 
-                    var xmlDoc = new XmlDocument
+                    textBox1.TextChanged -= TextBox1_TextChanged;
+                    try
                     {
-                        XmlResolver = null
-                    };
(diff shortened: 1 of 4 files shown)


Samples/SVGViewer/SvgViewer.cs
```diff
@@ -22,12 +22,20 @@ private void Open_Click(object sender, EventArgs e)
                     var svgDoc = SvgDocument.Open(openSvgFile.FileName);
                     RenderSvg(svgDoc);
 
-                    var xmlDoc = new XmlDocument
+                    textBox1.TextChanged -= new EventHandler(TextBox1_TextChanged);
```
mrbean-bremen: I don't understand this - how do you unsubscribe from a handler you just created?
Can you please elaborate a bit what this PR does?

H1Gdev: @mrbean-bremen 

Thank you for your review.

This event handler has already been registered in Designer.
Update text on line 35 of updated file.
Temporarily unsubscribe TextChanged event handler so that SVG is not loaded again by TextChanged event.

mrbean-bremen: Ah, thanks - now it makes sense to me!

Issue #138, opened 2015-03-18
linearGradient not exported correctly if id begins with number


epuig-ssinergie:
When saving SvgDocument to PNG, the gradient area is rendered as solid black.

After several tests, it seems to be directly related to the linearGradient which is beginning with a number, eg:
<linearGradient id="41cb5de8-82e7-4ab5-94d7-87e10775c848"

![ko2](https://cloud.githubusercontent.com/assets/7677220/6708440/8bb2f62e-cd70-11e4-9107-c9e15233c195.png)

As a workaround i'm just adding a letter at the beginning of id and it works perfectly.

![ok1](https://cloud.githubusercontent.com/assets/7677220/6708439/8a4c96e6-cd70-11e4-8700-d4310b15a53e.png)

Thanks

epuig-ssinergie:
same issue with <radialGradient> elements

mrbean-bremen:
According to W3C, an id name cannot start with a number, so this is not a valid SVG - not supporting this is not a bug.
Is there any standard application that generates this kind of ids? In this case, it may make sense to support this nonetheless.

epuig-ssinergie:
svg was generated using ASP.net Telerik library with getSVGString()
https://www.telerik.com/support/code-library/exporting-radhtmlchart-to-png-and-pdf

mrbean-bremen:
Ok, thanks - while this is basically a bug in the converter, it would make sense to handle this in that case.

PR #465, opened 2019-05-21
Resolve: #138 'linearGradient not exported correctly if id begins with number' issue.

H1Gdev:
According to W3C, an id name cannot start with a full stop (.) character, and a hyphen-minus (-) character, too.
But check only number.

So simply delete number check.

### DIFF
--- a/Source/SvgElementIdManager.cs
+++ b/Source/SvgElementIdManager.cs
@@ -153,15 +153,6 @@ namespace Svg
                 return id;
             }
 
-            if (char.IsDigit(id[0]))
-            {
-                if (autoForceUniqueID)
-                {
-                    return EnsureValidId("id" + id, true);
-                }
-                throw new SvgIDWrongFormatException("ID cannot start with a digit: '" + id + "'.");
-            }
-
             if (this._idValueMap.ContainsKey(id))
             {
                 if (autoForceUniqueID)
--- a/Source/SvgElementIdManager.cs
+++ b/Source/SvgElementIdManager.cs
@@ -137,12 +137,11 @@ namespace Svg
         }
 
         /// <summary>
-        /// Ensures that the specified ID is valid within the containing <see cref="SvgDocument"/>.
+        /// Ensures that the specified ID is unique within the containing <see cref="SvgDocument"/>.
         /// </summary>
         /// <param name="id">A <see cref="string"/> containing the ID to validate.</param>
         /// <param name="autoForceUniqueID">Creates a new unique id <see cref="string"/>.</param>
         /// <exception cref="SvgException">
-        /// <para>The ID cannot start with a digit.</para>
         /// <para>An element with the same ID already exists within the containing <see cref="SvgDocument"/>.</para>
         /// </exception>
         public string EnsureValidId(string id, bool autoForceUniqueID = false)


Source/SvgElementIdManager.cs
```diff
@@ -153,15 +153,6 @@ public string EnsureValidId(string id, bool autoForceUniqueID = false)
                 return id;
             }
 
-            if (char.IsDigit(id[0]))
```
mrbean-bremen: Please adapt the documentation:
- Ensures that the specified ID is unique (instead of valid, as validity is not checked anymore)
- remove: The ID cannot start with a digit.

Source/SvgElementIdManager.cs
```diff
@@ -153,15 +153,6 @@ public string EnsureValidId(string id, bool autoForceUniqueID = false)
                 return id;
             }
 
-            if (char.IsDigit(id[0]))
-            {
-                if (autoForceUniqueID)
-                {
-                    return EnsureValidId("id" + id, true);
```
mrbean-bremen: Maybe leave this part in...

H1Gdev: @mrbean-bremen 

Thank you for your review.

ID is converted as follows:
`41cb5de8-82e7-4ab5-94d7-87e10775c848` -> `id41cb5de8-82e7-4ab5-94d7-87e10775c848`

However, other unusable characters (. and -) are not converted, so there is no this issue.

So I think this conversion is unnecessary.

Source/SvgElementIdManager.cs
```diff
@@ -153,15 +153,6 @@ public string EnsureValidId(string id, bool autoForceUniqueID = false)
...
-                if (autoForceUniqueID)
-                {
-                    return EnsureValidId("id" + id, true);
-                }
-                throw new SvgIDWrongFormatException("ID cannot start with a digit: '" + id + "'.");
-            }
-
             if (this._idValueMap.ContainsKey(id))
             {
                 if (autoForceUniqueID)
```
H1Gdev: Here, if an ID is duplicated, a unique ID is generated.
`41cb5de8-82e7-4ab5-94d7-87e10775c848` -> `41cb5de8-82e7-4ab5-94d7-87e10775c848#1`, `41cb5de8-82e7-4ab5-94d7-87e10775c848#2` ...

H1Gdev: However, this can not keep perfect uniqueness.

```svg
<svg version="1.1" baseProfile="basic" id="svg-root"
  width="100%" height="100%" viewBox="0 0 480 360"
  xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <g id="test-body-content" font-family="SVGFreeSansASCII,sans-serif" font-size="18">
    <linearGradient id="Grad1a" gradientUnits="objectBoundingBox" x1="0" y1="0" x2="1" y2="0">
      <stop stop-color="blue" offset="0"/>
      <stop stop-color="lime" offset="1"/>
    </linearGradient>
    <!-- this is referenced... -->
    <linearGradient id="Grad1a" gradientUnits="objectBoundingBox" x1="0" y1="0" x2="1" y2="0">
      <stop stop-color="green" offset="0"/>
      <stop stop-color="yellow" offset="1"/>
    </linearGradient>
    <!-- correctly this is referenced. -->
    <linearGradient id="Grad1a#1" gradientUnits="objectBoundingBox" x1="0" y1="0" x2="1" y2="0">
      <stop stop-color="red" offset="0"/>
      <stop stop-color="white" offset="1"/>
    </linearGradient>
    <rect x="20" y="20" width="440" height="80" fill="url(#Grad1a#1)"/>
    <text font-size="30" x="20" y="130">Linear gradient.</text>
  </g>
</svg>
```

Source/SvgElementIdManager.cs
mrbean-bremen: Ok, thanks!

PR #466, opened 2019-05-22
Fix: Dispose old Bitmap in SVG Viewer.

H1Gdev:
Check about SVG Viewer code is over.

### DIFF
--- a/Samples/SVGViewer/SvgViewer.cs
+++ b/Samples/SVGViewer/SvgViewer.cs
@@ -71,6 +71,12 @@ namespace SVGViewer
 
         private void RenderSvg(SvgDocument svgDoc)
         {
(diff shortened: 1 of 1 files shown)


Samples/SVGViewer/SvgViewer.cs
```diff
@@ -71,6 +71,12 @@ private void TextBox1_KeyDown(object sender, KeyEventArgs e)
 
         private void RenderSvg(SvgDocument svgDoc)
         {
+            if (svgImage.Image != null)
+            {
+                svgImage.Image.Dispose();
+                svgImage.Image = null;
```
mrbean-bremen: this is actually not needed, but doesn't hurt either...

PR #469, opened 2019-05-23
Dispose IDisposable instance.

H1Gdev:


Source/Text/GdiFontDefn.cs
```diff
@@ -44,50 +42,43 @@ public IList<RectangleF> MeasureCharacters(ISvgRenderer renderer, string text)
...
-                {
-                    var bmp = new Bitmap(1, 1);
-                    _graphics = Graphics.FromImage(bmp);
-                }
-                return _graphics;
-            }
-            else
-            {
-                return provider.GetGraphics();
+                throw new NotImplementedException("renderer is not IGraphicsProvider");
```
H1Gdev: Just receive from user without generating in class.

mrbean-bremen: I'm not sure about this one - maybe there is a use case for it, it looks like a bug fix.
@erdomke - in case you are around, do you still remember why this (the 1-byte bitmap fallback) was needed?

H1Gdev: @mrbean-bremen 

I analyzed with code and execution.

This code is performed in the following cases:

1. `renderer` is null.
2. `renderer` does not implement `IGraphicsProvider`.(`renderer` always implements `ISvgRenderer`.)

The first case does not occur with `SvgTextBase` class.(line 304 in SvgTextBase.cs)

The second case occurs with a `renderer` that implements `ISvgRenderer`, such as `DebugRenderer`, but does not implement `IGraphicsProvider`.

Looking at this code, I do not know if implement `ISvgRenderer` but not `IGraphicsProvider`...
In that sense, there is also a way to implement `IGraphicsProvider` with `ISvgRenderer`.

```cs
public interface ISvgRenderer : IGraphicsProvider, IDisposable
```

However, this way forces custom renderer users to make changes.

Issue #473, opened 2019-05-28
SvgText wrong position in some scenario


francois-dorin:
### Scenario
In some scenario, the position of an SvgText element can be incorrect. Example : 
```CSharp
SvgText text = new SvgText("Hello world!");
text.X.Add(new SvgUnit(text.Bound.Width));
text.Y.Add(new SvgUnit(200));
```
The text will be rendered at 0,0, whatever are the values for X and Y.

### Workaround
A workaround is to instanciate new coordinates like bellow: 
```CSharp
text.X = new SvgUnitCollection() { new SvgUnit(text.Bound.Width)) };
text.Y = new SvgUnitCollection() { new SvgUnit(200)); }
```

### Explanation 
I investigate to find twhat cause the bug. In fact, using the `Bound` property put some data in cache (the `_path` attribute). Setting a new collection for X and/or Y invalidates the cache, whereas adding new elements does not.

### Possible fixes
Several approches can be used to fix this issue.

#### Using a read only collection
X and Y properties can be changed for read-only collection. No Add method, no bug! But it break compatibility and can have a large impact on existing code.

#### Using an `ObservableCollection`
Using an `ObservableCollection` for `SvgUnitCollection` instead of a `List` will allow to be notified when the underlying collection is modified (and then invalidates the cache)

#### Disabling cache
Disabling the cache will avoid this issue, but can lead to performance issues.

According to you, what is the best approach to fix this issue? I believe using an `ObservableCollection` is the better one, but I would like confirmation before submitting a PR

mrbean-bremen:
Thanks - `ObservableCollection` sounds like a sensible way to fix this, a respective PR would be very nice!

francois-dorin:
Thank you for your feedback. I will make a PR a soon as possible!

PR #475, opened 2019-05-30
Bug fix for issue #473 : SvgText wrong position in some scenario

francois-dorin:
Pull request for fix the bug #473.

I proceed by changing the base class of `SvgUnitCollection` from `List<SvgUnit>` to `ObservableCollection<SvgUnit>`. I made the needed changes and wrote a unit test.

This PR may cause some compatibility issue if methods specific to List were used (like AddRange, but I provided an implementation to minimize the impact).

I had to add a reference to an assembly:
- WindowsBase.dll for Net35
- Nuget package "System.ObjectModel" for other targets

I do not test each target platform (I do not know how to do it), but unit test and SvgW3CTestRunner run without any problem.

### DIFF
--- a/Source/Basic Shapes/SvgVisualElement.cs
+++ b/Source/Basic Shapes/SvgVisualElement.cs
@@ -288,7 +288,7 @@ namespace Svg
                                     }
 
                                     /* divide by stroke width - GDI behaviour that I don't quite understand yet.*/
-                                    pen.DashPattern = this.StrokeDashArray.ConvertAll(u => ((u.ToDeviceValue(renderer, UnitRenderingType.Other, this) <= 0) ? 1 : u.ToDeviceValue(renderer, UnitRenderingType.Other, this)) /
+                                    pen.DashPattern = this.StrokeDashArray.Select(u => ((u.ToDeviceValue(renderer, UnitRenderingType.Other, this) <= 0) ? 1 : u.ToDeviceValue(renderer, UnitRenderingType.Other, this)) /
                                         ((strokeWidth <= 0) ? 1 : strokeWidth)).ToArray();
 
                                     if (this.StrokeLineCap == SvgStrokeLineCap.Round)
--- a/Source/Svg.csproj
+++ b/Source/Svg.csproj
@@ -90,7 +90,7 @@ NetStandard does not fully support the Drawing2D package - so has been left out.
     <Reference Include="System.Drawing" />
     <Reference Include="System.Web" />
     <Reference Include="System.Xml" />
-	<Reference Include="WindowsBase" />
+    <Reference Include="WindowsBase" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)'=='net40'">
--- a/Source/Svg.csproj
+++ b/Source/Svg.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2;net35</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.2;net35;net452;net472</TargetFrameworks>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <RootNamespace>Svg</RootNamespace>
     <AssemblyName>Svg</AssemblyName>
@@ -37,6 +37,7 @@ Supports multiple targets v3.5 thru to dotnetcore2.2.
 NetStandard does not fully support the Drawing2D package - so has been left out.
     </PackageReleaseNotes>
     <PackageProjectUrl>https://github.com/vvvv/SVG</PackageProjectUrl>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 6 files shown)


Source/Svg.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2;net35</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.2;net35;net452;net472</TargetFrameworks>
```
mrbean-bremen: Why you needed to add these? The 3.5 version should be compatible with later versions.

Source/Svg.csproj
```diff
@@ -37,6 +37,7 @@ Supports multiple targets v3.5 thru to dotnetcore2.2.
 NetStandard does not fully support the Drawing2D package - so has been left out.
     </PackageReleaseNotes>
     <PackageProjectUrl>https://github.com/vvvv/SVG</PackageProjectUrl>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
```
mrbean-bremen: Has this been added intentionally?

Source/Basic Shapes/SvgVisualElement.cs
```diff
@@ -288,7 +288,7 @@ protected internal virtual bool RenderStroke(ISvgRenderer renderer)
                                     }
 
                                     /* divide by stroke width - GDI behaviour that I don't quite understand yet.*/
-                                    pen.DashPattern = this.StrokeDashArray.ConvertAll(u => ((u.ToDeviceValue(renderer, UnitRenderingType.Other, this) <= 0) ? 1 : u.ToDeviceValue(renderer, UnitRenderingType.Other, this)) /
+                                    pen.DashPattern = this.StrokeDashArray.Select(u => ((u.ToDeviceValue(renderer, UnitRenderingType.Other, this) <= 0) ? 1 : u.ToDeviceValue(renderer, UnitRenderingType.Other, this)) /
```
mrbean-bremen: Is this related to the fix?

mrbean-bremen:
Thanks for the PR! I added a couple of questions, and I noticed that one of the image comparison tests failed (painting-stroke-06-t.svg). Can you please check this?

Source/Svg.csproj
francois-dorin: ObservableCollections is not in the same assembly depending of the targetted framework. With the 3.5 it is in WindowsBase.dll. In 4.x it is in System.ObjectModel.
Without adding these frameworks, an application targetting 4.7.2 for example need to explicitly references the WindowsBase.dll.

Source/Svg.csproj
francois-dorin: No. This one can be reverted.

Source/Basic Shapes/SvgVisualElement.cs
francois-dorin: Yes. ConvertAll is a method defined in List<T> but not in ObservableCollection, and do the same thing that Select (which works with any IEnumerable as it is a Linq method).

francois-dorin:
I check the failing unit test. It seems not related to my PR as it also failed on commit 21d61ec "Create CONTRIBUTING.md"

Source/Basic Shapes/SvgVisualElement.cs
mrbean-bremen: Ah ok, thanks!

Source/Svg.csproj
mrbean-bremen: Ok, I see - makes sense.

mrbean-bremen:
Ok, I see now - I hadn't noticed this, as it passed in master. Seems to be some transient problem with loading the file, I will check if I can understand it...

Source/Svg.csproj
francois-dorin: And to complete: as WindowsBase.dll is not in the Framework 4.x, to use WindowsBase.dll we have to provide a full path to the assembly! So, adding new target framework here seems a better approach to me.

mrbean-bremen: Yes, I agree. I also just read up about `AutoGenerateBindingRedirects`, and I think it makes sense to let this switched on, so I will merge this as is.

mrbean-bremen:
Hm, the test still fails in master, and it looks like it is related to the change - it crashes on these lines:
```c#
    if (this.StrokeDashArray.Count % 2 != 0)
    {
         // handle odd dash arrays by repeating them once
         this.StrokeDashArray.AddRange(this.StrokeDashArray);
     }
```
as `StrokeDashArray` is now not mutable. I wonder how it passed for you? It also crashes in the W3CTestRunner.

francois-dorin:
Hmmm, it is a special case where the collection is added to itself! I will correct it

mrbean-bremen:
Well, I can do that if you want - it is not a big change.

mrbean-bremen:
I just tried it, and it works now.

mrbean-bremen:
Ok, I checked it in - my responsibility now, as I have overlooked it in the review :)

mrbean-bremen:
Looks like the other failing test in my own commit was due to a forced commit, so this was not related. I got confused... Anyway, the tests are now green - thanks again for the contribution!

francois-dorin:
No problem. Thank you for correcting it ;)

PR #476, opened 2019-05-30
Add net462 build target

jonthysell:
To package (and publish) a .NET desktop app to the Microsoft Store via the "Desktop Bridge", the app must target .NET 4.6.2 or later.

See: https://docs.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-prepare

This change adds the net462 build target, for app developers who need to target that minimum version.

### DIFF
--- a/Source/Svg.csproj
+++ b/Source/Svg.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.2;net35;net452;net472</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.2;net35;net452;net462;net472</TargetFrameworks>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <RootNamespace>Svg</RootNamespace>
     <AssemblyName>Svg</AssemblyName>
@@ -72,6 +72,12 @@ NetStandard does not fully support the Drawing2D package - so has been left out.
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)'=='net462'">
+    <Title>Svg for .Net Framework 4.6.2</Title>
+    <DefineConstants>$(DefineConstants);NETFULL;NET462</DefineConstants>
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(TargetFramework)'=='net472'">
     <Title>Svg for .Net Framework 4.7.2</Title>
     <DefineConstants>$(DefineConstants);NETFULL;NET472</DefineConstants>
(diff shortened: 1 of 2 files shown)


chucker:
Is there something the 4.7.2 target does that the 4.6.2 one doesn’t? Do we need them both?

jonthysell:
If a developer doesn't want to (or can't) target their code to .NET 4.7.2, then they can't use SVG and publish to the Microsoft Store.

While it may be possible to include the 4.5.2 dll in a Desktop Bridge MSIX package, I'm not sure if the Store will accept it.

I get that SVG probably doesn't want to target every target frameworks version, but I think having a net462 target makes it easier to line up all of the versions on 4.6.2.

Now, if you're asking if SVG needs both 4.6.2 and 4.7.2, I'm not sure.

mrbean-bremen:
Hm, maybe I don't understand it, but for me it doesn't sound like an requirement to target a specific version, but just to make sure it works on that version, because you cannot distribute earlier .NET versions:
> You need to make sure your application runs on .NET 4.6.2. You cannot require or redistribute versions earlier than 4.6.2. This is the version of .NET that shipped in the Windows 10 Anniversary Update. Verifying your application works on this version will ensure that your application will continue to be compatible with future updates of Windows 10. If your application targets the .NET Framework 4.0 or later, it is expected to run on .NET 4.6.2 but you should still test it.

chucker:
> Now, if you're asking if SVG needs both 4.6.2 and 4.7.2

Yes.

There's _already_ `netcoreapp2.2;net35;net452;net472`, which is quite a lot. (Though, if we ever do get .NET Standard 2.0 working without limitations, all of 4.6.2, 4.7.2 and Core 2.2 will be redundant, as all three of those implement .NET Standard 2.0.)

Given that it _does_ compile in 4.5.2, I don't see the added value in it also compiling in 4.6.2 and 4.7.2.

mrbean-bremen:
> Given that it does compile in 4.5.2, I don't see the added value in it also compiling in 4.6.2 and 4.7.2.

Agreed, that's what I think, too.

jonthysell:
> Hm, maybe I don't understand it, but for me it doesn't sound like an requirement to target a specific version, but just to make sure it works on that version, because you cannot distribute earlier .NET versions

Yeah, the documentation is undlear, I already have an issue opened on them: https://github.com/MicrosoftDocs/windows-uwp/issues/1670

mrbean-bremen:
Ah ok - let's wait for a response in this case :)

chucker:
Doesn't this part answer it?

> If your application targets the **.NET Framework 4.0 or later**, it is expected to run on .NET 4.6.2 but you should still test it.

4.0 is required because 4.0 is a different runtime than 2.0/3.0/3.5. Beyond that, 4.x releases are just additive.

Also, [one of their samples uses 4.5.2](https://github.com/microsoft/DesktopBridgeToUWP-Samples/blob/master/Samples/WinformsAppService/WinformsDataApp/WinformsDataApp.csproj).

jonthysell:
"Expected to run" might just mean that _when_ you re-target it should work - but you'd still have to re-target. If you try to create a Desktop Bridge packaging project in VS it will throw a build error if you choose a desktop app targeting 4.5.2. It seems odd that the entry-point _must_ be 4.6.2 or higher, but having older libraries is okay. Which is why I opened the issue on the docs.

tebjan:
@jonthysell your app can target whatever newer framework there is, this should be unrelated to the target framework of the SVG nuget that you reference in your app. Or am i missing something here?

jonthysell:
@tebjan Turns out I was able to publish my app to the app store as long as the entry exe is net462, it didn't matter that the other libraries I consumed were net452.

I'm trying to get the docs updated, but for anyone else that might run into this:

> The actual requirements to package and publish a desktop app using Desktop Bridge are:
> 
> - The entry executable must target 4.6.1 or later.
> - Other .NET assemblies just need to be able to run on .NET 4.6.1, they can target any version less than or equal to the entry executable.

Issue #315, opened 2017-05-10
Crash on hover class  - and wrong colors


ststeiger:
The SVG converter crashes if the SVG contains a hover-style. 
Also, it uses the hover-color instead of the fill-color, if I add the pseudo-selector "empty" for hover.  
Example attached. 

```
<style type="text/css">
path:hover 
{
    fill: orange;
    stroke-width: 10;
    stroke: gold;
    transition: all .2s ease-in-out; 
    transform: scale(1.0);
}

path
{
    transform: scale(1.0);
}
</style>

```

Example-SVG:

```

<?xml version="1.0"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="21cm" height="29.7cm" viewBox="0 0 2098 2969.3" version="1.1">
  <style type="text/css">
path:hover {
    fill: orange;
    stroke-width: 10;
    stroke: gold;
    transition: all .2s ease-in-out; 
    transform: scale(1.0);
}
    
path
{
    transform: scale(1.0);
}
</style>
  <title>tempSvg</title>
  <rect x="0" y="0" width="2098.04" height="2969.26" fill="#FFFFFF" fill-opacity="0" stroke="none" />
  <g id="FM_OBJEKT_RAUM" stroke-width="0.1">
    <path data-handle="25401" d="M282.53 471.08L282.53 477.37L280.43 477.37L280.44 275.3L282.53 275.3L303.51 275.3L303.51 287.88L282.53 287.88L282.53 275.3L280.44 275.3L280.44 79.51L282.53 79.51L303.51 79.51L303.51 92.1L282.53 92.1L282.53 79.51L280.44 79.51L280.44 62.03L699.98 62.03L699.97 79.51L693.68 79.51L693.68 100.49L693.68 463.39L685.99 463.39L685.99 477.37L303.51 477.37L303.51 471.08L282.53 471.08L282.53 477.37L280.43 477.37L282.53 477.37 z" fill="#001400" stroke="#000000" />
[...]

PR #478, opened 2019-06-02
Fix: #315 'Crash on hover class - and wrong colors' issue.

H1Gdev:
No one was catching when an exception occurred with Pseudo-Class that is not supported by Fizzler like hover.
Change to catch and ignore.

[This patch](https://github.com/atifaziz/Fizzler/issues/49) has the same effect as not specifying Pseudo-Class.(`hover` style is applied.)
So I think that can not use this issue.

It is necessary to check the rendering of `<path>` for cause of the slightly different result image !

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -392,13 +392,20 @@ namespace Svg
 
                     foreach (var selector in selectors)
                     {
-                        elemsToStyle = svgDocument.QuerySelectorAll(rule.Selector.ToString(), elementFactory);
-                        foreach (var elem in elemsToStyle)
+                        try
                         {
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
> It is necessary to check the rendering of <path> for cause of the slightly different result image !
I had a look, and there seems to be two things: the slightly different dimensions (lower element height) in the reference image is probably wrong - if checking it in a browser, it seems to be correct as rendered by the lib.
The other thing is the strokes that are black in the rendering, but almost not visible in the reference image. This is due to the problem of rendering strokes with a width < 1 (handled in another issue) - the stroke width is set to 0.1

PR #480, opened 2019-06-06
Omit xml:space if not needed, while writing back SVG files.

H1Gdev:
ref. [this comment](https://github.com/vvvv/SVG/pull/471#pullrequestreview-242359019).

## sample SVG.

```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" width="100" height="80">
<text fill="blue"  font-size="12" font-family="Verdana"                      x="10" y="20"> SVG  is  XML. </text>
<text fill="red"   font-size="12" font-family="Verdana" xml:space="default"  x="10" y="40"> SVG  is  XML. </text>
<text fill="green" font-size="12" font-family="Verdana" xml:space="preserve" x="10" y="60"> SVG  is  XML. </text>
</svg>
```

## output.svg

```svg
<svg x="0" y="0" width="100" height="80" preserveAspectRatio="xMidYMid" font-size="0" xml:space="preserve" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1">
  <text x="10" y="20" font-family="Verdana" font-size="12" style="fill:blue;"> SVG  is  XML. </text>
  <text xml:space="default" x="10" y="40" font-family="Verdana" font-size="12" style="fill:red;"> SVG  is  XML. </text>
  <text xml:space="preserve" x="10" y="60" font-family="Verdana" font-size="12" style="fill:green;"> SVG  is  XML. </text>
</svg>
```

### DIFF
--- a/Source/Document Structure/SvgFragment.cs
+++ b/Source/Document Structure/SvgFragment.cs
@@ -150,7 +150,7 @@ namespace Svg
 
         public override XmlSpaceHandling SpaceHandling
         {
-            get { return GetAttribute("space", false, XmlSpaceHandling.@default); }
+            get { return GetAttribute("space", Inherited, XmlSpaceHandling.@default); }
             set { base.SpaceHandling = value; this.IsPathDirty = true; }
         }
 
--- a/Source/Document Structure/SvgFragment.cs
+++ b/Source/Document Structure/SvgFragment.cs
@@ -148,6 +148,12 @@ namespace Svg
             set { this.Attributes["font-family"] = value; }
         }
 
+        public override XmlSpaceHandling SpaceHandling
+        {
+            get { return GetAttribute("space", false, XmlSpaceHandling.@default); }
+            set { base.SpaceHandling = value; this.IsPathDirty = true; }
+        }
+
         /// <summary>
         /// Applies the required transforms to <see cref="ISvgRenderer"/>.
         /// </summary>
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -287,9 +287,9 @@ namespace Svg
         protected internal TAttributeType GetAttribute<TAttributeType>(string attributeName, bool inherit, TAttributeType defaultValue = default(TAttributeType))
         {
             if (inherit)
-                return Attributes.GetInheritedAttribute<TAttributeType>(attributeName, defaultValue);
+                return Attributes.GetInheritedAttribute(attributeName, defaultValue);
             else
-                return Attributes.GetAttribute<TAttributeType>(attributeName, defaultValue);
+                return Attributes.GetAttribute(attributeName, defaultValue);
         }
 
         /// <summary>
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -282,6 +282,16 @@ namespace Svg
             }
         }
 
+        protected bool Inherited { get; set; } = true;
+
+        protected internal TAttributeType GetAttribute<TAttributeType>(string attributeName, bool inherit, TAttributeType defaultValue = default(TAttributeType))
+        {
+            if (inherit)
+                return Attributes.GetInheritedAttribute<TAttributeType>(attributeName, defaultValue);
+            else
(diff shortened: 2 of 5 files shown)


Source/SvgElement.cs
```diff
@@ -407,8 +417,8 @@ public string ID
         [SvgAttribute("space", SvgAttributeAttribute.XmlNamespace)]
         public virtual XmlSpaceHandling SpaceHandling
         {
-            get { return (this.Attributes["space"] == null) ? XmlSpaceHandling.@default : (XmlSpaceHandling)this.Attributes["space"]; }
-            set { this.Attributes["space"] = value; }
+            get { return GetAttribute("space", Inherited, XmlSpaceHandling.inherit); }
```
H1Gdev: Other properties also I think that there is more better to call this method.

Source/Document Structure/SvgFragment.cs
```diff
@@ -148,6 +148,12 @@ public override string FontFamily
             set { this.Attributes["font-family"] = value; }
         }
 
+        public override XmlSpaceHandling SpaceHandling
+        {
+            get { return GetAttribute("space", false, XmlSpaceHandling.@default); }
```
H1Gdev: Other properties also I think that there is more better to call this method.

Source/SvgElement.cs
```diff
@@ -282,6 +282,16 @@ protected internal virtual SvgAttributeCollection Attributes
             }
         }
 
+        protected bool Inherited { get; set; } = true;
+
+        protected internal TAttributeType GetAttribute<TAttributeType>(string attributeName, bool inherit, TAttributeType defaultValue = default(TAttributeType))
+        {
+            if (inherit)
```
mrbean-bremen: Why not just use `Inherited` here instead of adding that extra parameter?

Source/SvgElement.cs
```diff
@@ -609,65 +619,81 @@ protected virtual void WriteAttributes(XmlTextWriter writer)
...
-                                forceWrite = true;
+                                object parentValue;
+                                if (TryResolveParentAttributeValue(attr.Attribute.Name, out parentValue))
+                                {
+                                    if ((parentValue == propertyValue)
+                                        || ((parentValue != null) && parentValue.Equals(propertyValue)))
+                                        continue;
+
+                                    forceWrite = true;
+                                }
```
mrbean-bremen: The compare tool really does a bad job here...

H1Gdev: I think so, too...

Source/SvgElement.cs
H1Gdev: Thank you for your review.

Some attributes are always `false` like SvgFragment, so add that extra parameter.

H1Gdev: Oh...
`svg` element can nest another `svg` element.

mrbean-bremen: Right. So with that in mind, I still think that this parameter is not needed - it is better in my opinion to make sure that the `Inherited` property is set correctly, and use that inside `GetAttribute`

H1Gdev: Please compare "id" attribute with "xml: space" attribute.

Source/SvgElement.cs
```diff
@@ -393,7 +403,7 @@ protected RectangleF TransformedBounds(RectangleF bounds)
         [SvgAttribute("id")]
         public string ID
         {
-            get { return this.Attributes.GetAttribute<string>("id"); }
+            get { return GetAttribute<string>("id", false); }
```
H1Gdev: `inherit` argument is always `false` because "Id" attribute is not inherited.

Source/SvgElement.cs
mrbean-bremen: Ok, thanks - I had some mind block, obviously. Sorry for the peckering...

PR #484, opened 2019-06-10
Update SvgAttributeAttribute.

H1Gdev:
- Change `InAttributeDictionary` default value to `true`.
- Set `InAttributeDictionary` correctly.

### DIFF
--- a/Source/Basic Shapes/SvgLine.cs
+++ b/Source/Basic Shapes/SvgLine.cs
@@ -15,7 +15,7 @@ namespace Svg
         private SvgUnit _endY;
         private GraphicsPath _path;
 
-        [SvgAttribute("x1")]
+        [SvgAttribute("x1", false)]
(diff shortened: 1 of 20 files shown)


H1Gdev:
ref. #105.

H1Gdev:
In most cases, attribute values are set to `Attributes` property.
So I change `InAttributeDictionary` default value to `true`.

mrbean-bremen:
> In most cases, attribute values are set to Attributes property.

Well, technically true (I checked: 40 vs 31 cases) :)

Is this just a consolidation, or does it fix something?

H1Gdev:
@mrbean-bremen 

Thank you for your review!

> Is this just a consolidation, or does it fix something?

This is just a consolidation.
I did not know what to set in `InAttributeDictionary` property even if I looked at code before...

PR #487, opened 2019-06-14
Update Attribute.

H1Gdev:
- Attribute get properties use `GetAttribute` method. 

1. get from Indexer or `SvgAttributeCollection` `GetInheritedAttribute` method.
-> Change to `GetAttribute(..., Inherited, ...)`
2. get from `SvgAttributeCollection` `GetAttribute` method.
-> Change to `GetAttribute(..., false, ...)`

ref. #480.

### DIFF
--- a/Source/Basic Shapes/SvgVisualElementStyle.cs
+++ b/Source/Basic Shapes/SvgVisualElementStyle.cs
@@ -1,12 +1,4 @@
-using System;
-using System.Collections.Generic;
-using System.Drawing;
-using System.Text;
-using System.Reflection;
 using System.ComponentModel;
-using Svg.DataTypes;
-using System.Text.RegularExpressions;
-using System.Linq;
 
 namespace Svg
 {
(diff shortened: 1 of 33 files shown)


Source/Paths/SvgPath.cs
```diff
@@ -32,66 +28,55 @@ public SvgPathSegmentList PathData
...
-                        _path = null;
+                        if (PathData.Count > 0)
+                        {
+                            // special case with one move command only, see #223
+                            var segment = PathData.Last;
+                            _path.AddLine(segment.End, segment.End);
+                            Fill = SvgPaintServer.None;
+                        }
+                        else
+                            _path = null;
```
H1Gdev: @mrbean-bremen 

Is this needed ?

I think _path should be kept.
Because this process is repeated next time.

mrbean-bremen: This is a fix for an odd case that will happen very rarely, so it should not matter that much.

Source/Basic Shapes/SvgVisualElementStyle.cs
```diff
@@ -30,21 +22,14 @@ public virtual bool Visible
...
         {
-            get
-            {
-                string checkForDisplayNone = this.Attributes["display"] as string;
-                if ((!string.IsNullOrEmpty(checkForDisplayNone)) && (checkForDisplayNone == "none"))
-                    return false;
-                else
-                    return true;
-            }
+            get { return !string.Equals(Display, "none"); }
```
mrbean-bremen: much better :)

PR #495, opened 2019-06-19
Netcore2.2

tebjan:
This is the branch currently used for the new nugets with major version 3. I'll keep this here so we can discuss issues and get feedback whether the nuget works.
https://www.nuget.org/packages/Svg/3.0.3-gecdc54b12b


mrbean-bremen:
@tebjan - as far as I can see, the only difference to master here is the [automatic git versioning](https://github.com/vvvv/SVG/commit/ec9e0ec45c9d0ba2e7a1d964e5fd36428f0a766e) and the version number. Is there a reason not to add this directly to master?

mrbean-bremen:
Ah, and a suggestion - I noticed that you added me to the authors, maybe you can also add H1Gdev, who did most of the latest fixes and refactoring.

tebjan:
yes, good point. indeed, the idea is to merge to master once there is some positive feedback on the nuget preview.

gvheertum:
I was wondering, does this mean that it's also non-Windows compatible? I tried building on VSCode on my Macbook, but there were still some references to Windows related libraries. Since we work with graphics for the bitmap conversion, it's plausible that Windows libs are still needed (at least until .NET Core 3, where they will possibly also port forms and all other windows specific stuff). Is that something one of you looked into already? Just wondering :)

mrbean-bremen:
This should work with MacOS and Linux, though it has not really been tested yet (see #427 for a test @thedonciuxx made under Linux). You may try that also with master, as there seems to be some [problem with the pre-release branch](https://github.com/vvvv/SVG/issues/494)).

ErlendSB:
I'm hitting the NullReferenceException (https://github.com/vvvv/SVG/issues/427#issuecomment-471226864) when running on linux (works fine on windows). I'm guessing the commit mentioned there is not part of this version (3.0.3)?

gvheertum:
> This should work with MacOS and Linux, though it has not really been tested yet (see #427 for a test @thedonciuxx made under Linux). You may try that also with master, as there seems to be some [problem with the pre-release branch](https://github.com/vvvv/SVG/issues/494)).

Yep, my bad. I just did a dotnet build, which tried to build both target-frameworks (including .net 4.5.2, which ofcourse will not work on a Mac). If I specify the framework `dotnet build -f netcoreapp2.2` it compiles as expected (except the SVGViewer, which makes sense).

Nice job getting it ported to .NET Core! 👍

mrbean-bremen:
@ErlendSB - the commit in question is part of this branch, so this should not be the issue.

mrbean-bremen:
Just to put a reference here: under Linux and MacOs you need libgdiplus installed. There is a [packaging project on Github](https://github.com/CoreCompat/libgdiplus-packaging) that makes this easier and also contains installation instructions (copied here for convenience):

### Using libgdiplus on Ubuntu Linux

You can install libgdiplus on Ubuntu Linux using the Quamotion PPA. Follow these steps:
```
sudo add-apt-repository ppa:quamotion/ppa
sudo apt-get update
sudo apt-get install -y libgdiplus
```
### Using libgdiplus on macOS

On macOS, add a reference to the runtime.osx.10.10-x64.CoreCompat.System.Drawing package:

```dotnet add package runtime.osx.10.10-x64.CoreCompat.System.Drawing```

gvheertum:
> @ErlendSB - the commit in question is part of this branch, so this should not be the issue.
> Do you have libgdiplus installed (which is a [requirement under Linux](https://github.com/vvvv/SVG/pull/450#issuecomment-496270065))?

I can imagine it's a bit confusing. The SvgDocument.Open() will in some cases ignore the gdi error and just return a null. This at least explained why together with the dll reference issues, I also had some tests failing with null pointer exceptions if the CoreCompat package is not installed. For example, the tests that use the drawing elements like the one in SvgTextTests will yield the TypeInitializationException (the do calls directly on SVG elements), while the LexerIssueTests yield a null pointer since they use the SvgDocument.Open() which seems to ignore this type of error.

I think it might be more useful for the users if we did something to catch that error and rethrow it or wrap it with a message that makes more sense to the user. In the most likely scenario's users will not work on the SVG elements, but will start by creating an SvgDocument from a string or XML.

ErlendSB:
Installing libgdiplus fixed it.

tebjan:
also merged master into the netcore branch, new nuget preview is ready to test:
https://www.nuget.org/packages/Svg/3.0.9-g6ea47b4039

tebjan:
resolved by #503

PR #500, opened 2019-06-25
Fix: #479 "SVG re-throws ArgumentException due to bounds zero and if opacity is not one".

### LINKED ISSUE #479: SVG re-throws ArgumentException due to bounds zero and if opacity is not one
Svg.dll will re-throw an ArgumentException from CBitmap if rendering a path where the bounds width or height is zero and opacity is not one (the default).

The above is reproducible using master at commit 80674303f5abebcc247a3c4dbe4d6bb2de1bf2cd. SVGViewer can be used to verify the bug.

The original exception is thrown by the CBitmap constructor in line 154 of Source\Basic Shapes\SvgVisualElement.cs

Test svg:
```xml
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="200" height="200">
  <g id="Layer 1">
    <path d="M100,50 L100,150" stroke="black" opacity="0.5" />
  </g>
</svg>
```

Expected result:
line_opacity

H1Gdev:
#### Reference Issue

Fixes #479

And
- Fix: An exception is thrown with empty string in d attribute.
- Small refactoring.


Source/Paths/SvgClosePathSegment.cs
```diff
@@ -1,25 +1,22 @@
...
+                // Important for custom line caps. Force the path the close with an explicit line, not just an implicit close of the figure.
+                var last = pathData.Points.Length - 1;
+                if (!pathData.Points[0].Equals(pathData.Points[last]))
                 {
-                    int i = pathData.Points.Length - 1;
-                    while (i >= 0 && pathData.Types[i] > 0) i--;
-                    if (i < 0) i = 0;
-                    graphicsPath.AddLine(pathData.Points[pathData.Points.Length - 1], pathData.Points[i]);
+                    var i = last;
+                    while (i > 0 && pathData.Types[i] > 0) --i;
```
mrbean-bremen: that was reallly strange code :)

Source/Paths/SvgPathBuilder.cs
```diff
@@ -26,7 +26,7 @@ public class SvgPathBuilder : TypeConverter
         /// <param name="path">A <see cref="string"/> containing path data.</param>
         public static SvgPathSegmentList Parse(string path)
         {
-            if (string.IsNullOrEmpty(path))
+            if (path == null)
```
mrbean-bremen: Did you encounter this condition?

Source/Paths/SvgPath.cs
```diff
@@ -66,6 +66,13 @@ public override GraphicsPath Path(ISvgRenderer renderer)
                         else
                             _path = null;
                     }
+                    else if (renderer == null)
+                    {
+                        var radius = StrokeWidth * 2;
```
mrbean-bremen: Can you please add a comment here? This is not self-explanatory, I think.

Source/Paths/SvgPathBuilder.cs
H1Gdev: @mrbean-bremen 

Thank you for your review.

The following SVG file:
```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="10" height="10">
  <path d="" opacity="0.5" />
</svg>
```

Source/Paths/SvgPath.cs
H1Gdev: @mrbean-bremen 

I added a comment.
But I think that interface specification of `Path` method is cause of confusingness...

`renderer`
null -> Boundary **with** stroke.
other -> Boundary **without** stroke.

mrbean-bremen: Yes, that's what I meant.

cyberjaxx: Why were the ellipses added to this code.  I just spent all evening trying to figure out why a simple line with two arrow markers were drawing 4 circles.  I presume it was done so that someone like me would stumble upon the case that was causing the code path to be executed.
While this is not my first rodeo, I haven't spent much time trekking through your source code, so I hesitate to dive in and propose a fix.

Below is the SVG generated by Inkscape.
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:cc="http://creativecommons.org/ns#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   id="svg2075"
   version="1.1"
   viewBox="0 0 210 297"
   height="297mm"
   width="210mm">
  <defs
     id="defs2069">
    <marker
       style="overflow:visible;"
       id="Arrow1Lend"
       refX="0.0"
       refY="0.0"
       orient="auto">
      <path
         transform="scale(0.8) rotate(180) translate(12.5,0)"
         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
         d="M 0.0,0.0 L 5.0,-5.0 L -12.5,0.0 L 5.0,5.0 L 0.0,0.0 z "
         id="path2082" />
    </marker>
    <marker
       style="overflow:visible"
       id="Arrow1Lstart"
       refX="0.0"
       refY="0.0"
       orient="auto">
      <path
         transform="scale(0.8) translate(12.5,0)"
         style="fill-rule:evenodd;stroke:#000000;stroke-width:1pt;stroke-opacity:1;fill:#000000;fill-opacity:1"
         d="M 0.0,0.0 L 5.0,-5.0 L -12.5,0.0 L 5.0,5.0 L 0.0,0.0 z "
         id="path2079" />
    </marker>
  </defs>
  <metadata
     id="metadata2072">
    <rdf:RDF>
      <cc:Work
         rdf:about="">
        <dc:format>image/svg+xml</dc:format>
        <dc:type
           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
        <dc:title></dc:title>
      </cc:Work>
    </rdf:RDF>
  </metadata>
  <g
     id="layer1">
    <path
       id="path2077"
       d="M 15.119047,40.732133 H 185.96428"
       style="fill:none;stroke:#000000;stroke-width:0.26458332px;stroke-linecap:butt;stroke-linejoin:miter;stroke-opacity:1;marker-start:url(#Arrow1Lstart);marker-end:url(#Arrow1Lend)" />
  </g>
</svg>

H1Gdev: @cyberjaxx

These ellipses are not actually drawn.(It is used in boundary calculation as in comment.)

> Below is the SVG generated by Inkscape.

Please provide SVG if issue occurs.

cyberjaxx: Did you try the SVG that I supplied in my previous comment?  I can repro it on the lated source code using SvgViewer.
I reduced the dimensions  so that you don't need to maximize the window to see the line in SVGViewer.
Btw, would you like me to provide updated code for SvgViewer so that one can view files with a large viewBox either by scaling or adding scroll bars?

H1Gdev: @cyberjaxx

> Did you try the SVG that I supplied in my previous comment?

Do you mean this ?
```
rdf:RDF
<cc:Work
rdf:about="">
dc:formatimage/svg+xml</dc:format>
<dc:type
rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
dc:title</dc:title>
</cc:Work>
</rdf:RDF>
```

This is not SVG...

pablojmp: @H1Gdev @mrbean-bremen i was using Path(null) in order to translate from SVG to a custom vector format via the GraphicsPath. i know this fixes #479, but by adding the ellipses you are indeed breaking this kind of usage. am i using the library incorrectly? i have a workaround in place: implement a "do nothing" ISvgRenderer and pass that instead of null, but i'm not happy with this solution.

pablojmp: perhaps you can add the ellipses in a temporal path in Bounds() and revert the changes to Path()? something like:
`        public override RectangleF Bounds
        {
            get
            {
                var boundsPath = (GraphicsPath) Path(null).Clone();
                var radius = StrokeWidth * 2;
                var bounds = boundsPath.GetBounds();
                boundsPath.AddEllipse(bounds.Left - radius, bounds.Top - radius, 2 * radius, 2 * radius);
                boundsPath.AddEllipse(bounds.Right - radius, bounds.Bottom - radius, 2 * radius, 2 * radius);
                return boundsPath.GetBounds();
            }
        }`

H1Gdev: @pablojmp

I don't know what you are going to do...

Please write sample code, expected and actual value in new issue.

PR #502, opened 2019-06-26
Gdi+ related errors are now wrapped in a specific exception with more detailed information

### LINKED ISSUE #501: When gdi+ cannot be loaded on MacOs or Linux the system should give a more descriptive exception
As found in PR #495 loading problems regarding gdi+ on MacOs and Linux will often result in a confusing exception. Depending on the entry point to the SVG library the user will sometimes not be presented with an error, but will get a null reference to the document.

When on a MacOs system a call is done to library classes like SvgText, we will get a TypeInitializationException, while calls to the SvgDocument.Load(xml) will yield a null as result and will likely cause nullpointer exceptions lateron, when the user works with the document reference.

I propose that at least in the SvgDocument.Load() we watch for the type exceptions and wrap the exception stating the current requirements regarding the gdi+ on MacOs and Linux. Catching calls in all Svg classes (like SvgText) is undoable, so I suggest leaving that there. These will directly result in the type errors, which can be documents (and I saw @mrbean-bremen already doing some additions to the help files, so maybe we can also add the error thrown in that case so people can find it more easily via search engines and github).

gvheertum:
#### Reference Issue

Fixes #501

SvgDocument.Open now has some additional checks to see if the error might be related to the gdi+ issues described in #501. We scan the exception for the presence of a certain type of exception and message in the exception. The specific exceptions is re-thrown (while other exceptions are only logged and not thrown) and is handled at the end of the Load function. If we are handling the specific gdi+ exception, we will wrap it in an exception with a bit more detail.

#### Any other comments?
Added a commented ItemGroup for the CoreCompat package in the Svg.csproj file. This is commented by default, to enable building/running on Mac uncomment the ItemGroup.

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -280,139 +280,177 @@ namespace Svg
 
         private static T Open<T>(XmlReader reader) where T : SvgDocument, new()
         {
-            var elementStack = new Stack<SvgElement>();
-            bool elementEmpty;
-            SvgElement element = null;
-            SvgElement parent;
-            T svgDocument = null;
-            var elementFactory = new SvgElementFactory();
-
-            var styles = new List<ISvgNode>();
-
-            while (reader.Read())
+            try
             {
-                try
+                var elementStack = new Stack<SvgElement>();
+                bool elementEmpty;
+                SvgElement element = null;
+                SvgElement parent;
+                T svgDocument = null;
+                var elementFactory = new SvgElementFactory();
+
+                var styles = new List<ISvgNode>();
+
+                while (reader.Read())
                 {
-                    switch (reader.NodeType)
+                    try
                     {
-                        case XmlNodeType.Element:
-                            // Does this element have a value or children
-                            // (Must do this check here before we progress to another node)
-                            elementEmpty = reader.IsEmptyElement;
-                            // Create element
-                            if (elementStack.Count > 0)
-                            {
-                                element = elementFactory.CreateElement(reader, svgDocument);
-                            }
-                            else
-                            {
-                                svgDocument = elementFactory.CreateDocument<T>(reader);
-                                element = svgDocument;
-                            }
-
-                            // Add to the parents children
-                            if (elementStack.Count > 0)
-                            {
-                                parent = elementStack.Peek();
-                                if (parent != null && element != null)
-                                {
-                                    parent.Children.Add(element);
-                                    parent.Nodes.Add(element);
-                                }
-                            }
-
-                            // Push element into stack
-                            elementStack.Push(element);
-
-                            // Need to process if the element is empty
-                            if (elementEmpty)
-                            {
-                                goto case XmlNodeType.EndElement;
-                            }
-
-                            break;
-                        case XmlNodeType.EndElement:
-
-                            // Pop the element out of the stack
-                            element = elementStack.Pop();
-
-                            if (element.Nodes.OfType<SvgContentNode>().Any())
-                            {
-                                element.Content = (from e in element.Nodes select e.Content).Aggregate((p, c) => p + c);
-                            }
-                            else
-                            {
-                                element.Nodes.Clear(); // No sense wasting the space where it isn't needed
-                            }
-
-                            var unknown = element as SvgUnknownElement;
-                            if (unknown != null && unknown.ElementName == "style")
-                            {
-                                styles.Add(unknown);
-                            }
-                            break;
-                        case XmlNodeType.CDATA:
-                        case XmlNodeType.Text:
-                            element = elementStack.Peek();
-                            element.Nodes.Add(new SvgContentNode() { Content = reader.Value });
-                            break;
-                        case XmlNodeType.EntityReference:
-                            reader.ResolveEntity();
-                            element = elementStack.Peek();
-                            element.Nodes.Add(new SvgContentNode() { Content = reader.Value });
-                            break;
-                    }
-                }
-                catch (Exception exc)
-                {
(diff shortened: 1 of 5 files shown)


gvheertum:
I just noticed that this also pushed the netcore changes... This one is kinda related to that branch, so once that is merged I think this one can be merge in.

mrbean-bremen:
Hm, I'm not sure that it is a good idea to mix these branches... the netcore branch seems to be branched from another test branch I made earlier for the netstandard stuff, and I'm not sure how that merge would work out - there has been the netstandard stuff added, adapted, partly removed again... I would prefer to have a cleaner PR here, otherwise it is difficult to understand the changes.
I can check that tonight, maybe I can merge your changes, and the relevant changes @tebjan made separately.
Thanks for the PR by the way - good to have someone getting this to work on MacOs!

tebjan:
i can merge this PR into the netcore branch, make a new prerelease and see how it works.

tebjan:
also merged master into the netcore branch, new nuget preview is ready to test: 
https://www.nuget.org/packages/Svg/3.0.9-g6ea47b4039

mrbean-bremen:
Thanks @tebjan! I'm still a bit unclear what are the changes of the .NetCore2.2 branch w/r to master (except the automatic git versioning). Shall we merge this change also to master?

tebjan:
it's based on the older netstadard2.0 branch, has the new SDK style project format with automatic nuget build and the netcore build target. if we get positive feedback from @gvheertum about his changes for macOS we can merge back, i'd say.

mrbean-bremen:
ah ok - I had abandonded that branch in favor of the changes for .NetCore2.2, and probably should have deleted it, as it has diverged from master. We have to be careful how to merge back these changes to master, but I agree that it would be a good idea to do after the feedback.

gvheertum:
> it's based on the older netstadard2.0 branch, has the new SDK style project format with automatic nuget build, auto versioning and the netcore build target. if we get positive feedback from @gvheertum about his changes for macOS we can merge back, i'd say.

Looks good to me. Did a testrun on MacOs and it seems to work as expected. So you've got my blessing. :)

I still have to do some research on how to get it as easy as possible for MacOs to get it compiling, maybe I'll add a platform specific csproj or buildstep somewhere, because currently I have to remove the net35 target from the project file before I can build with the additional CoreCompat package (as long as the net35 is in the frameworks the package will yield errors). But that's something for another time (I still have to get used to VSCode and working with csproj files, because I am a bit too used to the VS 2017/2019 way of working).

tebjan:
> Did a testrun on MacOs and it seems to work as expected.

did you use the nuget or built from source?

gvheertum:
> > Did a testrun on MacOs and it seems to work as expected.
> 
> did you use the nuget or built from source?

Source build, do you want me to test the NuGet (that'll take a bit longer since I don't have MacOs projects/builds that use the Svg package, only been playing with the unittests directly from the repo sources).

gvheertum:
@tebjan I found a way to test the NuGet version also (changed the unit-test project to use the NuGet instead of the project reference). That however required me to also put the package include for the CoreCompat in the project file (since that one was previously carried over from the Svg project include).

But I now can confirm that both the nuget and the source behave as expected.

mrbean-bremen:
I created a separate PR (#503) to merge the changes together with the .NetCore changes made by @tebjan, to be sure to not get any leftovers from my botched netstandard branch. Still works nicely under Windows :)

Source/SvgDocument.cs
```diff
@@ -280,139 +280,177 @@ public static T Open<T>(Stream stream, Dictionary<string, string> entities) wher
...
                                 }
                             }
-                        }
-                        catch (Exception ex)
-                        {
-                            Trace.TraceWarning(ex.Message);
+                            catch (Exception ex)
+                            {
+                                Trace.TraceWarning(ex.Message);
+                                if(ExceptionCaughtIsGdiPlusRelated(ex)) { throw; } //Gdi+ errors should be rethrown
```
H1Gdev: This block is application of styles.
In my understanding, I think that GDI+ is irrelevant in this process.

Is this necessary ?

gvheertum:
Good point. I was not really sure about that part. There was some error handling there already, which “swallowed” the errors that happened there. I added the check there to be sure that if a gdi+ exception occurred there it would not be ignored.

However, I guess the code won’t even reach that part of there is a gdi+ issue, since I expect it to fail earlier.

But as said, it’s just for additional safety and a bit with regard to consistency. I did not like the fact that exceptions were ignored since that caused odd behavior in some cases. I think leaving it in is not that much of an issue, if so I think that the check can be removed there. There could be a small performance impact if there are many issues in the style parsing (since the check code will run each time to see if it’s a gdi+ error, but I have the feeling that impact will be quite small).

H1Gdev:
@gvheertum 

Thanks for your response.

> I think leaving it in is not that much of an issue, if so I think that the check can be removed there. 

If this is not a direct solution to this issue, I think that it should be removed.
Increasing insurance code significantly reduces readability.

gvheertum:
Fair point. The (doc) processing code is complex enough already. I’ll do some additional validation soon and will remove the redundant code if indeed no gdi+ interaction is to be expected.

I’m also considering doing a different check for the gdi+ compatibility and wire that up in the beginning of the function. I’m considering a simple function with a call to invoke gdi+, which can also be used as a compatibility check for developers before invoking the svg library. I think that’s going to help readability and portability of that specific part. The result of this call can be cached in the svg lib to mitigate possible performance issues.

Will be continued :)

Issue #346, opened 2017-11-13
netstandard compatibility


MaitreDede:
Hi,
Is it possible to refactor SVG a bit so core SVG parsing is compatible with `netstandard`, and rendering is split into other libraries ? So there can be an implementation for the `framework` (the actual one), and implementations for dotnet core (based on [ImageSharp](http://imagesharp.net/) for example).
Thanks.

tebjan:
hello @MaitreDede, the library is open source. you are welcome do refactor it to your liking. if got something running, don't hesitate to open a pull request.

wyshmily:
@MaitreDede do you have a solution now?

MaitreDede:
@wyshmily The solution would be to fork and refactor. 

The `System.Drawing` library is deeply integrated inside `SVG`, so does `System.Web`. Removing dependencies on both needs to rethink the library.

This will be really time-consuming.

mmosca:
It may be possible to still depend on System.Drawing basic types and
refactor the rendering itself to use something else.

.Net Core does have a lot of the types in System.Drawing available, just
not the GDI rendering code.

MaitreDede:
The Gdi rendering part is in preview. I think it should be better to split
SVG handling and rendering to allow custom rendering engines.
For example I would be pleased to have a rendering based on ImageSharp.

Le 22 nov. 2017 19:24, "Marcelo Bezerra" <notifications@github.com> a
écrit :

JimBobSquarePants:
Having had a quick look at the library it would take serious refactoring to allow enough abstraction to use another rendering engine like ImageSharp since the integration is so [tightly coupled](https://github.com/vvvv/SVG/blob/7ce4de6297dbd6be898de5bc7cb6ca5cb9cd2090/Source/SvgDocument.cs#L426).

Since System.Drawing will never be part of the NetStandard and is being worked on now as part of the [Windows Compatibility Pack](https://blogs.msdn.microsoft.com/dotnet/2017/11/16/announcing-the-windows-compatibility-pack-for-net-core/) to help with migrations only I would recommend making such an abstraction a priority if NetStandard support is a goal for the library.

JaedsonBarbosa:
Hi, I'll try to port that for the .NET Standard, but that port won't have the renderer system, it just will generate the SVG files using a black and white image.

wieslawsoltes:
[...]

PR #509, opened 2019-07-08
Added netstandard2.0 support

wieslawsoltes:
#### Reference Issue

Fixes: #346

Adds support for targeting netstandard2.0, some code was disabled as its not supported in  netstandard2.0, support should come later in  netstandard2.1.

### DIFF
--- a/Source/Painting/SvgColourConverter.cs
+++ b/Source/Painting/SvgColourConverter.cs
@@ -6,6 +6,7 @@ using System.Threading;
 
 namespace Svg
 {
+#if !NETSTANDARD20
     /// <summary>
     /// Converts string representations of colours into <see cref="Color"/> objects.
     /// </summary>
@@ -277,4 +278,5 @@ namespace Svg
             return rgb;
         }
     }
+#endif
 }
--- a/Source/Painting/SvgColourServer.cs
+++ b/Source/Painting/SvgColourServer.cs
@@ -56,11 +56,11 @@ namespace Svg
                 return "inherit";
 
             Color c = this.Colour;
-
+#if !NETSTANDARD20
             // Return the name if it exists
             if (c.IsKnownColor)
                 return c.Name;
-
+#endif
             // Return the hex value
             return String.Format("#{0}", c.ToArgb().ToString("x").Substring(2));
         }
--- a/Source/Painting/SvgPaintServerFactory.cs
+++ b/Source/Painting/SvgPaintServerFactory.cs
@@ -9,13 +9,14 @@ namespace Svg
 {
     internal class SvgPaintServerFactory : TypeConverter
     {
+#if !NETSTANDARD20
         private static readonly SvgColourConverter _colourConverter;
 
         static SvgPaintServerFactory()
         {
             _colourConverter = new SvgColourConverter();
         }
-
+#endif
         public static SvgPaintServer Create(string value, SvgDocument document)
         {
             // If it's pointing to a paint server
@@ -53,6 +54,7 @@ namespace Svg
                     {
                         switch (CountHexDigits(value, 1))
                         {
+#if !NETSTANDARD20
                             case 3:
                                 servers.Add(new SvgColourServer((Color)_colourConverter.ConvertFrom(value.Substring(0, 4))));
                                 value = value.Substring(4).Trim();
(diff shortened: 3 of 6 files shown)


mrbean-bremen:
@wieslawsoltes - thanks for the PR!
@tebjan - what do you think? This has no impact on the other targets, and gets at least some support for Netstandard. The missing color converter stuff will lead to black images in some cases, but it would be a start. I thought that we can wait for Netstandard 2.1 (we discussed this elsewhere), but looking at this PR I'm inclined to merge this. With the .NETCore support in, this is not a big change now.

mrbean-bremen:
@tebjan - any comments?

tebjan:
Looks good, but I wonder whether we can just set up our own color converter and have it completely working and not so many ifdefs. The code of the color converter should be in the . net reference sources.

mrbean-bremen:
Last time I looked it hadn't been there - just the interfaces in netstandard. But that was some time ago, I'll check this again. Generally I agree, of course.

mrbean-bremen:
Actually I'm a bit undecided about this one. According to [this site](https://github.com/dotnet/standard/milestone/3), Netstandard 2.1 is 96% finished, and it looks like it is right around the corner now. Maybe we can just wait this out...

mrbean-bremen:
@wieslawsoltes - I rebased this against master to fix merge conflicts, forgot it is not my own branch...

wieslawsoltes:
> @wieslawsoltes - I rebased this against master to fix merge conflicts, forgot it is not my own branch...

No problem.

wieslawsoltes:
I am also not sure about this, but `.NET Standard 2.1` is also not the best option to wait for. In my software I use `.NET Standard 2.0` as its the best option today.
> Given many of the API additions in .NET Standard 2.1 require runtime changes in order to be meaningful, .NET Framework 4.8 will remain on .NET Standard 2.0 rather than implement .NET Standard 2.1. .NET Core 3.0 as well as upcoming versions of Xamarin, Mono, and Unity will be updated to implement .NET Standard 2.1.

> Library authors who need to support .NET Framework customers should stay on .NET Standard 2.0. In fact, most libraries should be able to stay on .NET Standard 2.0, as the API additions are largely for advanced scenarios. However, this doesn’t mean that library authors cannot take advantage of these APIs even if they have to support .NET Framework. In those cases they can use multi-targeting to compile for both .NET Standard 2.0 as well as .NET Standard 2.1. This allows writing code that can expose more features or provide a more efficient implementation on runtimes that support .NET Standard 2.1 while not giving up on the bigger reach that .NET Standard 2.0 offers.
```
https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/

mrbean-bremen:
Ok, that sounds convincing. 
There is [code for the color converter](https://github.com/dotnet/corefx/blob/master/src/System.Drawing.Common/src/System/Drawing/ColorConverter.cs) in .NET Core, I'm not sure if we can use it (both technically and legally), but I may have a look at it.
As I understand it, your code does not need the non-implemented stuff to function, right?

wieslawsoltes:
> Ok, that sounds convincing.
> There is [code for the color converter](https://github.com/dotnet/corefx/blob/master/src/System.Drawing.Common/src/System/Drawing/ColorConverter.cs) in .NET Core, I'm not sure if we can use it (both technically and legally), but I may have a look at it.

I think if you leave copyright original headers then it should be possible to use as its MIT licensed but not 100% sure if Ms-PL is compatible with MIT.

> As I understand it, your code does not need the non-implemented stuff to function, right?

I don't think I need this code.

mrbean-bremen:
@tebjan - given the discussion above, I still think it makes sense to merge this as is, and try to implement the missing stuff later. There is obviously a use case for this without the color converter stuff...

wieslawsoltes:
> @tebjan - given the discussion above, I still think it makes sense to merge this as is, and try to implement the missing stuff later. There is obviously a use case for this without the color converter stuff...

I have successfully used the code from this PR in my app using `.NET Core 3.0 preview 6` and `.NET Standard 2.0`. I would vote for merging this PR.

wieslawsoltes:
@mrbean-bremen I am using code from this PR for lat few days without any issues. I vote for merging. You can see working code here: https://github.com/wieslawsoltes/Draw2D/tree/SVG/svg/Svg.Skia

mrbean-bremen:
@tebjan - I really would like to merge this as is - see the previous comments. Regarding the local implementation of the missing functionality - I had looked shortly into this, but I will not be able to put much work into this in the next weeks. Anyway, this can be done later, we just have to clearly state in the release notes that NetStandard is incompletely supported - the main feature for the 3.0 release will be .NET Core support.
Speaking of which - when are you planning to put a release together?

tebjan:
ok, i think this can be merged now, since we have a 3.0 release up with current master.

mrbean-bremen:
Ok, thanks - I'll resolve the conflicts and merge.

Issue #508, opened 2019-07-08
Problem with Round cap and dash-strokearray


MichaelBjerregaard:
#### Description
In the attached SVG the dasharray (1.40 , 1.00) first becomes (0.56, 0.40) in SVGVisualElement line 291.
When adjusting for round cap, it becomes (1.56, -0.59) which is invalid.

Changing line 303 to 
      pattern[i] = Math.Abs(pen.DashPattern[i] + offset);  
will avoid the exception, but will probably not paint a correct dash.

#### Example data

[SVG1155.svg.txt](https://github.com/vvvv/SVG/files/3368658/SVG1155.svg.txt)

[DashArrayTests.cs.txt](https://github.com/vvvv/SVG/files/3368665/DashArrayTests.cs.txt)

#### Used Versions
master branch

mrbean-bremen:
Hm, I see the problem - this basically happens if the caps overlap. Using `Abs` may not be the best solution, as that would widen the gap. As GDI is not able to draw an overlapped dashed line (at least not directly), we either have to decompose the dash and draw each part separately (cumbersome and probably not worth it, given that overlapping dashes are not exactly a widely used feature), setting the gap size to a very small value (this would make the gaps wider to not have overlap), or not use a dash at all in this case. Any thoughts?

MichaelBjerregaard:
I can see in #191 what you wanted to achieve, and in my opinion there is no need to make a lot of work out of it.
It seems that a large stroke-width combined with relatively small dash/gap values can create the negative values. In my case a stroke-width of 2.5 and a dash/gap of 1.4/1, which will turn into a solid line if properly adjusted for round caps.
I will suggest that if the average width of dash + gap is less than 2 * stroke-width then ignore round caps.

mrbean-bremen:
Agreed - thanks. I will see if can do this tonight.

mrbean-bremen:
@MichaelBjerregaard - you can have a look at the PR. I added your test to make sure your example doesn't crash, and a simpler example (shown in the PR) to check the functionality.

MichaelBjerregaard:
Thank you for acting so swift.
It looks good.


Fra: mrbean-bremen <notifications@github.com>
Sendt: 10. juli 2019 21:04
Til: vvvv/SVG <SVG@noreply.github.com>
Cc: Michael Bjerregaard <mbp@rib-software.dk>; Mention <mention@noreply.github.com>
[...]

PR #510, opened 2019-07-10
Handle overlapping caps by joining the lines

mrbean-bremen:
- overlapping cups led to a crash - see #508
- workaround to avoid the crash and not deviate too much from correct drawing

To show what it does:
Here is the example used in the test in Chrome:
grafik

And here in the testrunner using SVG lib:
grafik


mrbean-bremen:
@H1Gdev - please have a look!

Source/Basic Shapes/SvgVisualElement.cs
```diff
@@ -279,41 +279,87 @@ protected internal virtual bool RenderStroke(ISvgRenderer renderer)
                         {
                             using (var pen = new Pen(brush, strokeWidth))
                             {
-                                if (this.StrokeDashArray != null && this.StrokeDashArray.Count > 0)
+                                if (StrokeDashArray != null && StrokeDashArray.Count > 0)
                                 {
-                                    if (this.StrokeDashArray.Count % 2 != 0)
+                                    strokeWidth = strokeWidth <= 0 ? 1 : strokeWidth;
```
H1Gdev: What this condition is or will be true in case ?

Source/Basic Shapes/SvgVisualElement.cs
```diff
@@ -279,41 +279,87 @@ protected internal virtual bool RenderStroke(ISvgRenderer renderer)
...
                                     {
                                         // handle odd dash arrays by repeating them once
-                                        this.StrokeDashArray.AddRange(this.StrokeDashArray);
+                                        StrokeDashArray.AddRange(StrokeDashArray);
                                     }
 
-                                    /* divide by stroke width - GDI behaviour that I don't quite understand yet.*/
-                                    pen.DashPattern = this.StrokeDashArray.Select(u => ((u.ToDeviceValue(renderer, UnitRenderingType.Other, this) <= 0) ? 1 : u.ToDeviceValue(renderer, UnitRenderingType.Other, this)) /
-                                        ((strokeWidth <= 0) ? 1 : strokeWidth)).ToArray();
+                                    SvgUnit dashOffset = StrokeDashOffset != null ? StrokeDashOffset : 0;
```
H1Gdev: Use `var` ?

Source/Basic Shapes/SvgVisualElement.cs
mrbean-bremen: This has been used below twice - I just factored out for better readability of the code where it is used.
It seems to make sure that stroke width is set to 1 if StrokeWidth is invalid.

Source/Basic Shapes/SvgVisualElement.cs
mrbean-bremen: Doesn't work with 0, I think I tried that.

Source/Basic Shapes/SvgVisualElement.cs
```diff
@@ -279,41 +279,87 @@ protected internal virtual bool RenderStroke(ISvgRenderer renderer)
...
-                                            pattern[i] = pen.DashPattern[i] + offset;
+                                            pattern[i] = dashPattern[i] + offset;
+                                            if (pattern[i] <= 0)
+                                            {
+                                                // overlapping caps - remove the gap for simplicity, see #508
+                                                if (i < length - 1)
+                                                {
+                                                    // add the next dash segment to the current one
+                                                    dashPattern[i - 1] += dashPattern[i] + dashPattern[i + 1];
+                                                    dashPattern[i] = dashPattern[i + 2];
```
mrbean-bremen: Note to self: this line can be incorporated into the loop below.

mrbean-bremen:
@H1Gdev - thanks for the review!

Source/Basic Shapes/SvgVisualElement.cs
H1Gdev: Use `SvgUnit.Empty` !

H1Gdev:
@mrbean-bremen 

> Yes - nothing we can do about that 😛

Is Chrome the ideal rendering ?

Source/Basic Shapes/SvgVisualElement.cs
H1Gdev: It is necessary to confirm in line 248. . .
```cs
StrokeWidth > 0
```

Source/Basic Shapes/SvgVisualElement.cs
mrbean-bremen: Well, turns out it works with 0 too - I was wrong!

mrbean-bremen: Done.

Source/Basic Shapes/SvgVisualElement.cs
mrbean-bremen: Done.

mrbean-bremen:
> Is Chrome the ideal rendering ?

Maybe not ideal, but overall it seems to be nearer to the standard than the other browsers after my checks, at least most times. So yes, I tend to use it as a standard, if there is nothing matching in the W3C examples, though that maybe debatable.
Anyway I think if the browsers do not agree on details, we shall not overly worry about our implementation of these details.

Thanks for your review! I'll merge this now to fix the crash - we can always think about more improvements later.

PR #511, opened 2019-07-11
Gdi+ capability check revision

gvheertum:
#### Reference Issue
Following the remarks in #502 by @H1Gdev 

#### What does this implement/fix? Explain your changes.
Added a function to query gdi+ capabilities in the SVGDocument and changed the working of the document load to call an ensure function that will throw the gdi exception instead of having a large try-catch block with multiple exception type queries. The new functions are "tested" in 2 simple unittests that allows developers to check their capabilities by executing the tests.

Also updated the csproj of the SVG project to only include the mac version for the targeted framework (I encoutered some issues with targeting).

PR #512, opened 2019-07-11
Gdi+ capability check revision (clean)

gvheertum:
### Reference Issue

Resubmit of #511 
Following the remarks in #502 by @H1Gdev

### What does this implement/fix? Explain your changes.

Added a function to query gdi+ capabilities in the SVGDocument and changed the working of the document load to call an ensure function that will throw the gdi exception instead of having a large try-catch block with multiple exception type queries. The new functions are "tested" in 2 simple unittests that allows developers to check their capabilities by executing the tests.

Also updated the csproj of the SVG project to only include the mac version for the targeted framework (I encoutered some issues with targeting).

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -412,7 +412,6 @@ namespace Svg
                             catch (Exception ex)
                             {
                                 Trace.TraceWarning(ex.Message);
-                                if (ExceptionCaughtIsGdiPlusRelated(ex)) { throw; } // GDI+ errors should be rethrown
                             }
                         }
                     }
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -189,6 +189,11 @@ namespace Svg
             {
                 return false;
             }
+            catch(Exception)
+            {
+                //If somehow another type of exception is raised by the ensure function we will let it bubble up, since that might indicate other issues/problems
+                throw;
+            }
             return true;
         }
 
@@ -210,8 +215,9 @@ namespace Svg
                 {
                     // Throw only the customized exception if we are sure GDI+ is causing the problem
                     throw new SvgGdiPlusCannotBeLoadedException(e);
-                }   
-                //Ignore any other form of error
+                }
+                //If the Matrix creation is causing another type of exception we should just raise that one
+                throw;   
             }
         }
 
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -176,6 +176,68 @@ namespace Svg
         }
 
         /// <summary>
+        /// Validate whether the system has GDI+ capabilities (non Windows related).
+        /// </summary>
+        /// <returns>Boolean whether the system is capable of using GDI+</returns>
+        public static bool SystemIsGdiPlusCapable()
+        {
+            try 
+            {
+                EnsureSystemIsGdiPlusCapable();
+            }
+            catch(SvgGdiPlusCannotBeLoadedException)
+            {
+                return false;
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Ensure that the running system is GDI capable, if not this will yield a
+        /// SvgGdiPlusCannotBeLoadedException exception.
+        /// </summary>
+        public static void EnsureSystemIsGdiPlusCapable()
+        {
+            try
+            {
+                var matrix = new Matrix(0, 0, 0, 0, 0, 0);
+            }
+            // GDI+ loading errors will result in TypeInitializationExceptions, 
+            // for readability we will catch and wrap the error
+            catch (Exception e)
+            {
+                if (ExceptionCaughtIsGdiPlusRelated(e))
+                {
+                    // Throw only the customized exception if we are sure GDI+ is causing the problem
+                    throw new SvgGdiPlusCannotBeLoadedException(e);
+                }   
+                //Ignore any other form of error
+            }
+        }
+
+        /// <summary>
+        /// Check if the current exception or one of its children is the targeted GDI+ exception. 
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
Thanks! I will have a closer look tonight.
@H1Gdev - please have a look, too.

Source/SvgDocument.cs
```diff
@@ -175,6 +175,68 @@ public virtual SvgElement GetElementById(string id)
...
+            // GDI+ loading errors will result in TypeInitializationExceptions, 
+            // for readability we will catch and wrap the error
+            catch (Exception e)
+            {
+                if (ExceptionCaughtIsGdiPlusRelated(e))
+                {
+                    // Throw only the customized exception if we are sure GDI+ is causing the problem
+                    throw new SvgGdiPlusCannotBeLoadedException(e);
+                }   
+                //Ignore any other form of error
```
mrbean-bremen: I would rather re-raise here, as there must be something really wrong if this raises.

Source/SvgDocument.cs
```diff
@@ -280,181 +342,143 @@ public static T Open<T>(Stream stream, Dictionary<string, string> entities) wher
...
+            SvgElement element = null;
+            SvgElement parent;
+            T svgDocument = null;
+            var elementFactory = new SvgElementFactory();
 
-                var styles = new List<ISvgNode>();
+            var styles = new List<ISvgNode>();
 
-                while (reader.Read())
+            while (reader.Read())
```
mrbean-bremen: This is now outside the exception handler. I this intentional?

Tests/Svg.UnitTests/GdiPlusTests.cs
```diff
@@ -0,0 +1,26 @@
...
+        public void GdiPlus_QueryCapability_YieldsTrue()
+        {
+            Assert.True(SvgDocument.SystemIsGdiPlusCapable(), "The gdiplus check should yield true, please validate gdi+ capabilities");
+        }
+
+        [Test]
+        public void GdiPlus_EnsureCapability_YieldsNoError()
+        {
+            SvgDocument.EnsureSystemIsGdiPlusCapable();
+            //This should not cause problems
```
mrbean-bremen: Don't understand the comment... does it belong to the line above?

Tests/Svg.UnitTests/GdiPlusTests.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+    }
+}
```
H1Gdev: Please end with an empty line.

Source/SvgDocument.cs
```diff
@@ -175,6 +175,68 @@ public virtual SvgElement GetElementById(string id)
...
+        /// <summary>
+        /// Check if the current exception or one of its children is the targeted GDI+ exception. 
+        /// It can be hidden in one of the InnerExceptions, so we need to iterate over them.
+        /// </summary>
+        /// <param name="e">The exception to validate against the GDI+ check</param>
+        private static bool ExceptionCaughtIsGdiPlusRelated(Exception e)
+        {
+            var currE = e;
+            int cnt = 0; // Keep track of depth to prevent endless-loops
+            while (currE != null && cnt < 10)
```
H1Gdev: Have you exceeded 10 in your experience ?

H1Gdev:
@gvheertum 

I think that this PR has become very easy to understand !!
Thanks !

Source/SvgDocument.cs
gvheertum: Fair point, I would not expect it to happen, but I share the vision that this should yield an error for the caller to handle.

Source/SvgDocument.cs
gvheertum: Unfortunately yes (not in this library), since we get an error from a deeper library, we cannot always be sure that the content of the exception is "safe". There is a slight possiblity that somewhere somebody wired 2 exceptions as innerexceptions of each other causing an loop, also if we encounter issues like stack of memory issues, there is a possibility that the exception has a lot of inner exceptions, so I guess it's safe to stop at 10. I won't expect it to be deeper than 4 or 5, but if we have to look beyond 10 children, I think we can assume we have a bigger issues (and most likely not a GDI+ issue).

Source/SvgDocument.cs
gvheertum: I checked the older versions, and it seems that prior my changes this code was outside of the try catch. During the initial changes I was not sure where the issues regarding GDI+ were raised, so I put the whole call chain in a try/catch, with additions to the original catch blocks to validate the exception type. Since we now perform the Ensure function at the start of the call we no longer need the try catch construction regarding the GDI+ errors. Other errors are not to be expected, so I thinks it's safe to leave it out, the errors will be raised to the caller since I am not sure how and what we should handle if we put this in a big try/catch.

Tests/Svg.UnitTests/GdiPlusTests.cs
gvheertum: Yeah, it's my silly way to say, "the above call is void and we should arrive here safely", I updated this part.

Tests/Svg.UnitTests/GdiPlusTests.cs
gvheertum: Updated, was not aware of that convention.

gvheertum:
@mrbean-bremen and @H1Gdev, I added some comments and updated the branch with a few commits to resolve the remarks/comments. Good to hear that the pushed changes were well received and made the GDI+ check more readable/usable.

Could you have another look and resolve the conversation (not sure if I should do that if I pushed an update). Thanks. :)

PR #513, opened 2019-07-12
Fix: #507 "SvgText fill missing when custom attribute style set"

### LINKED ISSUE #507: SvgText fill missing when custom attribute style set
Using Svg from nuget version 2.4.3, I'm creating text that has a fill color, and also a custom style attribute set. Back in v1.7 this used to work, however since upgrading, the element is serialized without an explicit `fill` attribute. My guess is the library serializes the fill as a `style`, but my custom style overwrites that.

```
new SvgText("Test")
{
    Fill = blueColor,
    CustomAttributes = { { "style", "cursor: pointer" } }
}
```

I'd expect this to give me

```
<text fill="blue" style="cursor: pointer">Test</text>
```

But instead I get:

```
<text style="cursor: pointer">Test</text>
```

Is this use of the style attribute to set fill, instead of outputting an attribute, a deliberate change in the API?

H1Gdev:
#### Reference Issue

Fixes #507

### DIFF
--- a/Tests/Svg.UnitTests/StyleTest.cs
+++ b/Tests/Svg.UnitTests/StyleTest.cs
@@ -1,0 +1,45 @@
+﻿using System.Drawing;
+using System.IO;
+using System.Xml;
+using NUnit.Framework;
+
+namespace Svg.UnitTests
+{
+    [TestFixture]
+    public class StyleTest
+    {
+        [Test]
+        public void TestStyleValue()
+        {
+            var document = new SvgDocument();
+            var text = new SvgText("Test")
+            {
+                X = { 0f },
+                Y = { 20f },
+                Fill = new SvgColourServer(Color.Blue),
+                CustomAttributes = { { "style", "test0:test0" } }
+            };
+            text.AddStyle("test1", "test1", 0);
+            document.Children.Add(text);
+            using (var stream = new MemoryStream())
+            {
+                document.Write(stream);
+                stream.Position = 0;
+
+                var xmlDoc = new XmlDocument
+                {
+                    XmlResolver = new SvgDtdResolver()
+                };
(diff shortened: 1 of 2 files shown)


H1Gdev:
### Test code.
```cs
var doc = new SvgDocument
{
    Width = 100,
    Height = 100,
};
var text = new SvgText("Test")
{
    X = new SvgUnitCollection(),
    Y = new SvgUnitCollection(),
    Fill = new SvgColourServer(System.Drawing.Color.Blue),
    CustomAttributes = { { "style", "cursor: pointer" } }
};
text.X.Add(new SvgUnit(0f));
text.Y.Add(new SvgUnit(20f));
doc.Children.Add(text);
doc.Write("output.svg");
```

### Output SVG file.
```svg
<svg width="100" height="100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1">
  <text x="0" y="20" style="fill:blue;cursor: pointer">Test</text>
</svg>
```

mrbean-bremen:
Thanks, that was fast!
Your test code actually almost looks like a unit test (just need to check the text style in the output image) - maybe you can add a respective regression test?

Tests/Svg.UnitTests/StyleTest.cs
```diff
@@ -0,0 +1,45 @@
...
+                Assert.IsNotNull(attribute);
+
+                var styles = attribute.InnerText.Split(';');
+                Assert.Contains("test0:test0", styles);
+                Assert.Contains("test1:test1", styles);
+                Assert.Contains("fill:blue", styles);
+            }
+        }
+    }
+}
```
mrbean-bremen: 👍

PR #515, opened 2019-07-12
Apply Styles

H1Gdev:
#### Reference Issue

Relation #507

Provides a way to apply styles by programming generation as follows:
```cs
var doc = new SvgDocument
{
    Width = 100f,
    Height = 100f,
};
var rect = new SvgRectangle()
{
    X = 10f,
    Y = 10f,
    Width = 80f,
    Height = 80f,
};
rect.AddStyle("fill", "blue", 0);
doc.Children.Add(rect);
doc.FlushStyles(true);  // Apply Styles
```

##### output

output

### DIFF
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -50,7 +50,6 @@ namespace Svg
 
         private Dictionary<string, SortedDictionary<int, string>> _styles = new Dictionary<string, SortedDictionary<int, string>>();
 
-
         public void AddStyle(string name, string value, int specificity)
         {
             SortedDictionary<int, string> rules;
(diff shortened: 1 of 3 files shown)


Source/SvgElement.cs
```diff
@@ -59,26 +58,30 @@ public void AddStyle(string name, string value, int specificity)
                 rules = new SortedDictionary<int, string>();
                 _styles[name] = rules;
             }
-            while (rules.ContainsKey(specificity)) specificity++;
+            while (rules.ContainsKey(specificity)) ++specificity;
             rules[specificity] = value;
         }
-        public void FlushStyles()
+
+        public void FlushStyles(bool children = false)
```
mrbean-bremen: Could you add documentation here, as this is a new public API?

mrbean-bremen: Thanks 👍 
Sorry for bothering you, but I would like a bit of information here _when_ to call this, e.g. why it is needed, as it is not as obvious as for example `AddStyle`.

H1Gdev: `SvgElement` originally has public `FlushStyles()` method.

mrbean-bremen:
Ok, thanks - my bad! So my comment should go to the original author :p

PR #516, opened 2019-07-16
Remove unused interface.

H1Gdev:

Remove `ISvgSupportsCoordinateUnits` interface because it is not used by anyone.


Source/Painting/SvgGradientServer.cs
```diff
@@ -215,10 +215,5 @@ public override SvgElement DeepCopy<T>()
 
             return newObj;
         }
-
-        public SvgCoordinateUnits GetUnits()
-        {
-            return GradientUnits;
-        }
     }
```
mrbean-bremen: Have you checked why this has been added in the first place? This is a public interface, and if it is used by someone (I doubt it, but you never know), it would break the code.

H1Gdev: @mrbean-bremen 

Thanks for your response.

This interface is Added on [this commit](d5c659a5475b66bf754bf3d7162775ad3be0ddaf).
And this interface were used in `SvgUnit`.
But this interface is unused on [this commit](3f4ee3330917fed60ae81d4640bab037bb01959e).
But not removed...

H1Gdev: Later or leave by adding `Obsolete` attribute ?

mrbean-bremen: The commit links don't work, I think they are mixed up with PR links... anyway, I would leave that for the time being.

H1Gdev: Sorry...
I edited it.

mrbean-bremen: Ok, thanks - I see now. Probably an `Obsolete` attribute would make sense, and removal in a later version. Thanks for the digging!

PR #519, opened 2019-07-20
Added missing initialization for entities in SvgNodeReader

### LINKED ISSUE #518: NullReferenceException in SvgNodeReader.ResolveEntity when opening SVG with entity definitions
#### Description
My team uses SVG images in a handful of places within our product, and we've recently run into issues trying to use rebranded images from the marketing department.  After some troubleshooting, we tracked down the underlying issue to a NullReferenceException coming from [SvgNodeReader.ResolveEntity, Line 144](https://github.com/vvvv/SVG/blob/9b7d130a1cba516452ae5dd5b79759b0682d1150/Source/SvgNodeReader.cs#L144).

*Note: a quick-and-dirty fix is to reference `Entities` rather than `_entities`*
*Side note: it may also be preferable to use `TryGetValue` over a `ContainsKey` and indexer*

#### Example data
It looks like the issue is triggered by a bunch of garbage entities added by Adobe Illustrator.

I attached (and renamed to .txt due to GitHub's attachment filter) a minimal reproduction SVG.
[header_branding.txt](https://github.com/vvvv/SVG/files/3408608/header_branding.txt)

#### Used Versions
I originally encountered this in SVG 2.2.1.39233, then reproduced it in the latest nuget package, 2.4.3.29446, then pulled a copy of master and reproduced it there as well.

I am currently running on .NET Framework 4.7.2.

mrbean-bremen:
- prevents crash, fixes #518
- added test

### DIFF
--- a/Source/SvgNodeReader.cs
+++ b/Source/SvgNodeReader.cs
@@ -14,11 +14,7 @@ namespace Svg
         public SvgNodeReader(XmlNode node, Dictionary<string, string> entities)
             : base(node)
         {
-            _entities = entities;
-            if (_entities == null)
-            {
-                _entities = new Dictionary<string, string>();
-            }
+            _entities = entities ?? new Dictionary<string, string>();
         }
 
         /// <summary>
--- a/Source/SvgNodeReader.cs
+++ b/Source/SvgNodeReader.cs
@@ -1,10 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Xml;
-using System.IO;
-using System.Collections.Specialized;
 
 namespace Svg
 {
@@ -18,7 +14,11 @@ namespace Svg
         public SvgNodeReader(XmlNode node, Dictionary<string, string> entities)
             : base(node)
         {
-            this._entities = entities;
+            _entities = entities;
+            if (_entities == null)
+            {
+                _entities = new Dictionary<string, string>();
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


mrbean-bremen:
@H1Gdev - please have a look1

Source/SvgNodeReader.cs
```diff
@@ -139,18 +126,18 @@ private void ParseEntities()
         /// </summary>
         public override void ResolveEntity()
         {
-            if (this.NodeType == XmlNodeType.EntityReference)
+            if (NodeType == XmlNodeType.EntityReference)
             {
-                if (this._entities.ContainsKey(this.Name))
+                if (_entities.ContainsKey(Name))
```
H1Gdev: One possible fix is to change it to an `Entities` property, but is there a reason not to do so ?

Source/SvgNodeReader.cs
```diff
@@ -18,7 +14,11 @@ internal sealed class SvgNodeReader : XmlNodeReader
         public SvgNodeReader(XmlNode node, Dictionary<string, string> entities)
             : base(node)
         {
-            this._entities = entities;
+            _entities = entities;
```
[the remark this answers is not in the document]
mrbean-bremen: Yes, thanks - will change it.

Source/SvgNodeReader.cs
mrbean-bremen: It has been a property in the first place, and the bug came from not using it. In this case I see no reason for a property - it only adds complexity instead of removing it (being private, not automatic, and both readable and writable).

gvheertum:
@mrbean-bremen I noticed that somewhere in these commits the netcore 2.2 TargetFramework was removed from the unit test project. Was this intentionally?

mrbean-bremen:
Oops - certainly not! I will check this.

mrbean-bremen:
Thanks - fixed now!

PR #520, opened 2019-07-20
Fixed saving of attributes with default value

mrbean-bremen:
- these had not been written even if they overwrote an inherited attribute
- allow SvgUnitCollection to be none (needed to save stroke-dasharray="none")

I noticed while checking other images in the test runner that the image from PR #504 was not correctly saved, due to the stroke-dasharray (which was none, but overwrote the inherited setting) not being written.


mrbean-bremen:
@H1Gdev - please have a look at this one too, if you find the time.

mrbean-bremen:
Thanks for the review!

PR #523, opened 2019-07-23
Fix: #280 "Some visio images open black"

### LINKED ISSUE #280: Some visio images open black
e.g. the desktop, screen and plotter from Visio 2016's network components - 3d
in comparison the mainframe symbol from the same category is ok

H1Gdev:
#### Reference Issue

Fixes #280


Source/Painting/SvgPatternServer.cs
```diff
@@ -163,71 +157,84 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
                 curr = SvgDeferredPaintServer.TryGet<SvgPatternServer>(curr.InheritGradient, renderingElement);
             } while (curr != null);
 
-            var childElem = chain.Where(p => p.Children != null && p.Children.Count > 0).FirstOrDefault();
-            if (childElem == null) return null;
-            var widthElem = chain.Where(p => p.Width != null && p.Width != SvgUnit.None).FirstOrDefault();
-            var heightElem = chain.Where(p => p.Height != null && p.Height != SvgUnit.None).FirstOrDefault();
-            if (widthElem == null && heightElem == null) return null;
+            var hasChildren = chain.Where(p => p.Children.Count > 0).FirstOrDefault();
```
mrbean-bremen: `hasChildren` suggests a boolean, better use something like `firstChild`, or change the expression to get a boolean, e.g. 
```python
var hasChildren = chain.Where(p => p.Children.Count > 0).FirstOrDefault() != null;
```

Source/Painting/SvgPatternServer.cs
```diff
@@ -163,71 +157,84 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
...
+                return null;
+            var hasX = chain.Where(p => p.X != null && p.X != SvgUnit.None).FirstOrDefault();
+            var hasY = chain.Where(p => p.Y != null && p.Y != SvgUnit.None).FirstOrDefault();
+            var hasWidth = chain.Where(p => p.Width != null && p.Width != SvgUnit.None).FirstOrDefault();
+            var hasHeight = chain.Where(p => p.Height != null && p.Height != SvgUnit.None).FirstOrDefault();
+            if (hasWidth == null || hasHeight == null)
+                return null;
+            var hasPatternUnit = chain.Where(p => p.PatternUnits != SvgCoordinateUnits.Inherit).FirstOrDefault();
+            var hasPatternContentUnit = chain.Where(p => p.PatternContentUnits != SvgCoordinateUnits.Inherit).FirstOrDefault();
+            var hasViewBox = chain.Where(p => p.ViewBox != null && p.ViewBox != SvgViewBox.Empty).FirstOrDefault();
```
mrbean-bremen: Same for all these variable names starting with `has` - they all suggest Boolean values, while they are objects, and are later used as objects, so this reads a bit confusing.

Source/Painting/SvgPatternServer.cs
```diff
@@ -163,71 +157,84 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
...
+                    if (xUnit.Type != SvgUnitType.Percentage)
+                        x *= bounds.Width;
+                    if (yUnit.Type != SvgUnitType.Percentage)
+                        y *= bounds.Height;
+                    if (widthUnit.Type != SvgUnitType.Percentage)
+                        width *= bounds.Width;
+                    if (heightUnit.Type != SvgUnitType.Percentage)
+                        height *= bounds.Height;
+                    x += bounds.X;
+                    y += bounds.Y;
```
mrbean-bremen: yes, this is much more readable now!

Source/Painting/SvgPatternServer.cs
```diff
@@ -163,71 +157,84 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
...
+                    foreach (var child in hasChildren.Children)
+                        child.RenderElement(tileRenderer);
                 }
+
+                var textureBrush = new TextureBrush(tile, new RectangleF(0f, 0f, width, height))
+                {
+                    Transform = EffectivePatternTransform
+                };
+                textureBrush.TranslateTransform(x, y);
+                return textureBrush;
```
mrbean-bremen: This all looks very clean now, though I have trouble understanding where exactly the problem was in the old code. Can you elaborate a bit? (just out of interest)

Tests/W3CTestSuite/PassingTests.txt
```diff
@@ -228,6 +228,8 @@ __issue-227-02.svg
 __issue-239-01.svg
 __issue-242-01.svg
 __issue-279-01.svg
+__issue-280-01.svg
```
mrbean-bremen: I already had noticed that the reference images have been incorrect - thanks for correcting them!

Source/Painting/SvgPatternServer.cs
H1Gdev: @mrbean-bremen 

Thank you for your review.
Variable names were not decided by me, so I will get your idea !!

H1Gdev:
I rebase on latest master.

Source/Painting/SvgPatternServer.cs
H1Gdev: @mrbean-bremen 

It looks like a big change because there are many differences, but process flow has not changed.

Main change is:

- Unify `unit` (`patternUnit` and `patternContentUnit`).

PR #525, opened 2019-07-24
Fix ToSvgString

wieslawsoltes:

The `ToSvgString` method used current culture which produces invalid path data on some culture (uses `,` instead of `.` for separator.

### DIFF
--- a/Source/Paths/SvgPathBuilder.cs
+++ b/Source/Paths/SvgPathBuilder.cs
@@ -14,7 +14,7 @@ namespace Svg
     {
         public static string ToSvgString(this PointF p)
         {
-            return p.X.ToString() + " " + p.Y.ToString();
+            return p.X.ToString(CultureInfo.InvariantCulture) + " " + p.Y.ToString(CultureInfo.InvariantCulture);
         }
     }
 
--- a/Source/SvgNodeReader.cs
+++ b/Source/SvgNodeReader.cs
@@ -1,10 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Xml;
-using System.IO;
-using System.Collections.Specialized;
 
 namespace Svg
 {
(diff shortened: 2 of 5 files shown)


Source/Paths/SvgPathBuilder.cs
```diff
@@ -14,7 +14,7 @@ public static class PointFExtensions
     {
         public static string ToSvgString(this PointF p)
         {
-            return p.X.ToString() + " " + p.Y.ToString();
+            return p.X.ToString(CultureInfo.InvariantCulture) + " " + p.Y.ToString(CultureInfo.InvariantCulture);
```
mrbean-bremen: Thanks - not the first time I see such an error!

H1Gdev:
@wieslawsoltes 
@mrbean-bremen 

https://github.com/vvvv/SVG/blob/a6515547b5e49ee6844611d0423155c24199d434/Source/Paths/SvgPathBuilder.cs#L485-L496

I don't know which of way is good, but I think that it is better to unify either.

wieslawsoltes:
Some example:

https://github.com/dotnet/wpf/blob/9f9c0426b33dee76319460e56f9f0459aee22206/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/LineSegment.cs#L107

https://github.com/dotnet/wpf/blob/52f24ad1517fc4412c989ce684189d7249abd137/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Converters/Generated/PathFigureCollectionValueSerializer.cs#L108

wieslawsoltes:
Seems like `SvgPathBuilder` does not need to change culture after my latest PRs.

mrbean-bremen:
I would prefer to always use the explicit invariant culture for `float.ToString()` and `float.TryParse()`, where the representation is fixed (as it is in SVG) - this is less error-prone IMHO. 
So I like it the way @wieslawsoltes has done this, and I agree that the code in `SvgPathBuilder` can probably be removed (maybe can be added to the other PR).

PR #526, opened 2019-07-24
Fix ToString method for SvgArcSegment

wieslawsoltes:

The `ToString` method used current culture which produces invalid path data on some culture (uses `,` instead of `.` for separator.


Source/Paths/SvgArcSegment.cs
```diff
@@ -135,7 +136,7 @@ public override string ToString()
         {
             var arcFlag = Size == SvgArcSize.Large ? "1" : "0";
             var sweepFlag = Sweep == SvgArcSweep.Positive ? "1" : "0";
-            return "A" + RadiusX.ToString() + " " + RadiusY.ToString() + " " + Angle.ToString() + " " + arcFlag + " " + sweepFlag + " " + End.ToSvgString();
+            return "A" + RadiusX.ToString(CultureInfo.InvariantCulture) + " " + RadiusY.ToString(CultureInfo.InvariantCulture) + " " + Angle.ToString(CultureInfo.InvariantCulture) + " " + arcFlag + " " + sweepFlag + " " + End.ToSvgString();
```
mrbean-bremen: Thanks! While you are it - could you check for more occurencies and fix them, too? I did a quick check, and found at least a couple more in `SvgOrient.ToString` and in `SvgOrientConverter.ConvertFrom`.

wieslawsoltes: Fixed SvgOrient.ToString, not sure if SvgOrientConverter.ConvertFrom needs that.

wieslawsoltes: I have checked other ToString methods and could not find any.

mrbean-bremen: > not sure if SvgOrientConverter.ConvertFrom needs that.

I meant this line:
```C#
      if (!float.TryParse(value.ToString(), out fTmp))
```
As far as I know, this uses the current locale if not given one.

mrbean-bremen: But maybe I'm wrong here... for some reason the culture info is given as an argument in `ConvertFrom` here, but not used.

wieslawsoltes: Maybe we should leave this for now, as I am not convinced this needs invariant culture.

mrbean-bremen: Agreed. Thanks for the help!

PR #530, opened 2019-07-26
Fix: y2 default value.

H1Gdev:

In \<linearGradient\> element, `y2` attribute default value is '0%'.

https://www.w3.org/TR/SVG11/pservers.html#LinearGradientElementY2Attribute

### DIFF
--- a/Source/SvgDefinitionDefaults.cs
+++ b/Source/SvgDefinitionDefaults.cs
@@ -14,7 +14,7 @@ namespace Svg
                 { "SvgRadialGradientServer", new Dictionary<string, string>
                 { { "cx", "50%" }, { "cy", "50%" }, { "r", "50%" } } },
                 { "SvgLinearGradientServer", new Dictionary<string, string>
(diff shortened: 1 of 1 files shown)


Source/SvgDefinitionDefaults.cs
```diff
@@ -14,7 +14,7 @@ public static class SvgDefaults
                 { "SvgRadialGradientServer", new Dictionary<string, string>
                 { { "cx", "50%" }, { "cy", "50%" }, { "r", "50%" } } },
                 { "SvgLinearGradientServer", new Dictionary<string, string>
-                { { "x1", "0%" }, { "x2", "100%" }, { "y1", "0%" }, { "y2", "100%" } } },
+                { { "x1", "0%" }, { "x2", "100%" }, { "y1", "0%" }, { "y2", "0%" } } },
```
mrbean-bremen: Hah, that was easy! Thanks for the fix - you are really getting the bug count down!

PR #532, opened 2019-07-27
Small refactor in SvgOrient.

H1Gdev:

Small refactor in `SvgOrient` and `SvgOrientConverter`.

### DIFF
--- a/Source/DataTypes/SvgOrientConverter.cs
+++ b/Source/DataTypes/SvgOrientConverter.cs
@@ -6,38 +6,32 @@ namespace Svg.DataTypes
 {
     public sealed class SvgOrientConverter : TypeConverter
     {
-        public override object ConvertFrom(ITypeDescriptorContext context, System.Globalization.CultureInfo culture, object value)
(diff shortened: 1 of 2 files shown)


Source/DataTypes/SvgOrientConverter.cs
```diff
@@ -6,48 +6,40 @@ namespace Svg.DataTypes
 {
     public sealed class SvgOrientConverter : TypeConverter
     {
-        public override object ConvertFrom(ITypeDescriptorContext context, System.Globalization.CultureInfo culture, object value)
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
             if (value == null)
-            {
-                return new SvgUnit(SvgUnitType.User, 0.0f);
```
H1Gdev: I do not know why it was `SvgUnit`.
Surely I think that this condition did not become true.

mrbean-bremen: Checking the history, it looks like this got mixed in during an earlier revert a few years ago - obviously a mistake.

PR #537, opened 2019-08-02
Fix: Zero matrix transformation issues.

H1Gdev:

Fix the following issues:

- In \<text\>, \<image\>, and \<switch\>, zero matrix transformation is ignored.
`transform="matrix(0 0 0 0 0 0)"`
- Crash with zero matrix transformation.
`transform="... matrix(0 0 0 0 0 0)"`


H1Gdev:
@mrbean-bremen 

> how do find all these problems?

From code.
If I find something that isn't enough, create and test SVG files.
(Do not check return value of `PushTransforms`. etc.)

PR #539, opened 2019-08-04
Small refactoring for Transforms.

H1Gdev:

Make `Transforms` property check the same.


Source/SvgElement.cs
```diff
@@ -522,8 +522,6 @@ public SvgElement()
             this._elementName = string.Empty;
             this._customAttributes = new SvgCustomAttributeCollection(this);
 
-            Transforms = new SvgTransformCollection();
```
H1Gdev: It is not good to set property of attribute inside element class.

Issue #43, opened 2014-01-02
Filters (e.g. Gaussian blur) do not work


corliss:
SvgGaussianBlur has a bunch of code for computing the gaussian blur of an image, but none of it seems to actually be called. Am I missing something? Cheers.

mrbean-bremen:
Currently, none of the W3C filter examples seems to work.
The filter classes are created dynamically after a class lookup using reflection - though there is no direct call in the code, they are applied, so that is not the reason why they are not working.

AnReZa:
I'm also affected by this issue. I'm using a matrix transform filter to sharpen parts of the image. Instead, the whole image appears black.

mrbean-bremen:
Yes, I didn't really look into this until now - I hoped that someone with a better understanding would have a look (like the original developers...). Maybe I will have a go at this sometime later, but this doesn't look like an easy fix.

AnReZa:
Yes, just by the sheer fact, that this ticket is open for years now, I guess, that this isn't trivial to fix. But thank you anyways for your help!

mrbean-bremen:
>  I'm using a matrix transform filter to sharpen parts of the image. Instead, the whole image appears black.

Not sure if this is your problem, but if there are floats in your matrix, and you use a (non-English) locale where the decimal point is a comma, there has been a bug in reading the float values (fixed a few days ago).
Anyway, I'm going to have a look at the filter stuff, and if you have a concrete example image of a filter not working, I may have a look at that first.

AnReZa:
I guess, that this shouldn't be the case here. I've attached one example SVG for you. You can replace the linked picture with anything you like. The image should show a partially sharpened area in the lower right corner.
[Test.zip](https://github.com/vvvv/SVG/files/3176184/Test.zip)

mrbean-bremen:
Thanks, I will have a look later!

mrbean-bremen:
Ok, I can see the original image, but no filter applied.

mrbean-bremen:
...which is no surprise after checking the implementation: the only implemented filters are feColourMatrix, feGaussianBlur (partly), feMerge and feOffset.

mrbean-bremen:
These are exactly the ones that are used in one of the example images - `__issue-083-01.svg`. `feGaussianBlur` works with one value for `stdDeviation`, but not with two, and the result image looks different from the one in the browser (less blurred). The other 3 (more basic ones) seem to work correctly.

For reference, the following filter primitives are not implemented:
[...]

PR #540, opened 2019-08-04
SvgColourMatrixFix fix

wieslawsoltes:

Fixes string Split for matrix type.

#### Repro:

```svg
<svg width="100%" height="100%" viewBox="0 0 150 360"
    preserveAspectRatio="xMidYMid meet"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:xlink="http://www.w3.org/1999/xlink">

  <!-- ref -->
  <defs>
    <g id="circles">
      <circle cx="30" cy="30" r="20" fill="blue" fill-opacity="0.5" />
      <circle cx="20" cy="50" r="20" fill="green" fill-opacity="0.5" />
      <circle cx="40" cy="50" r="20" fill="red" fill-opacity="0.5" />
    </g>
  </defs>
  <use href="#circles" />
  <text x="70" y="50">Reference</text>

  <!-- matrix -->
  <filter id="colorMeMatrix">
    <feColorMatrix in="SourceGraphic"
        type="matrix"
        values="0 0 0 0 0
                1 1 1 1 0
                0 0 0 0 0
                0 0 0 1 0" />
  </filter>
  <use href="#circles" transform="translate(0 70)" filter="url(#colorMeMatrix)" />
  <text x="70" y="120">matrix</text>

  <!-- saturate -->
  <filter id="colorMeSaturate">
    <feColorMatrix in="SourceGraphic"
        type="saturate"
        values="0.2" />
  </filter>
  <use href="#circles" transform="translate(0 140)" filter="url(#colorMeSaturate)" />
  <text x="70" y="190">saturate</text>

  <!-- hueRotate -->
  <filter id="colorMeHueRotate">
    <feColorMatrix in="SourceGraphic"
        type="hueRotate"
        values="180" />
  </filter>
  <use href="#circles" transform="translate(0 210)" filter="url(#colorMeHueRotate)" />
  <text x="70" y="260">hueRotate</text>

  <!-- luminanceToAlpha -->
  <filter id="colorMeLTA">
    <feColorMatrix in="SourceGraphic"
        type="luminanceToAlpha" />
  </filter>
  <use href="#circles" transform="translate(0 280)" filter="url(#colorMeLTA)" />
  <text x="70" y="320">luminanceToAlpha</text>
</svg>
```


Source/Filter Effects/feColourMatrix/SvgColourMatrix.cs
```diff
@@ -87,7 +87,7 @@ public override void Process(ImageBuffer buffer)
                     };
                     break;
                 default: // Matrix
-                    var parts = this.Values.Replace("  ", " ").Split(new char[] { ' ', '\t', '\n', '\r', ',' });
+                    var parts = this.Values.Replace("  ", " ").Split(new char[] { ' ', '\t', '\n', '\r', ',' }, StringSplitOptions.RemoveEmptyEntries);
```
H1Gdev: I think replacement is also unnecessary.
```cs
var parts = this.Values.Split(new char[] { ' ', '\t', '\n', '\r', ',' }, StringSplitOptions.RemoveEmptyEntries);
```

wieslawsoltes: Yes, pushed change

Source/Filter Effects/feColourMatrix/SvgColourMatrix.cs
```diff
@@ -87,7 +87,7 @@ public override void Process(ImageBuffer buffer)
                     };
                     break;
                 default: // Matrix
-                    var parts = this.Values.Replace("  ", " ").Split(new char[] { ' ', '\t', '\n', '\r', ',' });
+                    var parts = this.Values.Split(new char[] { ' ', '\t', '\n', '\r', ',' }, StringSplitOptions.RemoveEmptyEntries);
                     colorMatrixElements = new float[5][];
```
mrbean-bremen: Thanks - good catch! Could you add your test svg together with the corresponsing png as a regression test?

wieslawsoltes: I can do this but later if that's ok?

mrbean-bremen: Sure!

wieslawsoltes: @mrbean-bremen Seems like the png is not rendered correctly using SVG lib. The issue was only related to the string parsing not rendering.

wieslawsoltes: SVGViewer_2019-08-04_22-03-26

wieslawsoltes: The output from my renderer:
element-feColorMatrix

You can see one issue, the `Filter` property is inherited by Children of the `use` element, which is not what spec says. The `Fileter` property is not inheritable: https://www.w3.org/TR/SVG11/filters.html#FilterProperty, the same issue I had was with `Opacity`: https://www.w3.org/TR/SVG11/masking.html#ObjectAndGroupOpacityProperties

mrbean-bremen: Thanks for digging into this! I will merge as is in this case. Filters are generally very poorly supported (see #43), so I'm not too surprised by this, though the inheritance problem is something that certainly shall be fixed.

PR #544, opened 2019-08-08
Fix: Dispose Matrix instances.

H1Gdev:

Dispose Matrix instances.

### DIFF
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -312,7 +312,7 @@ namespace Svg
             get { return this._customAttributes; }
         }
 
-        private readonly Matrix _zeroMatrix = new Matrix(0, 0, 0, 0, 0, 0);
+        private static readonly Matrix _zeroMatrix = new Matrix(0, 0, 0, 0, 0, 0);
 
         /// <summary>
         /// Applies the required transforms to <see cref="ISvgRenderer"/>.
@@ -328,13 +328,17 @@ namespace Svg
             if (transforms == null || transforms.Count == 0)
                 return true;
 
-            var transformMatrix = transforms.GetMatrix();
-            if (_zeroMatrix.Equals(transformMatrix))
-                return false;
+            using (var transformMatrix = transforms.GetMatrix())
+            {
+                if (_zeroMatrix.Equals(transformMatrix))
+                    return false;
 
-            var graphicsTransform = _graphicsTransform.Clone();
-            graphicsTransform.Multiply(transformMatrix);
-            renderer.Transform = graphicsTransform;
+                using (var graphicsTransform = _graphicsTransform.Clone())
+                {
+                    graphicsTransform.Multiply(transformMatrix);
+                    renderer.Transform = graphicsTransform;
+                }
+            }
 
             return true;
         }
@@ -346,6 +350,7 @@ namespace Svg
         protected internal virtual void PopTransforms(ISvgRenderer renderer)
         {
             renderer.Transform = _graphicsTransform;
+            _graphicsTransform.Dispose();
             _graphicsTransform = null;
             renderer.SetClip(_graphicsClip);
             _graphicsClip = null;
@@ -397,10 +402,13 @@ namespace Svg
         {
             if (Transforms != null && Transforms.Count > 0)
             {
-                var path = new GraphicsPath();
-                path.AddRectangle(bounds);
-                path.Transform(Transforms.GetMatrix());
-                return path.GetBounds();
+                using (var path = new GraphicsPath())
+                using (var matrix = Transforms.GetMatrix())
+                {
+                    path.AddRectangle(bounds);
+                    path.Transform(matrix);
+                    return path.GetBounds();
+                }
             }
             return bounds;
         }
@@ -879,15 +887,16 @@ namespace Svg
                     if (!(child is SvgGroup))
                     {
                         var childPath = ((SvgVisualElement)child).Path(null);
-
                         if (childPath != null)
-                        {
-                            childPath = (GraphicsPath)childPath.Clone();
-                            if (child.Transforms != null)
-                                childPath.Transform(child.Transforms.GetMatrix());
+                            using (childPath = (GraphicsPath)childPath.Clone())
+                            {
+                                if (child.Transforms != null)
+                                    using (var matrix = child.Transforms.GetMatrix())
+                                        childPath.Transform(matrix);
 
-                            if (childPath.PointCount > 0) path.AddPath(childPath, false);
-                        }
+                                if (childPath.PointCount > 0)
+                                    path.AddPath(childPath, false);
+                            }
                     }
                 }
 
@@ -914,7 +923,8 @@ namespace Svg
                         if (childPath.PointCount > 0)
                         {
                             if (child.Transforms != null)
-                                childPath.Transform(child.Transforms.GetMatrix());
+                                using (var matrix = child.Transforms.GetMatrix())
+                                    childPath.Transform(matrix);
 
                             ret.AddPath(childPath, false);
                         }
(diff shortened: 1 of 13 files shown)


tebjan:
This looks a bit dubious to me. Why is the matrix type disposable in the first place? Shouldn't it just be a Struct?

wieslawsoltes:
> This looks a bit dubious to me. Why is the matrix type disposable in the first place? Shouldn't it just be a Struct?

Seems like it is using (allocating) Gdi objects internally:

https://github.com/dotnet/corefx/blob/e468cd3c45a1057cf649056ba61dacc99c7fa546/src/System.Drawing.Common/src/System/Drawing/Drawing2D/Matrix.cs#L16

https://github.com/dotnet/corefx/blob/e468cd3c45a1057cf649056ba61dacc99c7fa546/src/System.Drawing.Common/src/System/Drawing/Drawing2D/Matrix.cs#L71

H1Gdev:
@tebjan 

Many `Matrix` instance were not disposed, but `Matrix` class implements `IDispose`.

@wieslawsoltes 

Thanks for information.

wieslawsoltes:
@H1Gdev Looks like more objects are allocating native memory, maybe we should address all allocations same way as for matrixes?

https://github.com/dotnet/corefx/search?q=Gdip&unscoped_q=Gdip

H1Gdev:
@wieslawsoltes 

#443, #469 ...
I will fix this issue in sequence.

Source/SvgElement.cs
```diff
@@ -312,7 +312,7 @@ public SvgCustomAttributeCollection CustomAttributes
             get { return this._customAttributes; }
         }
 
-        private readonly Matrix _zeroMatrix = new Matrix(0, 0, 0, 0, 0, 0);
+        private static readonly Matrix _zeroMatrix = new Matrix(0, 0, 0, 0, 0, 0);
```
gvheertum: I wonder how this will behave on Mac/Unix without the GDI+ package. There is a test to validate the compatibility which generates a Matrix to test if the GDI+ package works (and will yield a specific exception which is caught and rewrapped in a more sane error), if the _zeroMatrix is now a static property I think this will be initialized somewhere in the beginning, triggering the GDI+ exception on a not really expected location. Not really sure when it will be initialized, but I guess this might break Linux and Mac compatibility.

Also, why make this static? If this element is nicely disposed upon disposal of the SvgElement it should be nicely disposed when the SvgElement is being disposed.

H1Gdev: @gvheertum 

This is because `Matrix` is generated for each instance of `SvgElement` unless it is made `static`.

gvheertum: Fair point, still wondering what happens on systems having the GDI+ problem (Mac and Linux without the proper packages). Maybe we should wrap it in a static property that initializes the matrix on the first request?

`
private static Matrix _zeroMatrix;
private static Matrix ZeroMatrix
{
    get { return _zeroMatrix = _zeroMatrix ?? new Matrix(0, 0, 0, 0, 0, 0);
}
`
And then use the property instead of the field. You can consider try doing this in the constructor (parallelization might get interesting there if multiple instances try to set/check the property, so we might need locking here). My biggest fear is that if the Matrix construction is called on a unexpected location (I haven't got the faintest idea where it will be constructed if it's a static field) the library might have some weird behavior.

Doing the construction on an expected location (in the property or the constructor) might prevent issues on Mac/Linux, because afaik the error occurs there on the constructor of the Matrix. There is some code handling the GDI+ errors and I'm afraid this might break now.

H1Gdev:
@garyranson 

This is only option right now...
Another issue is that `SvgElement` may implement `IDisposable`, so we will consider it again at that time.

Source/SvgElement.cs
```diff
@@ -330,8 +328,9 @@ protected internal virtual bool PushTransforms(ISvgRenderer renderer)
 
             using (var transformMatrix = transforms.GetMatrix())
             {
-                if (_zeroMatrix.Equals(transformMatrix))
-                    return false;
+                using (var zeroMatrix = new Matrix(0f, 0f, 0f, 0f, 0f, 0f))
+                    if (zeroMatrix.Equals(transformMatrix))
+                        return false;
```
mrbean-bremen: Wouldn't it be better to leave the static zero matrix and create it lazily on first use? 
@gvheertum - I think that would not affect the gdiplus check, correct?

Source/SvgElement.cs
```diff
@@ -346,6 +349,7 @@ protected internal virtual bool PushTransforms(ISvgRenderer renderer)
         protected internal virtual void PopTransforms(ISvgRenderer renderer)
         {
             renderer.Transform = _graphicsTransform;
+            _graphicsTransform.Dispose();
```
mrbean-bremen: Is it certain that `_graphicsTransform` cannot be `null` at this point?

Source/SvgElement.cs
gvheertum: It should indeed not affect the GDI+ handling if you do a lazy load (in the constructor or the property), since that would basically be the same as it is created now, being the Matrix getting constructed on the construction of the class or when called instead of when the runtime decides to create the static fields.

However, I think that it might limit readability and stability (in terms of future changes) if we use a static with a onetime initializer or a construction to get the field initially created in the constructor. Since the field resides in the class people can still accidentally use the non-wrapped field (thus ignoring a null check and initialization) or for example remove the initializer code or move the initializer back to the property/field (unless it's very well documented people tend to do this kind of small changes).

From a code quality perspective I think that having the matrix used this way is fine for now, because it's only used in one location/check, having additional code to initialize a constant that is only used once in a check might be a bit confusing and a bit over-done. It's nicer to do it in a static or const to limit allocation and memory footprint, but this might also be a bit over-engineering or micro-optimization (I don't have a clue of the memory impact of the Matrix, but I cannot really imagine it making a huge impact).

If I were to have this kind of constants in my code I would eventually move them to a specialized class/holder for the constant values (e.g. a SvgConstants class). This class can be a static class, and will expose static properties that do the initialization on the first request (if required together a lock { } mechanism if concurrency might be an issue). This keeps constants grouped and allows you to use constants throughout your library.

mrbean-bremen: Ok, thank you for the very thorough response - this answers my question! I agree that the current way is more readable and secure, and that we don't want premature optimization.

Makes me think about performance tests that could make sense... I noticed that in some of the example images rendering is very slow, so we certainly have some performance problems, though I have no idea what they are.

Source/SvgElement.cs
H1Gdev: @mrbean-bremen 

Thanks for your review.
Immediately after this, it is set to null.
Is the answer correct ?

mrbean-bremen: No, I mean can it be already `null` at this point? That would crash if trying to dispose it. I checked the code, but couldn't exclude that.

H1Gdev: @mrbean-bremen 

Previous step:
```cs
renderer.Transform = _graphicsTransform;
```
This setting does not allow `null`, never _graphicsTransform becomes `null`.

mrbean-bremen: Sorry, still don't get it - if `_graphicsTransform` is `null` before that (which it is initially, and also after calling this method), `null` would be assigned to the Transform matrix, and the next line would crash. So my question is: is it ensured that `_graphicsTransform` cannot be `null` when entering this method?

H1Gdev: @mrbean-bremen

PushTransforms and PopTransforms are pair.So, be called PopTransforms after PushTransforms.

ref. #537

mrbean-bremen: Yes, I understand that, but as this is not enforced... anyway, you are right - calling it otherwise would be a programming error and would be caught in the review or by the crash.
Generally, I'm used to put this kind of symmetric actions into a context, at least in Python and C++ (which I'm using far more than C#, so bear with me...), but this seems to be less common in C#. 
Anyway, I'm merging now - thanks again for the contribution!

H1Gdev: @mrbean-bremen 

I decided that this method should remain as it is because this method is `protected internal`, this time.

PR #545, opened 2019-08-08
Small refactoring for Render.

H1Gdev:

Small refactoring for `Render` method.

### DIFF
--- a/Source/Filter Effects/SvgFilter.cs
+++ b/Source/Filter Effects/SvgFilter.cs
@@ -85,7 +85,7 @@ namespace Svg.FilterEffects
 
         private RectangleF GetPathBounds(SvgVisualElement element, ISvgRenderer renderer, Matrix transform)
         {
-            var bounds = element.Path(renderer).GetBounds();
+            var bounds = element is SvgGroup ? element.Path(renderer).GetBounds() : element.Bounds;
             var pts = new PointF[] { bounds.Location, new PointF(bounds.Right, bounds.Bottom) };
             transform.TransformPoints(pts);
 
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -554,7 +554,7 @@ namespace Svg
         /// <param name="renderer">The <see cref="ISvgRenderer"/> that the element should use to render itself.</param>
         public void RenderElement(ISvgRenderer renderer)
         {
-            this.Render(renderer);
+            Render(renderer);
         }
 
         /// <summary>Derrived classes may decide that the element should not be written. For example, the text element shouldn't be written if it's empty.</summary>
@@ -844,10 +844,8 @@ namespace Svg
         /// <param name="renderer">The <see cref="ISvgRenderer"/> to render the child <see cref="SvgElement"/>s to.</param>
         protected virtual void RenderChildren(ISvgRenderer renderer)
         {
-            foreach (SvgElement element in this.Children)
-            {
+            foreach (var element in Children)
                 element.Render(renderer);
-            }
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


Source/Text/SvgTextBase.cs
```diff
@@ -257,56 +257,10 @@ public override RectangleF Bounds
             }
         }
 
-        /// <summary>
-        /// Renders the <see cref="SvgElement"/> and contents to the specified <see cref="Graphics"/> object.
-        /// </summary>
-        /// <param name="renderer">The <see cref="ISvgRenderer"/> object to render to.</param>
-        /// <remarks>Necessary to make sure that any internal tspan elements get rendered as well</remarks>
-        protected override void Render(ISvgRenderer renderer)
```
H1Gdev: Main purpose of this PR is to erase this method.

Source/Filter Effects/SvgFilter.cs
```diff
@@ -85,7 +85,7 @@ private Matrix GetTransform(SvgVisualElement element)
 
         private RectangleF GetPathBounds(SvgVisualElement element, ISvgRenderer renderer, Matrix transform)
         {
-            var bounds = element.Path(renderer).GetBounds();
+            var bounds = element is SvgGroup ? element.Path(renderer).GetBounds() : element.Bounds;
```
H1Gdev: This condition may not be enough, but I don't know now...

Source/Text/SvgTextBase.cs
mrbean-bremen: Hm, I don't understand that initial comment about the tspan elements, so I also don't understand how you got rid of this problem. Can you elaborate on this?

Source/SvgElement.cs
```diff
@@ -844,10 +844,8 @@ protected virtual void Render(ISvgRenderer renderer)
         /// <param name="renderer">The <see cref="ISvgRenderer"/> to render the child <see cref="SvgElement"/>s to.</param>
         protected virtual void RenderChildren(ISvgRenderer renderer)
         {
-            foreach (SvgElement element in this.Children)
-            {
+            foreach (var element in Children)
```
gvheertum: Any specific reason why to pick var instead of SvgElement? When writing new code I can imagine picking var, but any reason why you changed an existing explicit declaration to a var?

H1Gdev: @gvheertum 

It is basically replaced with `var`.

Source/Text/SvgTextBase.cs
H1Gdev: @mrbean-bremen 

The intent of this comment is to call `RenderChildren` method.
This has been moved to `RenderFillAndStroke` method.
I think this PR will be necessary in the future when using `filter` in text.

mrbean-bremen: Ok, thanks - I get it!

Source/SvgElement.cs
mrbean-bremen: @H1Gdev - this does not answer the question 😉 
I will merge now, but I agree with @gvheertum that we don't need to adapt all working code to our own coding style 😜

H1Gdev: I didn't understand your question.

This time I was reviewing the processing of `Render` method.
There was a change to `var` as part of this change.

Certainly, this method does not change anything from the original source code.
So I will revert to original source code.

mrbean-bremen:
No need to revert this - this is just to agree on some guidelines in the future!

gvheertum:
Yep, I'm with @mrbean-bremen, no need to revert, I was just curious. Generally the boyscout rule is a nice rule, but sometimes a developer is overly motivated to be a boy scout ;)

PR #547, opened 2019-08-10
Revert method to original code.

H1Gdev:
#### Reference Issue

PR https://github.com/vvvv/SVG/pull/545#discussion_r312618979


mrbean-bremen:
As discussed in the issue - no need for this revert. Let's just agree to be a bit less eager with the boy scout rule in the future, as @gvheertum said 😄 
(In case you don't know - the boy scout rule states: "Leave your code better than you found it.")

Issue #400, opened 2019-01-09
Support for .NET Core


legas1:
Hello, I want to thank you for this awesome lib. However when I deal with .svgs from Adobe Illustrator (not sure which version, because graphics man...) I got bunch of errors like:
`The CreateDocument method can only be used to parse root <svg> elements.` or `Stack empty.`

I use SVG 2.4.0, NetCore 2.1.

SVG sample here: https://pastebin.com/jsfeRi1y
Code I use is just simple open and after this the errors happen: `SvgDocument svgDocument = SvgDocument.Open(file);`

Edit: maybe it can be fixed by merging #326.

mrbean-bremen:
I adapted the issue title to clarify what this is about. Note that there is a new PR (#415) that specifically adds support for .NET Core 2.2, so that may work for you.

Noah1911:
I'll check it out later, thank you.

laci84:
please some workaround or merge pull request to support .net core

mrbean-bremen:
The committer wanted to make some changes I'm waiting for, and we are also waiting for @tebjan for clarification regarding VS version - so I can't give you a definitive date for when this will be merged...

tebjan:
i'd say VS2017 is fine. if it all works with VS2015, even better. but i think .net core and standard have much better support in 2017.

mrbean-bremen:
Ok, thanks! So merging the PR which requires VS2017 would be fine (once it is finished)?

mrbean-bremen:
@tebjan - I think it would make sense to make another minor release before that, just in case - there have been [a few fixes](https://github.com/vvvv/SVG/wiki/Release-Notes) since the last release.

laci84:
when can i except that release with .net core support?
thx!

tebjan:
@mrbean-bremen sure thing, i'll upload a new nuget.

mrbean-bremen:
@Noah1911 - it has been some time; if you are still interested, could you please check with current master?

Noah1911:
@mrbean-bremen - yes I'm still interested. I still don't have time to test it, but it looks promising. I try to test it at the end of this week.

nrbalda:
Hey I'm going to chime in here since I had the same problems with .NET core 2.1.  I was getting my This morning I updated everything to .NET core 2.2 and cloned master.  I popped the dll in there with a local reference and suddenly all my SVGs started rendering.

[...]

PR #549, opened 2019-08-10
Research svg root element error

gvheertum:
#### Reference Issue
Test case for issue #400 regarding the possible problems with multiple drawing on an open document. The testcase uses the SVG file provided @8m0 and tests multi drawings, one test with drawings after each other, another test with drawing inside of another using block with a bitmap.

Put multi-drawing in a test harness.

#### Any other comments?
Running with a different SVG causes no issues (on MacOs), running with the SVG provided by @8m0 however failed on Mac due to a font issue. The test might need to be validated on a Windows machine.


mrbean-bremen:
Thanks a lot! I will check this under Windows. 
BTW, I was sure that I added the netcore target back to the unit tests, but looks like I didn't push it to master...

PR #553, opened 2019-08-12
Dispose Matrix.

H1Gdev:
#### Reference Issue

Ref. #544 

`Matrix` had been created here...

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -206,7 +206,7 @@ namespace Svg
         {
             try
             {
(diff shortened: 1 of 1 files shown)


Source/SvgDocument.cs
```diff
@@ -206,7 +206,7 @@ public static void EnsureSystemIsGdiPlusCapable()
         {
             try
             {
-                var matrix = new Matrix(0, 0, 0, 0, 0, 0);
+                using (var matrix = new Matrix(0f, 0f, 0f, 0f, 0f, 0f)) { }
```
mrbean-bremen: ah, yes!

PR #555, opened 2019-08-15
Fix: DeepCopy method error.

H1Gdev:

Fixed `DeepCopy` method error in `SvgElement` derived class.


Tests/Svg.UnitTests/SvgElementDeepCopyTest.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        /// <summary>
+        /// Checks the document if it contains the correct information when exported to XML.
+        /// </summary>
+        /// <param name="svgDocument">The SVG document to check.</param>
+        private static void CheckDocument(SvgDocument svgDocument)
+        {
+            //TODO: check if the assignablefrom indeed is the correct replacement
+            Assert.AreEqual(2, svgDocument.Children.Count);
+            Assert.IsAssignableFrom(typeof(SvgDefinitionList), svgDocument.Children[0]);
```
mrbean-bremen: Shouldn't that be rather a type test? e.g. something like:
```C#
Assert.That(svgDocument.Children[0], Is.InstanceOf(typeof(SvgDefinitionList)));
```

gvheertum: @mrbean-bremen `IsAssignableFrom` is a type check. I believe the main difference between the two options is whether derived types also classify for the check.

Also, I would try to keep away from the `Assert.That` for readability. Most of the assertions are available directly from the Assert class, using an Assert.That (in my opinion) might impact readability for some developers.

mrbean-bremen: Ok, thanks! I think the readability of the `Assert.That`is a matter of taste, but I agree that we should stuck to one style.
I'll merge then, thanks @H1Gdev!

PR #556, opened 2019-08-15
Change SvgSymbol namespace.

H1Gdev:
#### Reference Issue

ref. #93.

Only `Svg.Document_Structure`, **_**(underscore) is included in namespace.
So, change namespace without **_**(underscore).

I think `Svg.Symbol` is good.(like `SvgDefinitionList`, `SvgFragment` and `SvgGroup`.)
However, because there is a possibility of build error in existing environments, I decided on this namespace.

### DIFF
--- a/Source/Document Structure/SvgSymbol.cs
+++ b/Source/Document Structure/SvgSymbol.cs
@@ -108,7 +108,6 @@ namespace Svg.Document_Structure
     /// An element used to group SVG shapes.
     /// </summary>
     [Obsolete("Use Svg.DocumentStructure.SvgSymbol.")]
-    [SvgElement("symbol")]
     public partial class SvgSymbol : SvgVisualElement
     {
         /// <summary>
@@ -199,6 +198,14 @@ namespace Svg.Document_Structure
         public override SvgElement DeepCopy()
         {
             return DeepCopy<SvgSymbol>();
+        }
+
+        public override void Write(System.Xml.XmlTextWriter writer)
+        {
+            var symbol = base.DeepCopy<Svg.DocumentStructure.SvgSymbol>() as Svg.DocumentStructure.SvgSymbol;
+            symbol.ViewBox = ViewBox;
+            symbol.AspectRatio = AspectRatio;
+            symbol.Write(writer);
         }
     }
 }
--- a/Source/Document Structure/SvgSymbol.cs
+++ b/Source/Document Structure/SvgSymbol.cs
@@ -2,7 +2,7 @@
 using System.Drawing;
 using System.Drawing.Drawing2D;
 
-namespace Svg.DocumentStructure
+namespace Svg
 {
     /// <summary>
     /// An element used to group SVG shapes.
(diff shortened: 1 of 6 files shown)


Source/Document Structure/SvgSymbol.cs
```diff
@@ -1,10 +1,8 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Drawing;
+using System.Drawing.Drawing2D;
 
-namespace Svg.Document_Structure
+namespace Svg.DocumentStructure
```
mrbean-bremen: Hm, I understand that this namespace fits better into the naming schema, but are you duplicating the whole class for this? I'm not sure that it is worth it...

H1Gdev: @mrbean-bremen 

`Svg.Document_Structure.SvgSymbol` will be removed.
So I added `Obsolete`.

mrbean-bremen:
Also referring this to @tebjan - I'm unsure about the policy here.

mrbean-bremen:
I'm not happy with this duplicated code just for the sake of a more convenient namespace name. If there were a way to just add an alias to the namespace it would be ok. As far as I know, a namespace alias can only be used on import and cannot be exported, but hopefully someone knows a better possibility?

H1Gdev:
@mrbean-bremen 

`SvgSymbol` seems to be stable class, so merge this PR ?

H1Gdev:
rebase.

mrbean-bremen (approved):
I like this better, thanks! Can you add an entry in the release notes for the changed API?

H1Gdev:
@mrbean-bremen 

I wrote the following in [this PR](https://github.com/svg-net/SVG/pull/556#issue-307565600):

> I think `Svg.Symbol` is good.(like `SvgDefinitionList`, `SvgFragment` and `SvgGroup`.)
> However, because there is a possibility of build error in existing environments, I decided on this namespace.

But, still do not need this namespace, so I decided to remove it in [this commit](https://github.com/svg-net/SVG/commit/a86808405c8235e3fa1dcc6c5198ecac600b5252).
[This commit](https://github.com/svg-net/SVG/commit/a86808405c8235e3fa1dcc6c5198ecac600b5252), will [CS0104](https://docs.microsoft.com/dotnet/csharp/misc/cs0104) build error in code, such as the following.

```cs
using Svg.Document_Structure;

...


var symbol = new SvgSymbol // error
{
    Fill = new SvgColourServer(Color.Red),
    Stroke = new SvgColourServer(Color.Black),
    StrokeWidth = 2
};
```
However, I think [this error](https://docs.microsoft.com/dotnet/csharp/misc/cs0104) would make users aware of the change.(It's easy to fix.)
(Of course, it will run correctly in pre-built executables.)

mrbean-bremen:
While this looks ok to me, I'm still a bit unsure about the consequences to users. 
@wieslawsoltes, can you please have another look?

mrbean-bremen (approved):
I'm going to merge this now, I think it is ok as is.

Issue #167, opened 2015-07-19
Provide ability to insert javascript in SVG


danielwalls:
I can't see anything in the library that would allow me to generate a javascript section. ie.

```
<svg>
    <script type="text/javascript">
        // <![CDATA[

        // ]]>
    </script>
</svg>
```

As an aside - is there any forum/documentation/wiki that helps people get started using this library? I have a number of questions to ask someone and I'm keen to contribute as well. Is there anyone to contact?

Emasoft:
Why do you want to use javascript for add interativity in SVG? SVG already provides an excellent set of animation and interactive functions called SMIL. You can see a lot of examples on codepen:
http://codepen.io/search?q=smil&limit=all&depth=everything&show_forks=false

I agree that this SVG library should provide support for SMIL. Many animation packages edit and save directly in SMIL. With SMIL support you can do in minutes things that would require days to do in code.

tebjan:
there is now a wiki... if you have helpful information please add it there. if you want to contribute just open a pull request.

gvheertum:
Sorry for necromancing an old issue, but since I also use a lot of SVG's with scripting this issue was something I wanted to fix anyway :)

mrbean-bremen:
Closing as support for `script` tag has been added.

PR #558, opened 2019-08-16
Added support for the SvgScript tag

gvheertum:
#### Reference Issue
Should address #167 which requests the addition of script tags.

Allows the creation of SvgScript in the SVG document and can parse the script tags in a SVG file in a type-safe way. Adding of SvgScript was already possible in the past by using a generic tag, having a typed tag for scripts will make handling scripts easier.

Included test cases to test behavior of the script tag on reading and outputting.

### DIFF
--- a/Source/Document Structure/SvgScript.cs
+++ b/Source/Document Structure/SvgScript.cs
@@ -9,9 +9,36 @@ namespace Svg
     [SvgElement("script")]
     public class SvgScript : SvgElement
     {
+
+        private string _scriptType;
+
+        [SvgAttribute("type")]
+        public string ScriptType  //TODO: Enum?
+        { 
+            get { return _scriptType; } 
+            set { _scriptType = value; } 
+        }
+
+        private string _crossOrigin;
+
+        [SvgAttribute("crossorigin")]
+        public string CrossOrigin 
+        {
+            get { return _crossOrigin; }
+            set { _crossOrigin = value; }
+        }
+
+        private string _href;
+        [SvgAttribute("href")]
+        public string Href //TODO: URI? 
+        {
+            get { return _href; }
+            set { _href = value; }
+        }
+
         public override SvgElement DeepCopy()
         {
-            throw new System.NotImplementedException();
+            return DeepCopy<SvgScript>();        
         }
     }
 }
--- a/Source/Document Structure/SvgScript.cs
+++ b/Source/Document Structure/SvgScript.cs
@@ -1,0 +1,17 @@
+using System.Drawing;
+using System.Drawing.Drawing2D;
+
+namespace Svg
+{
+    /// <summary>
+    /// An element used to define scripts within SVG documents.
+    /// </summary>
+    [SvgElement("script")]
+    public class SvgScript : SvgElement
+    {
+        public override SvgElement DeepCopy()
+        {
+            throw new System.NotImplementedException();
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


Source/Document Structure/SvgScript.cs
```diff
@@ -0,0 +1,81 @@
...
+            get { return this.Content; }
+            set { this.Content = value; }
+        }
+
+        private string _scriptType;
+
+        [SvgAttribute("type")]
+        public string ScriptType 
+        { 
+            get { return _scriptType; }
```
H1Gdev: If there is no special reason, set it to `Attributes` like any other element.

Source/Document Structure/SvgScript.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                writer.WriteAttributeString("crossorigin", CrossOrigin);
+            }
+            if(!string.IsNullOrEmpty(ScriptType))
+            {
+                writer.WriteAttributeString("type", ScriptType);
+            }
+        }
+
+        protected override void WriteChildren(System.Xml.XmlTextWriter writer)
```
H1Gdev:
```cs
using System.Xml;
```

Source/Document Structure/SvgScript.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            return DeepCopy<SvgScript>();        
+        }
+
+        public override bool ShouldWriteElement() 
+        { 
+            return true;    
+        }
+
+        protected override void WriteAttributes(System.Xml.XmlTextWriter writer)
```
H1Gdev: Do you need to override `WriteAttributes` ?

Source/Document Structure/SvgScript.cs
```diff
@@ -0,0 +1,81 @@
...
+            get { return _href; }
+            set { _href = value; }
+        }
+    
+        public override SvgElement DeepCopy()
+        {
+            return DeepCopy<SvgScript>();        
+        }
+
+        public override bool ShouldWriteElement()
```
H1Gdev: I think that it is not necessary because there is `SvgElementAttribute`.

Source/Document Structure/SvgScript.cs
gvheertum: Good point, autocomplete did this. Fixed.

Source/Document Structure/SvgScript.cs
gvheertum: I did not test the behaviour without the override, but I did have some troubles with the XML generation with regard to the attributes not being written, so I tinkered around with the code. The ShouldeWriteElement indeed is not necessary, so I dropped it in the new commit.

Source/Document Structure/SvgScript.cs
gvheertum: The code did not correctly output the attributes in the previous setup, so yeah at that moment I did. However using the attributes collection makes this code no longer necessary.

Source/Document Structure/SvgScript.cs
gvheertum: Refactored this. I think I used a wrong example to base this code on. Reworked this to use the attribute collection, this also removed the need to manually output the attributes.

gvheertum:
@H1Gdev good remarks, I mad the changes you suggested. Many of the changes were due to me using an incorrect (or outdated) element to base my code on.

Source/Document Structure/SvgScript.cs
H1Gdev: Have you forgotten this ?

gvheertum: 😊 whoops, missed that one. Commited a change for this.

PR #562, opened 2019-08-24
Added documentation generation using docfx

mrbean-bremen:
- added generated API doc
- added documentation from Wiki

This is a first try at generating GitHub Pages documentation. I added generated API doc, and copied the Wiki pages over.
This currently only works in my branch due to permission issues.

@tebjan - to get this to work in master, you need to generate a personal GitHub access token with public repo access (GitHub settings / Developer settings / Personal access tokens), encrypt it using [Appveyor](https://ci.appveyor.com/tools/encrypt), and replace the current token in appveyor.yml.
You also may have to enable GitHub pages on the repository settings site (options page, scroll down to GitHub Pages, and select "gh-pages branch").

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# SVG.NET [![NuGet version](https://badge.fury.io/nu/svg.svg)](https://badge.fury.io/nu/svg)
+# SVG.NET [![NuGet version](https://badge.fury.io/nu/svg.svg)](https://badge.fury.io/nu/svg) [![Gitter](https://badges.gitter.im/vvvv/SVG.svg)](https://gitter.im/vvvv/SVG?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
 Public fork of the C# SVG rendering library on codeplex: https://svg.codeplex.com/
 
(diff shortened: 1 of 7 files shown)


mrbean-bremen:
Ah, and the created page is [here](https://mrbean-bremen.github.io/SVG/).

gvheertum:
Looking great!

mrbean-bremen:
ping @tebjan

mrbean-bremen:
@tebjan - if you could enable gh-pages on the project (see original comment), we could at least update the documentation manually, even without the automatic stuff.

tebjan:
@mrbean-bremen GitHub Pages is on and the source is configured as gh-pages branch. other options are master or the /docs folder on master. options are explained here: https://help.github.com/en/articles/configuring-a-publishing-source-for-your-github-pages-site
let me know what would work best.

mrbean-bremen:
@tebjan - thanks! That is actually what I meant, though I would have expected the page showing up, as there is already a valid [gh-pages branch](https://github.com/vvvv/SVG/tree/gh-pages). The site should be [here](https://vvvv.github.io/SVG/), but something seems to be still wrong. Could you check on the settings site if this is the correct URL?

mrbean-bremen:
At the settings page for my fork, it states: 
```Your site is published at https://mrbean-bremen.github.io/SVG/```
so I would expect a similar entry in the settings for the main repo.

tebjan:
yes indeed, somehow it didn't get the trigger from the new branch, i had to switch to another source and then back to gh-pages branch. now it works: http://vvvv.github.io/SVG/

mrbean-bremen:
Ah cool - thanks! While you are at it, you could also replace the GitHub access token with your own - or I could just scrub the automatic check-in stuff and go for manual updates.

PR #567, opened 2019-08-30
Get Fizzler 1.1.0 from NuGet.

### LINKED ISSUE #563: Licensing Issue
Not a bug report. Was trying to fork this library and spin it for different uses (my own is currently big and will only grow bigger!, needed something smaller to start with) but had issues with the licensing.
@mrbean-bremen @erdomke @atifaziz

The Fizzler library is licensed under LGPL3+GPL3
https://github.com/vvvv/SVG/blob/master/Source/External/Fizzler/Selector.cs

Can the combined sources be licensed under MS-PL?
The original sources on Codeplex and under MS-PL never included the CSS library.

H1Gdev:
- version 1.1.0.
- get from NuGet.
- fixes #563.

#### Reference Issue

Fixes #563 Fizzler

- Output [NU1701 warning](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1701).


gvheertum:
I did a check on the compatibility for non Windows OS since the embedded package is a .NET Framework 4.6.1 package, which could be problematic for Linux and Mac.

As expected the compatibility message is shown on the restore and build:

> warning NU1701: Package 'Fizzler 1.1.0' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.2'. This package may not be fully compatible with your project.

But the tests and library seem to work just as fine as it did before. So from the non-Windows side: "good job! I approve this PR 👍 ".

To be honest, I am surprised, did not expect the core framework to be able to work with this, but seems the folks at Microsoft thought this one through 😄

H1Gdev:
@gvheertum

Thank you for checking on non-Windows !!

gvheertum:
@mrbean-bremen, since we are referencing this as a separate package I don't think we need to have an additional license. When adding the Fizzler package the user will be shown the own license of the package.

Regarding the mention, I think we can put something in the Readme.md of our project regarding the licenses that apply. I tried looking for examples, but haven't found a suitable example. I think we can have a paragraph regarding the license (stating our own license) and then also mention the dependent projects and their licenses. I'll get back on that topic soon.

mrbean-bremen:
> When adding the Fizzler package the user will be shown the own license of the package.

Ah, thanks,  that sounds right! I will merge in this case.

> Regarding the mention, I think we can put something in the Readme.md of our project regarding the licenses that apply. 

Yes, I agree. It may not be needed, but it's good style to mention other people's work we use.

gvheertum:
@mrbean-bremen I think we should add a part in the readme.md stating our license and mention the dependent projects.

### License
Licensed under the MS-PL license.

This project has dependencies on other open-source projects. These projects are referenced via NuGet packages and might be subject to different licenses.

|Project|Url|License|
|Fizzler|Link to git|LGPL|
|ExCSS|Link to forked git|MIT|

(The PR-reaction are not allowing me to add a nicelooking table, but I guess you get what I mean)

mrbean-bremen:
Yes, that looks good to me (fixed that table for you 😛).

gvheertum:
> Yes, that looks good to me (fixed that table for you 😛).

Haha, thanks! 😃 
Are you planning to put that in a PR (or append this change to your PR) or do you want me to do a PR for the change in the readme.md?

mrbean-bremen:
Well, you almost did this already, so go ahead, if you want!

mrbean-bremen:
@H1Gdev - after adding Fizzler as NuGet, we get [these warnings](https://ci.appveyor.com/project/tebjan/svg/builds/27095663/messages). I guess this is not a problem, though it would be nice if we could avoid them...

PR #577, opened 2019-09-11
Remove SvgDefaults class.

H1Gdev:
#### Reference Issue

Ref. #539 etc.

Remove `SvgDefaults` class.

This PR makes it unnecessary to manage default values in multiple source codes.
(`SvgDefaults` class and attribute properties in element class source code.)

For example:

https://github.com/vvvv/SVG/blob/721e13a795511130c8c14597ce25915e826edb57/Source/SvgDefinitionDefaults.cs#L56

https://github.com/vvvv/SVG/blob/721e13a795511130c8c14597ce25915e826edb57/Source/SvgElementStyle.cs#L136

#### Any other comments?

This PR alone is not enough for DeepCopy.
I will fix DeepCopy as soon as possible.


Source/SvgElement.cs
```diff
@@ -701,7 +701,7 @@ protected virtual void WriteAttributes(XmlTextWriter writer)
                                 }
 
                                 //Only write the attribute's value if it is not the default value, not null/empty, or we're forcing the write.
-                                if (forceWrite || (!string.IsNullOrEmpty(value) && !SvgDefaults.IsDefault(attr.Attribute.Name, attr.Property.ComponentType.Name, value)))
+                                if (forceWrite || !string.IsNullOrEmpty(value))
```
mrbean-bremen: Where it is ensured now that default attributes are not written?

H1Gdev: @mrbean-bremen 

It changes so that **set attribute property value** is output.

For example :

https://github.com/vvvv/SVG/blob/8307612b0d70feb11a0196fdb78bfc764b3edf3a/Source/SvgDefinitionDefaults.cs#L38-L42

- Input SVG

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="500" height="210" version="1.1">
<line x1="0" y1="0" x2="200" y2="200" style="stroke:rgb(255,0,0);stroke-width:2" />
</svg>
```

- Output (latest)
```svg
<svg width="500" height="210" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1">
  <line x2="200" y2="200" stroke-width="2" style="stroke:red;" />
</svg>
```

- Output(This PR)
```svg
<svg width="500" height="210" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1">
  <line x1="0" y1="0" x2="200" y2="200" stroke-width="2" style="stroke:red;" />
</svg>
```

mrbean-bremen: Ok, thanks. I think this is ok, even if it increases the size of the output file.

PR #579, opened 2019-09-13
Add Namespace at href attribute.

H1Gdev:

Add Namespace at `href` attribute.

https://www.w3.org/TR/SVG11/script.html#ScriptElementHrefAttribute

### DIFF
--- a/Source/Document Structure/SvgScript.cs
+++ b/Source/Document Structure/SvgScript.cs
@@ -1,5 +1,3 @@
-using System.Drawing;
-using System.Drawing.Drawing2D;
 using System.Xml;
 
 namespace Svg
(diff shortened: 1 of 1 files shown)


Source/Document Structure/SvgScript.cs
```diff
@@ -11,46 +9,43 @@ namespace Svg
...
-        public string CrossOrigin 
+        public string CrossOrigin
         {
-            get { return GetAttribute<string>("crossorigin",false); }
+            get { return GetAttribute<string>("crossorigin", false); }
             set { Attributes["crossorigin"] = value; }
         }
 
-        [SvgAttribute("href")]
+        [SvgAttribute("href", SvgAttributeAttribute.XLinkNamespace)]
```
H1Gdev: FYI: Deprecated in SVG2.

https://www.w3.org/TR/SVG2/linking.html#XLinkRefAttrs

PR #585, opened 2019-09-25
Fix: If value is 'inherit', stroke-dasharray disappears.

H1Gdev:
#### Reference Issue

Ref. #520 and #504.

##### SVG

```svg
<rect x="10" y="10" width="180" height="180" fill="red" stroke="black" stroke-width="4" stroke-dasharray="inherit" />
```

##### Expected
```svg
<rect x="10" y="10" width="180" height="180" stroke-width="4" stroke-dasharray="inherit" style="fill:red;stroke:black;" />
```

##### Actual
```svg
<rect x="10" y="10" width="180" height="180" stroke-width="4" style="fill:red;stroke:black;" />
```

### DIFF
--- a/Source/DataTypes/SvgPointCollection.cs
+++ b/Source/DataTypes/SvgPointCollection.cs
@@ -1,9 +1,7 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
+﻿using System.Collections.Generic;
 using System.ComponentModel;
 using System.Globalization;
+using System.Text;
 
 namespace Svg
 {
@@ -35,13 +33,10 @@ namespace Svg
     }
 
     /// <summary>
-    /// A class to convert string into <see cref="SvgUnitCollection"/> instances.
+    /// A class to convert string into <see cref="SvgPointCollection"/> instances.
     /// </summary>
     internal class SvgPointCollectionConverter : TypeConverter
     {
-        //private static readonly SvgUnitConverter _unitConverter = new SvgUnitConverter();
-
-
         /// <summary>
         /// Converts the given object to the type of this converter, using the specified context and culture information.
         /// </summary>
@@ -52,14 +47,11 @@ namespace Svg
         /// An <see cref="T:System.Object"/> that represents the converted value.
         /// </returns>
         /// <exception cref="T:System.NotSupportedException">The conversion cannot be performed. </exception>
-        public override object ConvertFrom(ITypeDescriptorContext context, System.Globalization.CultureInfo culture, object value)
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
             if (value is string)
             {
-                var strValue = ((string)value).Trim();
-                if (string.Compare(strValue, "none", StringComparison.InvariantCultureIgnoreCase) == 0) return null;
-
-                var parser = new CoordinateParser(strValue);
+                var parser = new CoordinateParser(((string)value).Trim());
                 var pointValue = 0.0f;
                 var result = new SvgPointCollection();
                 while (parser.TryGetFloat(out pointValue))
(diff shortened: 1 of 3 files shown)


Source/DataTypes/SvgUnitCollection.cs
```diff
@@ -13,7 +13,7 @@ namespace Svg
     [TypeConverter(typeof(SvgUnitCollectionConverter))]
     public class SvgUnitCollection : ObservableCollection<SvgUnit>
     {
-        internal bool IsNull { get; set; }
+        public string StringIfEmpty { get; set; }
```
H1Gdev: set `none` or `inherit`.

Source/DataTypes/SvgUnitCollection.cs
```diff
@@ -77,45 +80,43 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
         {
             if (value is string)
             {
-                var s = ((string)value).Trim();
-
-                if (s.Equals("inherit", StringComparison.OrdinalIgnoreCase))
-                    return null;
-
+                var points = ((string)value).Trim().Split(new char[] { ',', ' ', '\r', '\n', '\t' }, StringSplitOptions.RemoveEmptyEntries);
```
H1Gdev: Parse `list-of-coordinates` only.

Source/DataTypes/SvgUnitCollection.cs
```diff
@@ -77,45 +80,43 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
+                }
 
                 return units;
             }
 
             return base.ConvertFrom(context, culture, value);
         }
+    }
 
-        public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
```
H1Gdev: If destinationType is string, base class CanConvertTo always returns true and ConvertTo calls ToString.

Source/DataTypes/SvgUnitCollection.cs
```diff
@@ -77,45 +80,43 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
+    {
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
-            if (value != null && destinationType == typeof(string))
+            if (value is string)
             {
-                if (((SvgUnitCollection)value).IsNull)
-                    return "none";
-                return ((SvgUnitCollection)value).ToString();
+                var s = ((string)value).Trim();
```
H1Gdev: Only [stroke-dasharray](https://www.w3.org/TR/SVG11/painting.html#StrokeProperties) can set none and inherit.

Source/DataTypes/SvgPointCollection.cs
```diff
@@ -52,14 +47,11 @@ internal class SvgPointCollectionConverter : TypeConverter
         /// An <see cref="T:System.Object"/> that represents the converted value.
         /// </returns>
         /// <exception cref="T:System.NotSupportedException">The conversion cannot be performed. </exception>
-        public override object ConvertFrom(ITypeDescriptorContext context, System.Globalization.CultureInfo culture, object value)
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
             if (value is string)
             {
-                var strValue = ((string)value).Trim();
-                if (string.Compare(strValue, "none", StringComparison.InvariantCultureIgnoreCase) == 0) return null;
```
H1Gdev: `list-of-points` not set `none`.

Source/DataTypes/SvgUnitCollection.cs
mrbean-bremen: Should this be public? Wouldn't `internal` be better? Also I'm not sure about the name, it may be a bit misleading, though I'm struggling to find a better one, too. Maybe `StringForEmptyValue` or something like this?

Other that that, it all looks good - thanks!

H1Gdev: Thank you for your review.

> Should this be public?

Make `public` because it is necessary when generating SVG from user code.

> Also I'm not sure about the name, it may be a bit misleading, though I'm struggling to find a better one, too. Maybe StringForEmptyValue or something like this?

I use in sense that collection is empty.(like IsEmpty)

Source/DataTypes/SvgPointCollection.cs
```diff
@@ -52,14 +47,11 @@ internal class SvgPointCollectionConverter : TypeConverter
...
-
-                var parser = new CoordinateParser(strValue);
+                var parser = new CoordinateParser(((string)value).Trim());
                 var pointValue = 0.0f;
                 var result = new SvgPointCollection();
```
H1Gdev: Why not `SvgPointCollection` a `SvgPoint` list?

Source/DataTypes/SvgUnitCollection.cs
mrbean-bremen: Ah, ok - naming is hard, let's go with that then. Thanks!

Issue #541, opened 2019-08-04
Not all properties should be inherited


wieslawsoltes:
#### Description

Not all properties should be inherited: https://www.w3.org/TR/SVG11/propidx.html

#### Example data

`SvgElementStyle.cs`

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/SvgElementStyle.cs#L148

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/SvgElementStyle.cs#L158

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/SvgElementStyle.cs#L188

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/SvgElementStyle.cs#L238

`SvgVisualElement.cs`

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/Basic%20Shapes/SvgVisualElement.cs#L52

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/Basic%20Shapes/SvgVisualElement.cs#L82

`SvgVisualElementStyle.cs`

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/Basic%20Shapes/SvgVisualElementStyle.cs#L26

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/Basic%20Shapes/SvgVisualElementStyle.cs#L42

`SvgGradientStop.cs`

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/Painting/SvgGradientStop.cs#L63

https://github.com/vvvv/SVG/blob/f21eebf029e27967587180853ae22d978502cfe1/Source/Painting/SvgGradientStop.cs#L73

#### Used Versions

SVG 1.1 (Second Edition) – 16 August 2011

H1Gdev:
@wieslawsoltes 

`opacity` attribute requires ` inherit`.(`filter`, too.)

In which case is problem happening ?

wieslawsoltes:
When using filter on `g` or `use`.

wieslawsoltes:
Sample for `opacity`
```svg
  <g transform="translate(80, 80)"                    opacity="0.5">
[...]

PR #587, opened 2019-10-05
Redesign of 'inherit' and 'Inherited'.

H1Gdev:
#### Reference Issue

Ref. #541.

### DIFF
--- a/Source/SvgAttributeCollection.cs
+++ b/Source/SvgAttributeCollection.cs
@@ -39,7 +39,7 @@ namespace Svg
         /// </summary>
         /// <typeparam name="TAttributeType">The type of the attribute value.</typeparam>
         /// <param name="attributeName">A <see cref="string"/> containing the name of the attribute.</param>
-        /// <param name="inherited">The attribute is inherited or not.</param>
+        /// <param name="inherited">The attribute value if available and not set to "inherit"; the ancestors value for the same attribute if it exists and if either the attribute value is set to "inherit", or <paramref name="inherited"/> is true; the default value otherwise.</param>
         /// <param name="defaultValue">The value to return if a value hasn't already been specified.</param>
         /// <returns>The attribute value if available; otherwise the ancestors value for the same attribute; otherwise the default value of <typeparamref name="TAttributeType"/>.</returns>
         public TAttributeType GetInheritedAttribute<TAttributeType>(string attributeName, bool inherited, TAttributeType defaultValue = default(TAttributeType))
--- a/Source/SvgAttributeCollection.cs
+++ b/Source/SvgAttributeCollection.cs
@@ -41,7 +41,7 @@ namespace Svg
         /// <param name="attributeName">A <see cref="string"/> containing the name of the attribute.</param>
         /// <param name="inherited">Used only if the attribute value is not available. If set to true, the inherited value is returned in this case, otherwise the default value.</param>
         /// <param name="defaultValue">The value to return if a value hasn't already been specified.</param>
-        /// <returns>The attribute value if available; otherwise the ancestors value for the same attribute; otherwise the default value of <typeparamref name="TAttributeType"/>.</returns>
+        /// <returns>The attribute value if available and not set to "inherit"; the ancestors value for the same attribute if it exists and if either the attribute value is set to "inherit", or <paramref name="inherited"/> is true; the default value otherwise.</returns>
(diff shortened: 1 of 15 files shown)


Source/SvgAttributeCollection.cs
```diff
@@ -74,7 +90,7 @@ private bool IsInheritValue(object value)
         /// <returns>The attribute value associated with the specified name; If there is no attribute the parent's value will be inherited.</returns>
         public new object this[string attributeName]
         {
-            get { return GetInheritedAttribute<object>(attributeName); }
+            get { return GetInheritedAttribute<object>(attributeName, true); }
```
H1Gdev: Indexer no longer returns correct value, but remains for compatibility.

mrbean-bremen:
@wieslawsoltes - can you please have a look at this? I think you are more involved here...

H1Gdev:
@mrbean-bremen 

Take `opacity` as an example.

https://www.w3.org/TR/SVG11/masking.html#ObjectAndGroupOpacityProperties

```
‘opacity’
Value: <opacity-value> | inherit
Inherited: no
```

This attribute does not inherit unless value is `inherit`.

H1Gdev:
Fix of `opacity` issue is planned after this PR.

Source/SvgAttributeCollection.cs
```diff
@@ -39,32 +39,48 @@ public TAttributeType GetAttribute<TAttributeType>(string attributeName, TAttrib
         /// </summary>
         /// <typeparam name="TAttributeType">The type of the attribute value.</typeparam>
         /// <param name="attributeName">A <see cref="string"/> containing the name of the attribute.</param>
+        /// <param name="inherited">The attribute is inherited or not.</param>
```
mrbean-bremen: I would like a better description here. As I understand, `inherited` is only used if the value hasn't been specified yet, to differentiate between using the inherited value, if any, and the default value.

Source/SvgAttributeCollection.cs
```diff
@@ -39,32 +39,48 @@ public TAttributeType GetAttribute<TAttributeType>(string attributeName, TAttrib
         /// </summary>
         /// <typeparam name="TAttributeType">The type of the attribute value.</typeparam>
         /// <param name="attributeName">A <see cref="string"/> containing the name of the attribute.</param>
+        /// <param name="inherited">The attribute is inherited or not.</param>
         /// <param name="defaultValue">The value to return if a value hasn't already been specified.</param>
         /// <returns>The attribute value if available; otherwise the ancestors value for the same attribute; otherwise the default value of <typeparamref name="TAttributeType"/>.</returns>
```
mrbean-bremen: Also maybe expand this a bit, as this is not self-explanatary, e.g. something like:
The attribute value if available and not set to "inherit"; the ancestors value for the same attribute if it exists 
and if either the attribute value is set to "inherit", or <paramref name="inherited"/> is true; the default value otherwise.

Tests/Svg.UnitTests/SvgAttributeCollectionTests.cs
```diff
@@ -0,0 +1,47 @@
+﻿using NUnit.Framework;
+
+namespace Svg.UnitTests
+{
+    [TestFixture]
+    public class SvgAttributeCollectionTests
+    {
+        [Test]
+        public void TestGetAttribute()
```
mrbean-bremen: `TestGetInheritedAttribute` ?

H1Gdev: I changed test name to method name.

Source/SvgAttributeCollection.cs
mrbean-bremen: Sorry to be picky, but the description you added for `inherited` actually was meant for the return value - sorry I wasn't more clear about that.
The description for `inherited` would be something like:
```Used only if the attribute value is not available. If set to true, the inherited value is returned in this case, otherwise the default value.```

mrbean-bremen: ...and the comment from above belongs to `params`:

```The attribute value if available and not set to "inherit"; the ancestors value for the same attribute if it exists and if either the attribute value is set to "inherit", or <paramref name="inherited"/> is true; the default value otherwise.```

I will be more specific with documentation comments in the future to avoid these misunderstandings...

PR #599, opened 2019-10-22
Update DeepCopy method of each element.

H1Gdev:

Implementation of `DeepCopy` method of each element class copied attribute via property.
As a result, default value was set at copy destination.

This PR fixes this issue.

### DIFF
--- a/Source/DataTypes/SvgAspectRatio.cs
+++ b/Source/DataTypes/SvgAspectRatio.cs
@@ -1,5 +1,6 @@
-﻿using Svg.DataTypes;
+using System;
 using System.ComponentModel;
+using Svg.DataTypes;
 
 namespace Svg
 {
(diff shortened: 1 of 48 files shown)


Source/DataTypes/SvgAspectRatio.cs
```diff
@@ -1,13 +1,14 @@
...
+using Svg.DataTypes;
 
 namespace Svg
 {
     /// <summary>
     /// Description of SvgAspectRatio.
     /// </summary>
     [TypeConverter(typeof(SvgPreserveAspectRatioConverter))]
-    public class SvgAspectRatio
+    public class SvgAspectRatio : ICloneable
```
H1Gdev: Implement `ICloneable` for DeepCopy.

Source/SvgElement.cs
```diff
@@ -968,29 +968,29 @@ ISvgNode ISvgNode.DeepCopy()
 
         public virtual SvgElement DeepCopy<T>() where T : SvgElement, new()
         {
-            var newObj = new T();
-            newObj.ID = this.ID;
```
mrbean-bremen: I guess the ID does not need to be copied?

H1Gdev: `ID` is an attribute, so it is copied.

mrbean-bremen: Ah yes, thanks!

PR #621, opened 2019-12-12
Add new line joins defined by the SVG standard

JoaoBaptMG:

This pull request adds support for two new line joins defined by the SVG standard as stated [here](https://svgwg.org/svg2-draft/painting.html#LineJoin): miter-clip and arcs.

#### Any other comments?
The SVG WG documentation stated that those line joins are at risk because there is no browser implementation for it. I added support for them in [my project of svg renderer](https://github.com/JoaoBaptMG/PathRenderingLab/blob/master/PathRenderingLab/PathCompiler/StrokeUtils.cs#L248), so I would like to have support for them in the SVG loader.

### DIFF
--- a/Source/Basic Shapes/SvgVisualElement.cs
+++ b/Source/Basic Shapes/SvgVisualElement.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Drawing2D;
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
@JoaoBaptMG - thanks, this looks good to me! Do you have a simple example image to add to the test suite (e.g. SVG + PNG)?
 @H1Gdev - can you have another look at the PR please?

mrbean-bremen:
@H1Gdev , @wieslawsoltes - can one of you have another look? It seems to be good to merge to me.

JoaoBaptMG:
I am really sorry for not having responded before. The arcs is not really supported by GDI+, so it’s not really possible to test. It was added just because I am using this library to load SVG on my GPU tendering library ~~or at least I was before scrapping it~~.

H1Gdev:
@mrbean-bremen 

`miter-clip` and `arcs` are SVG2 specifications.
How do we implement SVG2 ?

mrbean-bremen:
> How do we implement SVG2 ?

We have a few things from SVG2 already in (don't remember which ones right now), and while we won't advertise it (yet), we could add the support incrementally where needed. 
The only problematic cases would be actual changes in behavior between 1.1 to 2 (instead of added new features), but we can address these later as needed.

mrbean-bremen:
I'm going to merge this - we can add it to the release notes afterwards.

PR #629, opened 2019-12-27
Ignore NuGet Packages

wieslawsoltes:

Ignore restored NuGet packages.

#### Any other comments?

The `.gitignore` rules are taken from default `.gitignore` for Visual Studio repositories:
```
# NuGet Packages
*.nupkg
# The packages folder can be ignored because of Package Restore
**/packages/*
# except build/, which is used as an MSBuild target.
!**/packages/build/
# Uncomment if necessary however generally it will be regenerated when needed
#!**/packages/repositories.config
# NuGet v3's project.json files produces more ignorable files
*.nuget.props
*.nuget.targets
```


H1Gdev:
A better way is to use [VisualStudio.gitignore](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) from [github/gitignore](https://github.com/github/gitignore).

wieslawsoltes:
Yes for sure.

wieslawsoltes:
> A better way is to use [VisualStudio.gitignore](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) from [github/gitignore](https://github.com/github/gitignore).

Update PR with VisualStudio.gitignore

PR #630, opened 2019-12-31
Fix some issues for netstandard2.0 target

wieslawsoltes:
#### Reference Issue
There are missing features (e.g. ColorConverter) when using `.NETStandard 2.0` target and using different executable than supported `netcoreapp2.2` or newer and `net452` or newer (e.g. Android/iOS/UWP).

I have ported `ColorConverter` class with minor fixes from https://github.com/dotnet/runtime/blob/master/src/libraries/System.Drawing.Common/src/System/Drawing/ColorConverter.cs

#### Any other comments?

I had to update `System.Drawing.Common` package.

### DIFF
--- a/Source/System/Drawing/ColorConverter.cs
+++ b/Source/System/Drawing/ColorConverter.cs
@@ -1,0 +1,433 @@
+﻿// https://github.com/dotnet/runtime/blob/master/src/libraries/System.Drawing.Common/src/System/Drawing/ColorConverter.cs
+#if NETSTANDARD20
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Drawing {
+    using System.Runtime.Serialization.Formatters;
+    using System.Runtime.InteropServices;
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using Microsoft.Win32;
+    using System.Collections;
+    using System.ComponentModel;
+    using System.ComponentModel.Design.Serialization;
+    using System.Globalization;
+    using System.Reflection;
+    using System.Threading;
+
+    /// <include file='doc\ColorConverter.uex' path='docs/doc[@for="ColorConverter"]/*' />
+    /// <devdoc>
+    ///      ColorConverter is a class that can be used to convert
+    ///      colors from one data type to another.  Access this
+    ///      class through the TypeDescriptor.
+    /// </devdoc>
+    public class ColorConverter : TypeConverter {
+        private static string ColorConstantsLock = "colorConstants";
+        private static Hashtable colorConstants;
+        private static string SystemColorConstantsLock = "systemColorConstants";
+        private static Hashtable systemColorConstants;
+        private static string ValuesLock = "values";
+        private static StandardValuesCollection values;
+
+        /// <include file='doc\ColorConverter.uex' path='docs/doc[@for="ColorConverter.ColorConverter"]/*' />
+        /// <devdoc>
+        ///    <para>[To be supplied.]</para>
+        /// </devdoc>
+        public ColorConverter() {
+        }
+
+        /// <include file='doc\ColorConverter.uex' path='docs/doc[@for="ColorConverter.Colors"]/*' />
+        /// <devdoc>
+        ///      Hashtable of color / value pairs (color name is key)
+        ///      for standard colors.
+        /// </devdoc>
+        private static Hashtable Colors {
+            get {
+                if (colorConstants == null) {
+                    lock(ColorConstantsLock) {
+                        if (colorConstants == null) {
+                            Hashtable tempHash = new Hashtable(StringComparer.OrdinalIgnoreCase);
+                            FillConstants(tempHash, typeof(Color));
+                            colorConstants = tempHash;
+                        }
+                    }
+                }
+
+                return colorConstants;
+            }
+        }
+
+        /// <include file='doc\ColorConverter.uex' path='docs/doc[@for="ColorConverter.SystemColors"]/*' />
+        /// <devdoc>
+        ///      Hashtable of color / value pairs (color name is key)
+        ///      for system colors.
+        /// </devdoc>
+        private static Hashtable SystemColors {
+            get {
+                if (systemColorConstants == null) {
+                    lock (SystemColorConstantsLock) {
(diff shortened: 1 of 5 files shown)


Source/System/Drawing/ColorConverter.cs
```diff
@@ -0,0 +1,433 @@
...
+                        // name.  Otherwise, we fabricate an ARGB value for it.
+                        //
+#if NETSTANDARD20
+                        if (c.IsNamedColor) {
+                            return "'" + c.Name + "'";
+                        }
+#else
+                        if (c.IsKnownColor) {
+                            return c.Name;
+                        }
```
mrbean-bremen: So basically you have removed this check - don't we have `IsKnownColor` (and `IsSystemColor` below)?

Source/System/Drawing/ColorConverter.cs
```diff
@@ -0,0 +1,433 @@
...
+#if NETSTANDARD20
+                    if (obj == null)
+                    {
+                        throw new ArgumentException(nameof(value));
+                    }
+#else
+                    if (obj == null) {
+                        throw new ArgumentException(SR.Format(SR.InvalidColor, text));
+                    }
+#endif
```
mrbean-bremen: I guess you use this to retain the original code and to mark all changes to the copied code, right?

Source/System/Drawing/ColorConverter.cs
wieslawsoltes: It's also removed using if/def when using netstandard2.0 target framework.

Source/System/Drawing/ColorConverter.cs
wieslawsoltes: Yes, that's correct.

Source/System/Drawing/ColorConverter.cs
mrbean-bremen: Ah ok, obviously it's not defined there. Anyway, this is not important, as it just means that hex codes will be used instead of color names.

wieslawsoltes:
@mrbean-bremen

> Very nice! Does that mean that we will have complete netstandard2.0 support after this change?

As much as it's currently possible. There are few properties missing from Color (IsKnownColor and IsNamedColor), fixing those is not as straightforward as this PR. Also some stuff may not work on not supported platforms by System.Drawing.Common, not much we can do with this right now.

> If that is so, it would probably make sense to make another release soon (somewhere next year, before starting to split out the drawing stuff).

+1 for new release

> Ah, and can you add an entry to the release notes, please?

Added.

mrbean-bremen:
👍

Source/System/Drawing/ColorConverter.cs
```diff
@@ -0,0 +1,433 @@
...
+
+                            foreach (Color c in Colors.Values) {
+                                if (c.ToArgb() == targetARGB) {
+                                    obj = c;
+                                    break;
+                                }
+                            }
+                        }
+                    }
+#if NETSTANDARD20
```
H1Gdev: I have one question.

Whole file seems to be surrounded by `NETSTANDARD20`.
Is this `#if` - `#endif` necessary ?

wieslawsoltes: We could solve this bt excluding this file in csproj for other targets.This file is only needed for netstandatd2.0 target frameworks, other targets defined in csproj has this class available in corefx framework. It is important to note that this is not only about supporting netstandard2.0 but also supporting target frameworks that implement the nonstandard2.0 but are not defined in csproj target frameworks, this includes things like android and iOS targets.

mrbean-bremen:
> As much as it's currently possible. There are few properties missing from Color (IsKnownColor and IsNamedColor), fixing those is not as straightforward as this PR.

I looked it up - these are actually not in netstandard2.0 (have been added in 3.0), so I think nothing has to be done here.

> We could solve this bt excluding this file in csproj for other targets.

This is a possibility, maybe a bit cleaner than the current state, though I don't have strong feelings about it.

Source/System/Drawing/ColorConverter.cs
H1Gdev: @wieslawsoltes 

Thank you for your response.
But I don't know if this is necessary...

ColorConverter.cs
```cs
// https://github.com/dotnet/runtime/blob/master/src/libraries/System.Drawing.Common/src/System/Drawing/ColorConverter.cs
#if NETSTANDARD20
// Licensed to the .NET Foundation under one or more agreements.
...
#if NETSTANDARD20
...
#else
...
#endif
...
#endif
```

If `NETSTANDARD20` is undefined, it looks like entire file is disabled.

wieslawsoltes: I prefer if/defs in .cs file than messing with csproj and conditional compilation.

mrbean-bremen: I think the question was about the nested ifdefs, which you has answered [above](https://github.com/vvvv/SVG/pull/630#discussion_r362266539).

H1Gdev: @mrbean-bremen 

Thanks !

@wieslawsoltes 

I understood.
You changed `ColorConverter.cs` from `dotnet/runtime` to solve this issue.
This if-else-endif was confusing to me...
It's not original file, so please put a hint at top of file.

For example:
```cs
// This file is based on
// https://github.com/dotnet/runtime/blob/master/src/libraries/System.Drawing.Common/src/System/Drawing/ColorConverter.cs
// 6a988c7d0389bf9e9aa20d23baa353e9393b2ea5
```

wieslawsoltes: I will submit updated comment as proposed.

mrbean-bremen:
@wieslawsoltes  - one thing I forgot to mention: all these `<include file='doc\ColorConverter.uex` lines currently give warnings during the build. I don't think we have to preserve these - I would just remove them. What do you think?

wieslawsoltes:
> @wieslawsoltes - one thing I forgot to mention: all these `<include file='doc\ColorConverter.uex` lines currently give warnings during the build. I don't think we have to preserve these - I would just remove them. What do you think?

Yes we can remove those warnings.

wieslawsoltes:
> @wieslawsoltes - one thing I forgot to mention: all these `<include file='doc\ColorConverter.uex` lines currently give warnings during the build. I don't think we have to preserve these - I would just remove them. What do you think?

@mrbean-bremen  https://github.com/vvvv/SVG/pull/636

PR #635, opened 2020-01-08
Xamarin Forms UWP support for netstandard2.0

wieslawsoltes:

Fixes exceptions thrown when using `netstandard2.0` and `Xamarin Forms` `UWP` + `Svg.Skia` rendering.

ApplicationFrameHost_2020-01-08_18-22-30

#### Any other comments?

Added suggested comment from review:
https://github.com/vvvv/SVG/pull/630#discussion_r363126766

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -206,7 +206,9 @@ namespace Svg
         {
             try
             {
+#if !NETSTANDARD20
                 using (var matrix = new Matrix(0f, 0f, 0f, 0f, 0f, 0f)) { }
+#endif
             }
             // GDI+ loading errors will result in TypeInitializationExceptions, 
             // for readability we will catch and wrap the error
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -206,9 +206,7 @@ namespace Svg
         {
             try
             {
-#if !NETSTANDARD20
                 using (var matrix = new Matrix(0f, 0f, 0f, 0f, 0f, 0f)) { }
-#endif
             }
             // GDI+ loading errors will result in TypeInitializationExceptions, 
             // for readability we will catch and wrap the error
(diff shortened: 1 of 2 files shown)


Source/SvgDocument.cs
```diff
@@ -206,7 +206,9 @@ public static void EnsureSystemIsGdiPlusCapable()
         {
             try
             {
+#if !NETSTANDARD20
                 using (var matrix = new Matrix(0f, 0f, 0f, 0f, 0f, 0f)) { }
+#endif
```
mrbean-bremen: @gvheertum - this removes the libgdiplus check for netstandard2, do you think this is a problem, or can be solved another way?

Source/System/Drawing/ColorConverter.cs
```diff
@@ -114,7 +118,9 @@ public class ColorConverter : TypeConverter {
             }
             // Ok, how about a system color?
             //
+#if !NETSTANDARD20
             color = SystemColors[name];
+#endif
```
mrbean-bremen: It seems a bit inconsistent - `Colors` is changed to return an empty hashtable and is used as before, while `SystemColors` is not changed and instead the usage is commented out. I would probably handle `SystemColors` the same way as `Colors`.

Source/SvgDocument.cs
wieslawsoltes: This check fails on netstandard2.0 and UWP (and probably other not supported netstandard2 targets). I have tested on UWP with Svg.Skia and removing this check enables Svg.Skia to work and System.Drawing (libgdiplus ) is not supported anyway on UWP. For now this is best solution I have found.

Source/System/Drawing/ColorConverter.cs
wieslawsoltes: Actually Colors should not be commented, instead SystemColors should be. Pushed fix.

Source/SvgDocument.cs
wieslawsoltes: I have removed if/def from EnsureSystemIsGdiPlusCapable and instead added it to Open method.

gvheertum: I'm wondering what happens if a Matrix is encountered in that case. What kind of error did you get that made you decide to remove the check? If the Matrix cannot be loaded I expect issues lateron when Matrix related stuff is encountered.

wieslawsoltes: I am not using System.Drawing part of SVG so I don't need Matrix and other classes from gdi. This check prevented me from using SVG when using netstandard2.0 version on platforms like UWP. Util we have splitted drawing from object model processing that's only solution I think.

gvheertum: Fair point, however this will break all other cases where people are using .NET Standard with graphics enabled. This will re-open the issues that caused us to put the check in (the exception that was thrown was not that clear). So honestly I am not a big fan of disabling the check for .NET standard as a whole.

Perhaps we can add a static flag somewhere in the SVG lib that you can set to false to disable the check? In that case we can also put warnings in the xml doc of the flag. Something like Svg.Configuration.SkipGDIPlusCapabilityCheck. In that case we will keep the check in the normal use-case and will be able to skip it when explicitly set.

wieslawsoltes: > static flag

I will submit PR with static flag proposal.

PR #637, opened 2020-01-09
Use static flag to disable GDI+ check

wieslawsoltes:

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -24,7 +24,7 @@ namespace Svg
     public class SvgDocument : SvgFragment, ITypeDescriptorContext
     {
         /// <summary>
-        /// Skip check  whether the GDI+ can be loaded.
+        /// Skip check whether the GDI+ can be loaded.
         /// </summary>
         /// <remarks>
         /// Set to true on systems that do not support GDI+ like UWP.
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -29,7 +29,8 @@ namespace Svg
         /// <remarks>
         /// Set to true on systems that do not support GDI+ like UWP.
         /// </remarks>
-        public static bool SkipGdiPlusCapabilityCheck = false;
+        public static bool SkipGdiPlusCapabilityCheck { get; set; }
+
         public static readonly int PointsPerInch = GetSystemDpi();
         private SvgElementIdManager _idManager;
 
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -23,6 +23,13 @@ namespace Svg
     /// </summary>
     public class SvgDocument : SvgFragment, ITypeDescriptorContext
     {
+        /// <summary>
+        /// Skip check  whether the GDI+ can be loaded.
+        /// </summary>
+        /// <remarks>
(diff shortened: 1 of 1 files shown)


Source/SvgDocument.cs
```diff
@@ -23,6 +23,13 @@ namespace Svg
     /// </summary>
     public class SvgDocument : SvgFragment, ITypeDescriptorContext
     {
+        /// <summary>
+        /// Skip check whether the GDI+ can be loaded.
+        /// </summary>
+        /// <remarks>
+        /// Set to true on systems that do not support GDI+ like UWP.
+        /// </remarks>
+        public static bool SkipGDIPlusCapabilityCheck = false;
```
H1Gdev: Use auto-implemented property.

H1Gdev: Is if-def unnecessary ?
```cs
#if NETSTANDARD20
public static bool SkipGDIPlusCapabilityCheck = true;
#else
public static bool SkipGDIPlusCapabilityCheck = true;
#endif
```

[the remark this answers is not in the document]
gvheertum: My bad I guess. Ik keep mixing GDI and Gdi since starting on the original fix

gvheertum: Yep, I suggest keeping the check on false unless explicitly set to false by the caller. We should for now assume that the callee will use some of the graphical elements, thus requiring us to check the GDI capability. If the callee is absolutely sure to not use the graphic elements he can ommit the check

Source/SvgDocument.cs
```diff
@@ -349,9 +356,10 @@ public static T Open<T>(Stream stream, Dictionary<string, string> entities) wher
 
         private static T Open<T>(XmlReader reader) where T : SvgDocument, new()
```
gvheertum: I am wondering whether we should not also allow the flag to be set on a "per-instance" case, e.g. in the constructor or having a specialized inheritor for this. For now the flag will be fine, but if you use multiple callees of the library you might encounter strange behavior (if you for example use the grapics but also have a component that will set the flag to false this might break the rendering part). Those specific situations are rare (and require you to have one or more sub libraries that use the flag/rendering) and not realy likely, but having system-wide flags might be a problem in the very far future.

wieslawsoltes: `SvgDocument` is created using `SvgElementFactory` so having per instance property would require setting the value of `SkipGdiPlusCapabilityCheck` when creating documents in factory. And you can have documents inside document and referenced external documents also created via factory.

Source/SvgDocument.cs
wieslawsoltes: Fixed in PR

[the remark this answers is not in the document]
wieslawsoltes: > Name is `SkipGdiPlusCapabilityCheck`.

Fixed in PR

Source/SvgDocument.cs
mrbean-bremen: @gvheertum - I agree with @wieslawsoltes that this would be difficult, and that the current state is sufficient. I will merge as is if you can live with that...

gvheertum: I totally agree :) Having to throw the flag all over the place is indeed not the way to go now! So I can totally live with it.

PR #641, opened 2020-01-18
Add filter effects classes

wieslawsoltes:

* Added filter effects classes:
  * [x] feBlend
  * [x] feComponentTransfer
  * [x] feComposite
  * [x] feConvolveMatrix
  * [x] feDiffuseLighting
  * [x] feDisplacementMap
  * [x] feFlood
  * [x] feImage
  * [x] feMorphology
  * [x] feSpecularLighting
  * [x] feTile
  * [x] feTurbulence
* Added component transfer function classes:
  * [x] feFuncA
  * [x] feFuncB
  * [x] feFuncG
  * [x] feFuncR
* Added light source classes:
  * [x] feDistantLight
  * [x] fePointLight
  * [x] feSpotLight

* Set correct default values for SvgFilter properties.
* Added FilterUnits and PrimitiveUnits properties to SvgFilter class.
* Added X, Y, Width and Height properties to SvgFilterPrimitive class.
* Added SvgNumberCollection data type similar to SvgPointCollection.

#### Any other comments?

I would also like to change type of the `StdDeviation` property of the `SvgGaussianBlur` filter effect as it's not following the SVG specification. This is off-course breaking change. I hope at least we can discuss this here as part of this PR ads `SvgNumberCollection` data type and uses it extensively. 

```C#
        [SvgAttribute("stdDeviation")]
        public SvgNumberCollection StdDeviation
        {
            get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f, 0f }); }
            set { Attributes["stdDeviation"] = value; }
        }
```

https://github.com/vvvv/SVG/blob/4580a4e7874afb26c2d0b0b45a68e6b8ae65640c/Source/Filter%20Effects/feGaussianBlur/SvgGaussianBlur.cs#L221-L238

https://www.w3.org/TR/SVG11/filters.html#feGaussianBlurElement
```
stdDeviation = "<number-optional-number>"
The standard deviation for the blur operation. If two <number>s are provided, the first number represents a standard deviation value along the x-axis of the coordinate system established by attribute ‘primitiveUnits’ on the ‘filter’ element. The second value represents a standard deviation in Y. If one number is provided, then that value is used for both X and Y.
A negative value is an error (see Error processing). A value of zero disables the effect of the given filter primitive (i.e., the result is the filter input image). If ‘stdDeviation’ is 0 in only one of X or Y, then the effect is that the blur is only applied in the direction that has a non-zero value.
If the attribute is not specified, then the effect is as if a value of 0 were specified.
Animatable: yes.
```


tebjan:
@wieslawsoltes thanks a lot for working on that. in order to merge the PR and to make it more understandable to other ürpgrammers, please use a more descriptive commit message. something like:  "Added FilterUnits and PrimitiveUnits properties to SvgFilter class"

"Update SvgFilter.cs" doesn't give new info, since the commit already shows that the diff is in that file.

i can highly recommend to use [GitExtensions](https://gitextensions.github.io/) which has text completion when writing a commit message.

wieslawsoltes:
@tebjan ok I will write more descriptive commit messages in the feature.

wieslawsoltes:
@tebjan I have fixed commit messages.

tebjan:
@wieslawsoltes thanks a lot, that looks good!

wieslawsoltes:
I have question about `Process` method for new filter primitive classes. It's out of the scope of this PR to implement filter primitive `Process` methods, is this acceptable solution to be included in PR?
```C#
        public override void Process(ImageBuffer buffer)
        {
            // TODO: Implement feBlend filter Process().
        }
```
https://github.com/wieslawsoltes/SVG/blob/e7a73d9082a0a03e287db813681e234521eae4ef/Source/Filter%20Effects/feBlend/SvgBlend.cs#L22-L25

Source/Filter Effects/feFlood/SvgFlood.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            // TODO: Implement feFlood filter Process().
+        }
+
+        public override SvgElement DeepCopy()
+        {
+            return DeepCopy<SvgFlood>();
+        }
+
+        public override SvgElement DeepCopy<T>()
```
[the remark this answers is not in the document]
wieslawsoltes: Fixed

Source/Filter Effects/feBlend/SvgBlend.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace Svg.FilterEffects
+{
+    [SvgElement("feBlend")]
+    public class SvgBlend : SvgFilterPrimitive
+    {
+        private SvgBlendMode _mode = SvgBlendMode.Normal;
```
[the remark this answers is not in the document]
wieslawsoltes: Fixed

Source/Filter Effects/feBlend/SvgBlend.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            return DeepCopy<SvgGaussianBlur>();
+        }
+
+        public override SvgElement DeepCopy<T>()
+        {
+            var newObj = base.DeepCopy<T>() as SvgBlend;
+
+            newObj._mode = _mode;
+            newObj.Input2 = Input2;
```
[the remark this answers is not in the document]
wieslawsoltes: Fixed

wieslawsoltes:
This PR is ready for initial review. I may add some tweaks along the way while I am trying to implement some of the filters in Svg.Skia.

wieslawsoltes:
I would also like to change type of the `StdDeviation` property of the `SvgGaussianBlur` filter effect as it's not following the SVG specification. This is off-course breaking change. I hope at least we can discuss this here as part of this PR ads `SvgNumberCollection` data type and uses it extensively. 

```C#
        [SvgAttribute("stdDeviation")]
        public SvgNumberCollection StdDeviation
        {
            get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f, 0f }); }
            set { Attributes["stdDeviation"] = value; }
        }
```

https://github.com/vvvv/SVG/blob/4580a4e7874afb26c2d0b0b45a68e6b8ae65640c/Source/Filter%20Effects/feGaussianBlur/SvgGaussianBlur.cs#L221-L238

https://www.w3.org/TR/SVG11/filters.html#feGaussianBlurElement
```
stdDeviation = "<number-optional-number>"
The standard deviation for the blur operation. If two <number>s are provided, the first number represents a standard deviation value along the x-axis of the coordinate system established by attribute ‘primitiveUnits’ on the ‘filter’ element. The second value represents a standard deviation in Y. If one number is provided, then that value is used for both X and Y.
A negative value is an error (see Error processing). A value of zero disables the effect of the given filter primitive (i.e., the result is the filter input image). If ‘stdDeviation’ is 0 in only one of X or Y, then the effect is that the blur is only applied in the direction that has a non-zero value.
If the attribute is not specified, then the effect is as if a value of 0 were specified.
Animatable: yes.
```

H1Gdev:
@wieslawsoltes 

Please use `rebase` instead of `merge` because differences are not displayed correctly.

Tests/Svg.UnitTests/SvgNumberCollectionTests.cs
```diff
@@ -0,0 +1,19 @@
...
+        [Test]
+        public void ToStringReturnsValidString()
+        {
+            var collection = new SvgNumberCollection
+            {
+                1.6f, 3.2f,
+                1.2f, 5f
+            };
+            Assert.AreEqual("1.6 3.2 1.2 5", collection.ToString());
+        }
```
H1Gdev: Please add string to number collection.(with various delimiters.)

Source/Filter Effects/SvgFilterPrimitive.cs
```diff
@@ -15,6 +15,34 @@ public abstract class SvgFilterPrimitive : SvgElement
         public const string FillPaint = "FillPaint";
         public const string StrokePaint = "StrokePaint";
 
+        [SvgAttribute("x")]
+        public SvgUnit X
+        {
+            get { return GetAttribute<SvgUnit>("x", false, new SvgUnit(SvgUnitType.Percentage, 0f)); }
```
H1Gdev: Please omit type.

Source/DataTypes/SvgNumberCollection.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            var points = new SvgPointCollection();
+            foreach (var point in this)
+                points.Add(point);
+            return points;
+        }
+
+        public override string ToString()
+        {
+            var builder = new StringBuilder();
```
H1Gdev: more simple.

```cs
return string.Join(" ", this.Select(v => v.ToString(CultureInfo.InvariantCulture)).ToArray());
```

mrbean-bremen:
> I would also like to change type of the StdDeviation property of the SvgGaussianBlur filter effect as it's not following the SVG specification. 

I also had noticed this, and I think it is ok to change. Given that almost none of the filters is currently implemented correctly (Gaussian blur is there, but only partly fillfulls the spec, and also is very slow), I think this kind of changes will not be a problem.

Source/Filter Effects/feBlend/SvgBlend.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        public override void Process(ImageBuffer buffer)
+        {
+            // TODO: Implement feBlend filter Process().
+        }
+
+        public override SvgElement DeepCopy()
+        {
+            return DeepCopy<SvgGaussianBlur>();
```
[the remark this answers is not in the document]
mrbean-bremen: Good point!

Source/Filter Effects/feBlend/SvgBlendMode.cs
```diff
@@ -0,0 +1,14 @@
...
+    [TypeConverter(typeof(EnumBaseConverter<SvgBlendMode>))]
+    public enum SvgBlendMode
+    {
+        Normal,
+        Multiply,
+        Screen,
+        Darken,
+        Lighten
+    }
+}
```
mrbean-bremen: This is from SVG 1.1, not sure if we want to add the SVG 2 additions (I has in [my PR](https://github.com/vvvv/SVG/pull/497/files#diff-6d83bf45faed77dbf6526d6089f96f76)).

mrbean-bremen:
As for the commit comments: I usually use squash merges for PRs with more than one commit (except than the commits are not related), and the commit comments end up as separate lines in the comment. I usually try to sanitize these, but it is good to have already meaningful comments.
Merge policy is one point to agree upon after the move to the new repo...

Source/Filter Effects/feBlend/SvgBlend.cs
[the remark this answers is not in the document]
wieslawsoltes: > Not `SvgGaussianBlur`.
> 
> To detect this mistake, add it to `ElementTypeList` of `SvgElementDeepCopyTest`.
> ("Filter Effects" starts around line 39.)

Fixed.

Source/Filter Effects/SvgFilterPrimitive.cs
wieslawsoltes: Foxed.

Source/DataTypes/SvgNumberCollection.cs
wieslawsoltes: Fixed.

Source/Filter Effects/feBlend/SvgBlendMode.cs
wieslawsoltes: > This is from SVG 1.1, not sure if we want to add the SVG 2 additions (I has in [my PR](https://github.com/vvvv/SVG/pull/497/files#diff-6d83bf45faed77dbf6526d6089f96f76)).

I can support those is my Svg.Skia to I will add them in PR, may be useful in feature for others.

Tests/Svg.UnitTests/SvgNumberCollectionTests.cs
wieslawsoltes: Added.

wieslawsoltes:
> > I would also like to change type of the StdDeviation property of the SvgGaussianBlur filter effect as it's not following the SVG specification.
> 
> I also had noticed this, and I think it is ok to change. Given that almost none of the filters is currently implemented correctly (Gaussian blur is there, but only partly fillfulls the spec, and also is very slow), I think this kind of changes will not be a problem.

I did try to retrofit SvgGaussianBlur filter effect with SvgNumberCollection as StdDeviation  property type but failed. 

My proposal is to remove current implementation of SvgGaussianBlur Process(and related code) and just fix property  type.

wieslawsoltes:
@mrbean-bremen @H1Gdev I have fixed all requested changes and I am done with all work I wanted to do. This is ready for final review and merge (preferable squashing merge?).

wieslawsoltes:
For reference I have partially implement all the filter effects added in this PR in my Svg.Skia library.

https://github.com/wieslawsoltes/Svg.Skia/blob/master/src/Svg.Skia/SvgFilterskExtensions.cs

Source/Filter Effects/feSpecularLighting/SvgSpecularLighting.cs
```diff
@@ -0,0 +1,74 @@
...
+            set { Attributes["lighting-color"] = value; }
+        }
+
+        public SvgElement LightSource
+        {
+            get
+            {
+                foreach (var child in this.Children)
+                {
+                    switch (child)
```
H1Gdev: cast is not needed.

```cs
if (child is SvgDistantLight || child is SvgPointLight || child is SvgSpotLight)
    return child;
```

Source/DataTypes/SvgNumberCollection.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <returns>
+        /// An <see cref="T:System.Object"/> that represents the converted value.
+        /// </returns>
+        /// <exception cref="T:System.NotSupportedException">The conversion cannot be performed. </exception>
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
+        {
+            if (value is string str)
+            {
+                var collection = new SvgNumberCollection();
+                if (string.IsNullOrEmpty(str))
```
H1Gdev: `str` is not null, so remove this block.

Source/Filter Effects/SvgFilter.cs
```diff
@@ -12,13 +12,27 @@ namespace Svg.FilterEffects
...
+        }
+
         /// <summary>
         /// Gets or sets the position where the left point of the filter.
         /// </summary>
         [SvgAttribute("x")]
         public SvgUnit X
         {
-            get { return GetAttribute<SvgUnit>("x", false); }
+            get { return GetAttribute<SvgUnit>("x", false, new SvgUnit(SvgUnitType.Percentage, -10f)); }
```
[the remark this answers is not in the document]
wieslawsoltes: Fixed.

Source/DataTypes/SvgNumberCollection.cs
wieslawsoltes: Fixed.

Source/Filter Effects/feSpecularLighting/SvgSpecularLighting.cs
wieslawsoltes: Fixed.

possibleshowdown:
Hello!  Thank you for all of your work on this.  To clarify, did this actually implement all of the filters listed, or are they not quite finished?  I am using this library to render PNGs from SVGs and have been having trouble with the filters working, particularly feBlend, feComposite, and feColorMatrix.  Thanks again!

wieslawsoltes:
> To clarify, did this actually implement all of the filters listed, or are they not quite finished?

I did not implement any filters effects.

possibleshowdown:
Ah OK, thank you for confirming!

wieslawsoltes:
@possibleshowdown 
I am working on alternative rendering backend using SkiaSharp. It does support some of the filters. https://github.com/wieslawsoltes/Svg.Skia

PR #640, opened 2020-01-13
CoordinateParser should handle NumState.invalid first

wieslawsoltes:
#### Reference Issue

**e-polyline-004.svg**
```SVG
<svg id="svg1" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <title>Stop processing on invalid data</title>
    <desc>`text` and everything after it should be ignored</desc>

    <!-- should be covered -->
    <path id="path1" d="M 20 40 L 160 180 L 30 150" fill="none" stroke="red"/>

    <polyline id="polyline1" points="20 40 160 180 30 150 text" fill="none" stroke="green"/>

    <!-- image frame -->
    <rect id="frame" x="1" y="1" width="198" height="198" fill="none" stroke="black"/>
</svg>
```

**e-polygon-004.svg**
```SVG
<svg id="svg1" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <title>Stop processing on invalid data</title>
    <desc>`text` and everything after it should be ignored</desc>

    <!-- should be covered -->
    <path id="path1" d="M 20 40 L 160 180 L 30 150" fill="red" stroke="red"/>

    <polygon id="polygon1" points="20 40 160 180 30 150 text" fill="green" stroke="green"/>

    <!-- image frame -->
    <rect id="frame" x="1" y="1" width="198" height="198" fill="none" stroke="black"/>
</svg>
```

PR #642, opened 2020-01-19
Fix CoordinateParser

wieslawsoltes:

Fixes broken CoordinateParser from #640

The #640 broke part of `paths-data-18-f.svg` test:
```SVG
<svg id="svg-root" width="100%" height="100%" viewBox="0 0 480 360" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <g id="test-body-content">
    <g fill="gold">
      <rect x="220" y="280" width="160" height="8"/>
    </g>
    <g transform="translate(60,-30) scale(8,2)">
      <path d="M 20 160 H 40" stroke-width="2" stroke="red" />
      <path d="M 20 160 H 40#90" stroke-width="4" stroke="black" />
    </g>
  </g>
</svg>

```


Source/Paths/CoordinateParser.cs
```diff
@@ -274,7 +274,7 @@ public bool TryGetFloat(out float result)
                         break;
                 }
 
-                if (_newState == NumState.invalid)
+                if (_currState == NumState.separator && _newState == NumState.invalid)
```
H1Gdev: I think this block should be deleted.

And you add the following condition in line 283:
```cs
if (_currState != NumState.separator && _newState < _currState)
```

wieslawsoltes: > I think this block should be deleted.
> 
> And you add the following condition in line 283:
> 
> ```cs
> if (_currState != NumState.separator && _newState < _currState)
> ```

@H1Gdev Yes this works. Updated the PR

PR #643, opened 2020-01-20
Set shape-rendering property default value to auto

wieslawsoltes:

Sets `shape-rendering` property default value to `auto`.

According to spec default/initial value is `auto`:
https://www.w3.org/TR/SVG11/painting.html#ShapeRenderingProperty
```
Value:  	auto | optimizeSpeed | crispEdges |
geometricPrecision | inherit
Initial:  	auto
Applies to:  	shapes
Inherited:  	yes
Percentages:  	N/A
Media:  	visual
Animatable:  	yes
```
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering
```
Value	auto | optimizeSpeed | crispEdges | geometricPrecision
Default value	auto
Animatable	Yes
```

### DIFF
--- a/Source/SvgElementStyle.cs
+++ b/Source/SvgElementStyle.cs
@@ -154,7 +154,7 @@ namespace Svg
         [SvgAttribute("shape-rendering")]
         public virtual SvgShapeRendering ShapeRendering
         {
-            get { return GetAttribute<SvgShapeRendering>("shape-rendering", true); }
+            get { return GetAttribute<SvgShapeRendering>("shape-rendering", true, SvgShapeRendering.Auto); }
             set { Attributes["shape-rendering"] = value; }
         }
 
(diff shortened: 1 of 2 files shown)


Source/SvgElementStyle.cs
```diff
@@ -154,7 +154,7 @@ public virtual float Opacity
         [SvgAttribute("shape-rendering")]
         public virtual SvgShapeRendering ShapeRendering
         {
-            get { return GetAttribute<SvgShapeRendering>("shape-rendering", true); }
+            get { return GetAttribute<SvgShapeRendering>("shape-rendering", true, SvgShapeRendering.Auto); }
```
H1Gdev: can omit type `<SvgShapeRendering>` in method.

wieslawsoltes: The type was in original, I can remove it if that's requested.

H1Gdev: If `defaultValue` is specified, type is omitted for other properties.

wieslawsoltes: I will remove the unnecessary type.

wieslawsoltes: Pushed the change.

PR #648, opened 2020-01-26
Fix StdDeviation property type of the SvgGaussianBlur class

### LINKED ISSUE #645: Invalid StdDeviation property type of the SvgGaussianBlur class
Note: Adding separate issue based on work done with https://github.com/vvvv/SVG/pull/641 so it can be tracked in the feature and not lost.

#### Description

I would also like to change type of the `StdDeviation` property of the `SvgGaussianBlur` filter effect as it's not following the SVG specification. This is off-course breaking change. I hope at least we can discuss this here as part of this PR ads `SvgNumberCollection` data type and uses it extensively. 

```C#
        [SvgAttribute("stdDeviation")]
        public SvgNumberCollection StdDeviation
        {
            get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f, 0f }); }
            set { Attributes["stdDeviation"] = value; }
        }
```

https://github.com/vvvv/SVG/blob/4580a4e7874afb26c2d0b0b45a68e6b8ae65640c/Source/Filter%20Effects/feGaussianBlur/SvgGaussianBlur.cs#L221-L238

https://www.w3.org/TR/SVG11/filters.html#feGaussianBlurElement
```
stdDeviation = "<number-optional-number>"
The standard deviation for the blur operation. If two <number>s are provided, the first number represents a standard deviation value along the x-axis of the coordinate system established by attribute ‘primitiveUnits’ on the ‘filter’ element. The second value represents a standard deviation in Y. If one number is provided, then that value is used for both X and Y.
A negative value is an error (see Error processing). A value of zero disables the effect of the given filter primitive (i.e., the result is the filter input image). If ‘stdDeviation’ is 0 in only one of X or Y, then the effect is that the blur is only applied in the direction that has a non-zero value.
If the attribute is not specified, then the effect is as if a value of 0 were specified.
Animatable: yes.
```

wieslawsoltes:
#### Reference Issue

Fixes #645


Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -222,23 +241,12 @@ public Bitmap Apply(Image inputImage)
...
-            {
-                if (value <= 0f)
-                {
-                    throw new InvalidOperationException("Radius must be greater then 0");
-                }
-                _stdDeviation = value;
-                Attributes["stdDeviation"] = value;
-                PreCalculate();
-            }
+            get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f, 0f }); }
```
H1Gdev: https://www.w3.org/TR/SVG11/filters.html#feGaussianBlurStdDeviationAttribute
`If the attribute is not specified, then the effect is as if a value of 0 were specified.`

So, 
```cs
get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f }); }
```

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -13,55 +13,74 @@ public enum BlurType
...
         private int[,] _multable;
 
-        public SvgGaussianBlur()
-            : this(1f, BlurType.Both)
+        private void PreCalculate()
         {
-        }
+            float stdDeviationX = 0f;
+            float stdDeviationY = 0f;
+            if (StdDeviation == null)
```
H1Gdev: not need this block.

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -13,55 +13,74 @@ public enum BlurType
...
+            }
 
-        public SvgGaussianBlur(float stdDeviation, BlurType blurType)
-            : base()
-        {
-            _stdDeviation = stdDeviation;
-            BlurType = blurType;
-            PreCalculate();
-        }
+            _stdDeviationX = stdDeviationX;
```
H1Gdev: This should be set when Precalc(_isPrecalculated set true) succeeds.

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -13,55 +13,74 @@ public enum BlurType
...
+                stdDeviationY = StdDeviation[1];
+            }
 
-        public SvgGaussianBlur(float stdDeviation)
-            : this(stdDeviation, BlurType.Both)
-        {
-        }
+            if (_stdDeviationX == stdDeviationX && _stdDeviationY == stdDeviationY)
+            {
+                _isPrecalculated = true;
```
H1Gdev: Is this unnecessary ?

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -13,55 +13,74 @@ public enum BlurType
...
             }
+
+            _isPrecalculated = true;
         }
 
         public Bitmap Apply(Image inputImage)
         {
+            PreCalculate();
+
+            if (_isPrecalculated == false)
```
H1Gdev:
```cs
if (!_isPrecalculated)
    return null;
```

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
wieslawsoltes: If only one value is set than second is set to first value. It is expected to have two values.

H1Gdev: Third argument of GetAttribute method is default value if not specified.
So, please match with specification.

The following are recognized as equivalent:

```cs
get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f }); }
```

```cs
get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f, 0f }); }
```

wieslawsoltes: Pushed changes to PR

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
wieslawsoltes: Pushed changes to PR

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
wieslawsoltes: Pushed changes to PR

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -13,55 +13,64 @@ public enum BlurType
...
+            }
 
-        public SvgGaussianBlur(float stdDeviation, BlurType blurType)
-            : base()
-        {
-            _stdDeviation = stdDeviation;
-            BlurType = blurType;
-            PreCalculate();
-        }
+            if (_isPrecalculated && _stdDeviationX == stdDeviationX && _stdDeviationY == stdDeviationY)
```
H1Gdev: Is it unnecessary to check _isPrecalculated ?
This method executes precalculate when XY is changed, regardless of _isPrecalculated, right ?

wieslawsoltes: This method only needs to be executed only when XY changes.

H1Gdev: So, change.

```cs
if (_stdDeviationX == stdDeviationX && _stdDeviationY == stdDeviationY)
```

wieslawsoltes: Pushed the change.

Source/Filter Effects/feGaussianBlur/SvgGaussianBlur.cs
```diff
@@ -222,23 +231,12 @@ public Bitmap Apply(Image inputImage)
...
-                if (value <= 0f)
-                {
-                    throw new InvalidOperationException("Radius must be greater then 0");
-                }
-                _stdDeviation = value;
-                Attributes["stdDeviation"] = value;
-                PreCalculate();
-            }
+            get { return GetAttribute("stdDeviation", false, new SvgNumberCollection() { 0f }); }
+            set { Attributes["stdDeviation"] = value; }
```
H1Gdev: Negative number check has been removed.
What happens if negative number is set ?

FYI:
Firefox: Do not display element.
Chrome: No effect of filter effect.

wieslawsoltes: It shouldn't throw in setter. It should be handled in filter processing part.

H1Gdev: What kind of image is output when negative number is set?

wieslawsoltes: It's error but I don't think throwing in setter is the right place to handle such errors. 

There are various ways proposed here:
https://www.w3.org/TR/SVG11/implnote.html#ErrorProcessing

H1Gdev: There is no problem with deletion.

H1Gdev: Behavior is as follows.

### Test SVG

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="120" height="120">
  <defs>
    <filter id="f1" x="0" y="0">
      <feGaussianBlur in="SourceGraphic" stdDeviation="-1" />
    </filter>
  </defs>
  <rect x="10" y="10" width="90" height="90" stroke="green" stroke-width="2" fill="yellow" filter="url(#f1)" />
</svg>
```

### Before PR

before

### After PR

output

become like firefox.

wieslawsoltes: https://github.com/vvvv/SVG/pull/659

PR #649, opened 2020-01-27
Remove binary build artifacts

wieslawsoltes:
### What does this implement/fix? Explain your changes.

- Remove binary build artifacts `bin` and `obj` folders and contents from `Samples\RuntimeUpdates\` path.

also I have noticed some other issues related to samples/solution:
- Fixed `Entities` and `RuntimeUpdates` samples.
- Removed unused images from Libraries folder.
- Added `XMLOutputTester` sample to solution.
- Moved SVGViewer to Samples solution folder.
- Removed unnecessary `Samples\.gitignore`.
- Added solution `Tests` folder for test projects.


H1Gdev:
@wieslawsoltes 
@mrbean-bremen 

Please tell us what the following applications can be used for.

- Entities
- SvgSample
- XMLOutputTester

H1Gdev:
>     * XMLOutputTester

I understood this application.

wieslawsoltes:
> @wieslawsoltes
> @mrbean-bremen
> 
> Please tell us what the following applications can be used for.
> 
> * Entities
> * SvgSample
> * XMLOutputTester

This are existing sample apps, I have just added them to solution and fixed build. If they are not helpful I propose to remove them from repo.

mrbean-bremen:
These are examples applications, and I agree that they are more visible if added to the solution. I haven't checked the usability of the examples per se - I agree with @wieslawsoltes that they shall be removed if they are obsolete.

H1Gdev:
@wieslawsoltes 
@mrbean-bremen 

> * Entities
> * SvgSample

These are sample codes.
Running it will raise an exception.
Because it is a CLI, you may need to specify arguments.

> * XMLOutputTester

This could be used for code-based testing if modified.

I think that it is better to write description of projects in readme or wiki, if add them to solution.

mrbean-bremen:
A description of the samples on a separate documentation page is a good idea! Maybe I will have a look later...

mrbean-bremen:
I checked the samples. Both are command line tools, but don't need a special command line (an example image is provided in both cases).
`SvgSample` reads an SVG file, changes the style of one of it's elements, and saves it to a png. This works for me.
`Entities` is supposed to read a file with some unknown entities, providing these entities as a dictionary. This fails, which seems to be a bug - the functionality is probably broken - I'll add an issue for this.
Concerning the documentation for the samples - I may get at this some time later (currently not up to it), or someone else could have a go.

PR #652, opened 2020-01-29
Fix: #651 'Providing entities in SvgDocument.Open does not work' issue.

### LINKED ISSUE #651: Providing entities in SvgDocument.Open does not work
This came up in PR #649.

#### Description
The example `Entities` is supposed to provide custom entities while reading a file where these entities are used, but not defined. The example raises an `System.Xml.XmlException` exception due to the undeclared entity.
Given that the example must have worked some time, I guess this is a regression that has been introduced at some time.

#### Used Versions
Current master.

The example can easily be converted to (or used as base for) a unit test if this issue is fixed.

H1Gdev:
#### Reference Issue

Fixes #651.
Ref. #649 and #90.


H1Gdev:
@mrbean-bremen 

> may just use the example itself for that

Is it possible to run Entities in regression test ?

mrbean-bremen:
> Is it possible to run Entities in regression test ?

I meant not directly, but just read the file from the sample and use the same open call (without saving it back). You can add a test to check the elements that use the elements to have the correct style.

PR #658, opened 2020-02-01
Fix: #252 'Won't display gradients if they're wider than 698 px' issue.

### LINKED ISSUE #252: Won't display gradients if they're wider than 698 px
The SVG engine does seem to not render anything if the attached gradient rectangle width is **699 pixels** or wider. If I make it **698 pixels** or less, or make it a solid color, the rectangle renders correctly.
[wide.zip](https://github.com/vvvv/SVG/files/483022/wide.zip)

H1Gdev:
#### Reference Issue

Fixes #252.

#### Any other comments?

Ref. #93


Source/Painting/SvgLinearGradientServer.cs
```diff
@@ -78,44 +69,37 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
 
                 using (var transform = EffectiveGradientTransform)
                 {
-                    var midPoint = new PointF((points[0].X + points[1].X) / 2, (points[0].Y + points[1].Y) / 2);
                     transform.Translate(bounds.X, bounds.Y, MatrixOrder.Prepend);
                     if (this.GradientUnits == SvgCoordinateUnits.ObjectBoundingBox)
                     {
                         // Transform a normal (i.e. perpendicular line) according to the transform
                         transform.Scale(bounds.Width, bounds.Height, MatrixOrder.Prepend);
-                        transform.RotateAt(-90.0f, midPoint, MatrixOrder.Prepend);
```
mrbean-bremen: I'm not sure what that rotation was supposed to do - can you explain that? Obviously it's not needed...

Source/Painting/SvgLinearGradientServer.cs
```diff
@@ -78,44 +69,37 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
...
-                        var startX = (float)((dy * dx * (midPoint.Y - y2) + Math.Pow(dx, 2) * midPoint.X + Math.Pow(dy, 2) * x2) /
+                        var startX = (float)((dy * dx * (midPoint.Y - y2) + Math.Pow(dx, 2) * midPoint.X + Math.Pow(dy, 2) * x1) /
                         (Math.Pow(dx, 2) + Math.Pow(dy, 2)));
-                        var startY = dy * (startX - x2) / dx + y2;
-                        points[0] = new PointF(startX, startY);
-                        points[1] = new PointF(midPoint.X + (midPoint.X - startX), midPoint.Y + (midPoint.Y - startY));
+                        var endY = dy * (startX - x1) / dx + y2;
+                        points[0] = new PointF(startX, midPoint.Y + (midPoint.Y - endY));
+                        points[1] = new PointF(midPoint.X + (midPoint.X - startX), endY);
                     }
```
mrbean-bremen: This looks correct, as far as I can see, but I don't understand why the original problem (gradients > 698px not rendered) happend. Can you explain this a bit?

wieslawsoltes:
> Looks good as far as I can see, maybe @wieslawsoltes can have another look - you guys reviewing each other seems to work nicely :)

I won't be able to review in the following week as I am on one week vacation.

mrbean-bremen:
> I won't be able to review in the following week as I am on one week vacation.

Ok, in that case have a nice vacation!

Source/Painting/SvgLinearGradientServer.cs
H1Gdev: Although it is a gradation in horizontal direction, `dx` does not become zero due to error due to rotation calculation.

`dx = -0.0000610351563`

As a result, diagonal line is calculated.

In case of horizontal (or vertical), starting and ending point do not need to be converted, so processing in that case is also deleted.

Source/Painting/SvgLinearGradientServer.cs
```diff
@@ -78,44 +69,37 @@ public override Brush GetBrush(SvgVisualElement renderingElement, ISvgRenderer r
...
                     if (this.GradientUnits == SvgCoordinateUnits.ObjectBoundingBox)
                     {
                         // Transform a normal (i.e. perpendicular line) according to the transform
                         transform.Scale(bounds.Width, bounds.Height, MatrixOrder.Prepend);
-                        transform.RotateAt(-90.0f, midPoint, MatrixOrder.Prepend);
                     }
                     transform.TransformPoints(points);
                 }
 
+                points[0].X = (float)Math.Round(points[0].X, 4);
```
H1Gdev: Unnecessary rotations have been deleted, but there are user-specified rotations, so round off to eliminate errors.

PR #661, opened 2020-02-05
Fixed SvgFontStyle values.

H1Gdev:
#### Reference Issue

Ref. #147.

##### font-style

https://www.w3.org/TR/SVG11/text.html#FontStyleProperty

-  ~~Remove `All`.~~
- Add `Inherit`.

#### Any other comments?

`font-face` element has `font-style`, too.

https://www.w3.org/TR/SVG11/fonts.html#FontFaceElementFontStyleAttribute

~~If we implement the font-face, it is necessary to devise such as divide namespace.~~


H1Gdev:
I revert `All` because I think enum can be used in common(`text` and `font-face`).

PR #663, opened 2020-02-06
Add tests for EnumConverters.

H1Gdev:

Add tests for EnumConverters and fixed errors.

### DIFF
--- a/Source/DataTypes/SvgFontWeight.cs
+++ b/Source/DataTypes/SvgFontWeight.cs
@@ -12,45 +12,45 @@ namespace Svg
         Inherit,
 
         /// <summary>Same as <see cref="W400"/>.</summary>
-        Normal = W400,
+        Normal = 1,
 
         /// <summary>Same as <see cref="W700"/>.</summary>
-        Bold = W700,
+        Bold = 2,
 
-        /// <summary>One font weight darker than the parent element.</summary>
-        Bolder = 512,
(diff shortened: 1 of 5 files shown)


Source/DataTypes/SvgFontWeight.cs
```diff
@@ -12,45 +12,45 @@ public enum SvgFontWeight
         Inherit,
 
         /// <summary>Same as <see cref="W400"/>.</summary>
-        Normal = W400,
+        Normal = 1,
 
         /// <summary>Same as <see cref="W700"/>.</summary>
-        Bold = W700,
+        Bold = 2,
```
mrbean-bremen: Doesn't that change the semantics? In [CSS](https://www.w3.org/TR/CSS2/fonts.html#font-boldness), Normal is defined as W400, and Bold as W700, as has been before. Where do you account for it now?

H1Gdev: @mrbean-bremen 

When writing SVG, "normal" is output as "400", so enum values are separated.

Please check 01666aae54a3c5a619d57e1fc29d99b55d3a86ca in this PR.
Add `Bold` in SvgElementStyle.cs.
(`Normal` and `W400` are not used.)

mrbean-bremen: Ok, thanks - I thought that this equivalency is used, but it seems not to - so this should be ok than.

PR #666, opened 2020-02-08
Refactor in EnumConverters

H1Gdev:

- Move EnumConverters.cs from "Painting" to "DataTypes".
- Support kebab-case.


Source/DataTypes/EnumConverters.cs
```diff
@@ -168,14 +168,6 @@ public sealed class SvgImageRenderingConverter : EnumBaseConverter<SvgImageRende
     public sealed class SvgFontVariantConverter : EnumBaseConverter<SvgFontVariant>
     {
         public SvgFontVariantConverter() : base(SvgFontVariant.Normal, CaseHandling.KebabCase) { }
-
-        public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
```
H1Gdev: By changing from `Smallcaps` to `SmallCaps` in `SvgFontVariant`, override of this method becomes unnecessary.

This enum value is already public.
However, FontVariant is not yet supported.

So, I decide that it is safe to change.

What do you think ?

mrbean-bremen: I think this is ok. @wieslawsoltes - can you have a look, please?

Issue #590, opened 2019-10-15
Consider splitting rendering and processing


gvheertum:
In the light of the issue #576 (where the author indicated a .NET standard build might be helpful) we might consider splitting some of the parts from the SVG library. As far as I known part of the rendering (System.Drawing) is the main culprit for the library not being able to be built against .NET standard. From my understanding the SVG library can facilitate 2 things: processing and transforming SVG files (the XML part) and rendering an SVG to a bitmap.

The first part could classify for a .NET standard build. Downside however is that this might result in 2 libraries/packages, but it might help people that want to use the SVG part in .NET standard applications.

Just a random thought, but perhaps something to consider in the future.

wieslawsoltes:
This would help a lot with my SkiaSharp port.

This is my current rendering interface used for SkiaSharp:
https://github.com/wieslawsoltes/Svg.Skia/blob/4ac1dfb7d58cf9121a472a1d086367b39ba0cf01/src/Svg.Skia/ISvgRenderer.cs#L9
https://github.com/wieslawsoltes/Svg.Skia/blob/master/src/Svg.Skia/SvgRenderer.cs#L11
and a lot of helper methods:
https://github.com/wieslawsoltes/Svg.Skia/blob/master/src/Svg.Skia/SvgHelper.cs#L16

mrbean-bremen:
So, I understand you are volunteering for this? 😛

wieslawsoltes:
> So, I understand you are volunteering for this? 😛

I will not be easy to decouple the System.Drawing dependencies from current implementation. 

Anyway I will have pretty good idea and some implementation (for Skia) for further discussion.

mrbean-bremen:
Sounds good! 👍

tebjan:
See the linked issue, we could evaluate whether we could join forces with the skia drawing repo.

paolofulgoni:
As I wrote on Gitter yesterday, I was thinking about extracting the SVG manipulation part in a separate library too, and make it fully netstandard compliant.
My original idea was a completely separate lib, not a split of the this one. But the latter may actually make more sense. I'm just a bit afraid it would be harder and more constrained.

paolofulgoni:
@wieslawsoltes  do you think you could achieve the full rendering functionality of vvvv/Svg with your Svg.Skia implementation?

wieslawsoltes:
[...]

PR #669, opened 2020-02-10
Enable checking if PatternUnits or PatternContentUnits was actually set

wieslawsoltes:
#### Reference Issue

The https://github.com/vvvv/SVG/pull/668 actually broke Svg.Skia pattern rendering, did not see this as breaking at first. In order to correctly render pattern, pattern property inheritance must be used to check for inherited values. As #668 removed invalid `inherited` enum value it broke Svg.Skia as it does not have access to internal fields and those are necessary to provide information if actually PatternUnits or PatternContentUnits was set.

Adds HasPatternUnits() and HasPatternContentUnits() methods to check if field has actually a value.

### DIFF
--- a/Source/Painting/SvgPatternServer.cs
+++ b/Source/Painting/SvgPatternServer.cs
@@ -141,6 +141,24 @@ namespace Svg
         }
 
         /// <summary>
+        /// Check if <see cref="PatternUnits"/> property value was set.
+        /// </summary>
+        /// <returns>True if <see cref="PatternUnits"/> has value.</returns>
+        public bool HasPatternUnits()
+        {
+            return _patternUnits.HasValue;
+        }
+
+        /// <summary>
+        /// Check if <see cref="PatternContentUnits"/> property value was set.
+        /// </summary>
+        /// <returns>True if <see cref="PatternContentUnits"/> has value.</returns>
+        public bool HasPatternContentUnits()
+        {
+            return _patternContentUnits.HasValue;
+        }
+
+        /// <summary>
         /// Gets a <see cref="Brush"/> representing the current paint server.
         /// </summary>
         /// <param name="renderingElement">The owner <see cref="SvgVisualElement"/>.</param>


wieslawsoltes:
This is my code in Svg.Skia used for pattern rendering:

https://github.com/wieslawsoltes/Svg.Skia/blob/1298cbf0cd386f8694d97f7911432909fefbfa01/src/Svg.Skia/SvgPaintingExtensions.cs#L570-L595

wieslawsoltes:
Hiding such internals neccessary for proper rendering will make #590 much harder to achieve.

H1Gdev:
I think this PR solution is ad hocery.
`Inherit` was returned...

https://github.com/vvvv/SVG/blob/3efa867fb69ee2c68a59f5c1e8b92c988e94e1c7/Source/Painting/SvgPatternServer.cs#L155-L183

I think we should think of a way to return linked value in properties.

wieslawsoltes:
> I think we should think of a way to return linked value in properties.

Agree with that, returning correct linked value in properties is important and necessary in other place too.

H1Gdev:
@wieslawsoltes 
@mrbean-bremen 

Temporary methods that have been added in this PR do not know when to be deleted.
So I think it should be stated in comments etc.

wieslawsoltes:
> @wieslawsoltes
> @mrbean-bremen
> 
> Temporary methods that have been added in this PR do not know when to be deleted.
> So I think it should be stated in comments etc.

I think we can revert this changes.

I am using different method and not relying on internals. 

https://github.com/wieslawsoltes/Svg.Skia/blob/fa2cdcfbab6280e359657f9b8d930b4867ccae29/src/Svg.Skia/SvgPaintingExtensions.cs#L771
https://github.com/wieslawsoltes/Svg.Skia/blob/fa2cdcfbab6280e359657f9b8d930b4867ccae29/src/Svg.Skia/SvgExtensions.cs#L277-L284

This comes from from my work on using href properties for linear/radial gradients in Svg.Skia

https://github.com/wieslawsoltes/Svg.Skia/blob/fa2cdcfbab6280e359657f9b8d930b4867ccae29/src/Svg.Skia/SvgPaintingExtensions.cs#L316-L347
https://github.com/wieslawsoltes/Svg.Skia/blob/fa2cdcfbab6280e359657f9b8d930b4867ccae29/src/Svg.Skia/SvgPaintingExtensions.cs#L501-L540

PR #671, opened 2020-02-11
Added enum converters for Filter Effects

wieslawsoltes:

When EnumBaseConverter is using directly it fails as its abstract class and does not have default empty constructor:
https://github.com/vvvv/SVG/search?q=EnumBaseConverter&unscoped_q=EnumBaseConverter

### DIFF
--- a/Source/DataTypes/EnumConverters.cs
+++ b/Source/DataTypes/EnumConverters.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.ComponentModel;
 using System.Globalization;
 using System.Text.RegularExpressions;
--- a/Source/DataTypes/EnumConverters.cs
+++ b/Source/DataTypes/EnumConverters.cs
@@ -19,7 +19,7 @@ namespace Svg
     }
 
     // converts enums to lower case strings
-    public abstract class EnumBaseConverter<T> : BaseConverter
+    public class EnumBaseConverter<T> : BaseConverter
         where T : struct
     {
         public enum CaseHandling
@@ -33,8 +33,13 @@ namespace Svg
         public CaseHandling CaseHandlingMode { get; }
 
         /// <summary>Creates a new instance.</summary>
+        public EnumBaseConverter() : this(CaseHandling.CamelCase)
+        {
+        }
+
+        /// <summary>Creates a new instance.</summary>
         /// <param name="caseHandling">Defines if the value shall be converted to camelCase, lowercase or kebab-case.</param>
-        public EnumBaseConverter(CaseHandling caseHandling = CaseHandling.CamelCase)
+        public EnumBaseConverter(CaseHandling caseHandling)
         {
             CaseHandlingMode = caseHandling;
         }
--- a/Source/DataTypes/EnumConverters.cs
+++ b/Source/DataTypes/EnumConverters.cs
@@ -62,7 +62,6 @@ namespace Svg
             if (destinationType == typeof(string) && value is T)
             {
                 var stringValue = ((T)value).ToString();
-
                 if (CaseHandlingMode == CaseHandling.CamelCase)
                     return string.Format("{0}{1}", stringValue[0].ToString().ToLower(), stringValue.Substring(1));
 
--- a/Source/DataTypes/EnumConverters.cs
+++ b/Source/DataTypes/EnumConverters.cs
@@ -25,8 +25,8 @@ namespace Svg
     {
         public enum CaseHandling
         {
-            NoneCase,
             CamelCase,
+            PascalCase,
             LowerCase,
             KebabCase,
         }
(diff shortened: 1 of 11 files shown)


H1Gdev:
@wieslawsoltes 

> https://github.com/vvvv/SVG/search?q=EnumBaseConverter&unscoped_q=EnumBaseConverter

In `Filter Effects`, you use base class directly. . .
(I wanted to add a test, too.)

This class seems to be designed to be inherited.
(`Base` is included in name, etc.)

@mrbean-bremen 

How much do we want to change ?

H1Gdev:
@wieslawsoltes 

`SvgBlendMode` is `kebab-case`.

https://developer.mozilla.org/en-US/docs/Web/CSS/blend-mode

wieslawsoltes:
I will  update PR with proper converters for each enum and leave the base class abstract.

mrbean-bremen:
> Do you continue to follow current enum converter design or not ?

I haven't been following closely lately due to other tasks, though I think you (@H1Gdev and @wieslawsoltes) can handle this just fine together. I'll probably invest more time into the lib some time later... I'm also waiting for @tebjan to do the transition of the library, and after that probably have a core dev group with both of you included.

wieslawsoltes:
@H1Gdev Updated the PR with the feedback notes. Ready for another review.

Source/DataTypes/EnumConverters.cs
```diff
@@ -32,9 +34,14 @@ public enum CaseHandling
         /// <summary>Defines if the enum literal shall be converted to camelCase, lowercase or kebab-case.</summary>
```
H1Gdev: Update doc.

Source/DataTypes/EnumConverters.cs
```diff
@@ -32,9 +34,14 @@ public enum CaseHandling
...
 
+        /// <summary>Creates a new instance.</summary>
+        public EnumBaseConverter() : this(CaseHandling.CamelCase)
+        {
+        }
+
         /// <summary>Creates a new instance.</summary>
         /// <param name="caseHandling">Defines if the value shall be converted to camelCase, lowercase or kebab-case.</param>
-        public EnumBaseConverter(CaseHandling caseHandling = CaseHandling.CamelCase)
+        public EnumBaseConverter(CaseHandling caseHandling)
```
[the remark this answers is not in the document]
wieslawsoltes: Fixed

Source/DataTypes/EnumConverters.cs
wieslawsoltes: Fixed

Source/DataTypes/EnumConverters.cs
```diff
@@ -24,6 +25,7 @@ public abstract class EnumBaseConverter<T> : BaseConverter
     {
         public enum CaseHandling
         {
+            NoneCase,
```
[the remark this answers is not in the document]
wieslawsoltes: Fixed

Source/DataTypes/EnumConverters.cs
```diff
@@ -33,7 +35,7 @@ public enum CaseHandling
         public CaseHandling CaseHandlingMode { get; }
```
H1Gdev: Please Add "PascalCase" here, too.

H1Gdev:
And please update title.

Source/DataTypes/EnumConverters.cs
wieslawsoltes: Fixed

wieslawsoltes:
> And please update title.

Changed

Source/DataTypes/EnumConverters.cs
```diff
@@ -1,8 +1,9 @@
-﻿﻿using System;
+﻿using System;
```
H1Gdev: Has the encoding of the source file changed ?
`editorconfig` not applied ?

wieslawsoltes: No Visual Studio adds automatically correct BOM headers for Utf8 text files.

H1Gdev: [EditorConfig](https://editorconfig.org/) is supported by many editors and IDEs. (Not just Visual Studio)😄

wieslawsoltes: > [EditorConfig](https://editorconfig.org/) is supported by many editors and IDEs. (Not just Visual Studio)😄

I know that 😄 it's just that VS adds by default BOM header to .cs files with Utf8 encoding.

wieslawsoltes: Restored the original header bytes.

wieslawsoltes: @H1Gdev We probably should have separate PR with clean-up of .cs BOM headers accoring to editor config:
```ini
[*.cs]
charset = utf-8-bom
```

H1Gdev: I don't think you need to do that.

If **we** follow `EditorConfig`, I think coding styles will be unified over time.
(If not all.)

wieslawsoltes:
I think this is ready to be merged.

PR #676, opened 2020-02-17
Sample console application

### LINKED ISSUE #614: Prebuilt minimal SVG to PNG converter
Hi! I'm trying to use this library as a simple SVG-to-PNG converter, but I'm not familiar with C#, so I would not be able to build it myself.
Is there are some prebuilt binaries or a build instruction for a complete beginner?

wieslawsoltes:
#### Reference Issue

Fixes #614

Adds simple command-line application to convert one or many input .svg files to .png images.

```
SvgConsole:
  Converts svg files to encoded png images.

Usage:
  SvgConsole [options]

Options:
  -f, --inputFiles <inputfiles>              The relative or absolute path to the input files
  -d, --inputDirectory <inputdirectory>      The relative or absolute path to the input directory
  -o, --outputDirectory <outputdirectory>    The relative or absolute path to the output directory
  --outputFiles <outputfiles>                The relative or absolute path to the output files
  --width <width>                            The output image width override
  --height <height>                          The output image height override
  --version                                  Show version information
  -?, -h, --help                             Show help and usage information
```

Example usage:

```
./SvgConsole -f ~/svg/Example.svg
```

```
./SvgConsole -f ~/svg/Example.svg --outputFiles ~/png/Example.png
```

```
./SvgConsole -f ~/svg/Example.svg -o ~/png
```

```
./SvgConsole -d ~/svg/
```

```
./SvgConsole -d ~/svg -o ~/png
```


RazrFalcon:
Could you add an option to render to a specific file and not only into a directory?

wieslawsoltes:
> Could you add an option to render to a specific file and not only into a directory?

Added.

H1Gdev:
Please add this sample app in doc/Samples.md.

H1Gdev:
I have a question.
Why multiple inputs only ?

Can this app be done as follows ?
`./SvgConsole -f a.svg b.svg c.svg --outputFile a.png b.png c.png`

wieslawsoltes:
> I have a question.
> Why multiple inputs only ?
> 
> Can this app be done as follows ?
> `./SvgConsole -f a.svg b.svg c.svg --outputFile a.png b.png c.png`

This can be done too ofcourse. I will add that feature in this PR.

wieslawsoltes:
> Please add this sample app in doc/Samples.md.

Done.

wieslawsoltes:
> I have a question.
> Why multiple inputs only ?
> 
> Can this app be done as follows ?
> `./SvgConsole -f a.svg b.svg c.svg --outputFile a.png b.png c.png`

Added.

H1Gdev:
I think it is better to match these, too.

`--inputDirectories` and `--outputDirectory`.

I think `--inputDirectory` and` --outputDirectory` are good for directory specification.

wieslawsoltes:
> I think it is better to match these, too.(issue is likely to occur with the same file name.)
> 
> `--inputDirectories` and `--outputDirectory`.
> 
> I think `--inputDirectory` and` --outputDirectory` are good for directory specification.

The concept was to have many input files and directories and one output directory. I agree that matching many input files with many output files makes sense, still matching directories is bit trickies, as output directory is used for input files and input directories, so you can do the following:
```
./SvgConsole -f a.svg b.svg c.svg -d ./some/path1/svg ./some/path2/svg --outputDirectory ./some/path/png
```

wieslawsoltes:
> I think it is better to match these, too.(issue is likely to occur with the same file name.)
> 
> `--inputDirectories` and `--outputDirectory`.
> 
> I think `--inputDirectory` and` --outputDirectory` are good for directory specification.

Changed to only allow one input directory.

H1Gdev:
I think that we should as much as possible output files is not overwritten.

> ```
> ./SvgConsole -f a.svg b.svg c.svg -d ./some/path1/svg ./some/path2/svg --outputDirectory ./some/path/png
> ```

In this case, if the following file exists, what is `./some/path/png/a.png` ?
- a.svg
- ./some/path1/svg/a.svg
- ./some/path2/svg/a.svg

wieslawsoltes:
> I think that we should as much as possible output files is not overwritten.
> 
> > ```
> > ./SvgConsole -f a.svg b.svg c.svg -d ./some/path1/svg ./some/path2/svg --outputDirectory ./some/path/png
> > ```
> 
> In this case, if the following file exists, what is `./some/path/png/a.png` ?
> 
> * a.svg
> * ./some/path1/svg/a.svg
> * ./some/path2/svg/a.svg

I think it's up to the user to provide proper input configuration or should we provide more complex logic in sample app? I have changed to allow only one input directory. We can block usage of both input files and input directory together.

Samples/SvgConsole/Program.cs
```diff
@@ -0,0 +1,205 @@
...
+                }
+            }
+
+            for (int i = 0; i < paths.Count; i++)
+            {
+                var inputPath = paths[i];
+                var outputFile = settings.OutputFiles != null ? settings.OutputFiles[i] : null;
+                try
+                {
+                    string outputPath = string.Empty;
```
H1Gdev: use `var`.

Samples/SvgConsole/Program.cs
```diff
@@ -0,0 +1,205 @@
...
+
+                    if (outputFile != null)
+                    {
+                        outputPath = outputFile.FullName;
+                    }
+                    else
+                    {
+                        var inputExtension = inputPath.Extension;
+                        outputPath = inputPath.FullName.Remove(inputPath.FullName.Length - inputExtension.Length) + ".png";
```
H1Gdev: better to use `Path.ChangeExtension`.

H1Gdev:
> I think it's up to the user to provide proper input configuration or should we provide more complex logic in sample app? I have changed to allow only one input directory.

Since it is a sample application, it is unclear how much should be allowed, but it is necessary to implement it without misunderstanding because it is a sample application.

Samples/SvgConsole/Program.cs
```diff
@@ -0,0 +1,205 @@
...
+            };
+
+            var optionHeight = new Option(new[] { "--height" }, "The output image height override")
+            {
+                Argument = new Argument<float?>(getDefaultValue: () => null)
+            };
+
+            var rootCommand = new RootCommand()
+            {
+                Description = "Converts a svg file to an encoded png image."
```
H1Gdev: `Converts svg files to encoded png images.`  :thumbsup:

Samples/SvgConsole/Program.cs
wieslawsoltes: Done.

Samples/SvgConsole/Program.cs
wieslawsoltes: Done.

Samples/SvgConsole/Program.cs
wieslawsoltes: Done.

PR #685, opened 2020-03-03
Svg Document PointsPerInch Xamarin (iOS, Android, UWP) Support

inforithmics:
Svg Platfrom Support

#### Reference Issue
https://github.com/wieslawsoltes/Svg.Skia/issues/16

Changes GetSystemDpi to a Property of Func<int> so that it can be changed on a Platform basis
If other Platforms need to be supported.

#### Any other comments?
This is to support UWP Release Mode (Native Compilation)
https://github.com/wieslawsoltes/Svg.Skia/issues/16


mrbean-bremen:
@gvheertum - can you please have a look?

H1Gdev:
@inforithmics 

> But it would be possible to Implement different Implementations for UWP Xamarin.Forms Linux Mac .....

How do you switch platforms ?

inforithmics:
First Possibility:

In a Xamarin Forms there is normally a Bootstrapper/ Initializer Class where you Setup / Initialize all Libraries. There you would Set SvgPlatformSupport.Instance Variable with a Custom Implementation.
That provides the SystemDPI.

```
  // Xamarin Essentials
    public class SvgXamarinEssentialsPlatformSupport : IPlatformSupport
    {
        public int GetSystemDpi()
        {
            return Xamarin.Essentials.DeviceDisplay.MainDisplayInfo.Density * 96.0;
        }
    }


    SvgPlatformSupport.Instance = new SvgXamarinEssentialsPlatformSupport();
```

A direct Dependency on Xamarin Essentials won't work because currently there is no Windows .Net Implementation, maybe in the Future other Platforms are supported.

Second Possibility:

The other pattern in Xamarin Forms Apps is used (for Example )in Xamarin.Essentials. 

https://github.com/xamarin/Essentials

Here The Library is MultiTargeted and for Every Target A Different Dll is Used.
So for example 

Could have an Implementation per Platform where this SvgPlatformSupport is set.

Third Possibility:
SvgDocument.PointsPerInch isn't readonly and can be written in the initialization Phase.
GetSystemDpi would have to be improved to catch exceptions for Operation. (Is a bit hacky).

Fourth Possibility:
Improve GetSystemDpi so that it detects that it is in UWP/Windows and works correctly.

Fifth Possibility:
MultiTarget the Svg library to ios,android,uwp  and handle the GetSystemDpi with preprocessor 
directives as it is already done with netcore.

#if WINDOWS_UWP
....
#if __ANDROID__
....
#if __IOS__
....

As  my pull request works it isn't complete for example it would need a documentation update and Unit Tests.

As a side note: 

I added a default Implementation of SvgPlatformSupport that is initially set, so that it behaves the same as before.

Source/Platforms/SvgPlatformSupport.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    /// <summary>
+    /// Centralized Class for Implementing Platform Specific Code
+    /// </summary>
+    public static class PlatformSupport
+    {
+        /// <summary>
+        /// Gets or sets Svg Platform Support, the Default Implementation works for Desktop .Net Frameworks
+        /// </summary>
+        public static ISvgPlatformSupport Instance { get; set; } = new SvgDefaultPlatformSupport();
```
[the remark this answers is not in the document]
gvheertum: You could also consider making an (abstract) factory for the creation of the instances.

gvheertum:
I like the way you are going, so kudo's for that!

I am pondering how to get this working for all platforms without breaking dependencies and such. The best way to do this might be through dependency injection, however, that might require users to do some DI-container setup before they can use the library. The library could scan for certain types in the loaded application scope and use that one and if no override is found fall back to the default (current) implementation. That means that for general implementations nothing changes and for specific implementations like Xamarin there might be an Xamarin.SVG library (or whatever) that contains our library and a specific implementation of the stuff that we want to have Dependency-Injected.

Hope that that story makes any sense to you :)

Source/SvgDocument.cs
```diff
@@ -31,7 +30,7 @@ public class SvgDocument : SvgFragment, ITypeDescriptorContext
         /// </remarks>
         public static bool SkipGdiPlusCapabilityCheck { get; set; }
 
-        public static readonly int PointsPerInch = GetSystemDpi();
+        public static readonly int PointsPerInch = PlatformSupport.Instance.GetSystemDpi();
```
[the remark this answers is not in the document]
inforithmics: Good point and a good Unit test Candidate

Source/Platforms/SvgPlatformSupport.cs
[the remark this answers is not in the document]
inforithmics: Good point, avoids the problem that someone could set it by accident to null

H1Gdev:
build error will occur in Visual Studio.

Source/SvgDocument.cs
```diff
@@ -41,8 +44,30 @@ public override SvgOverflow Overflow
             get { return GetAttribute("overflow", false, SvgOverflow.Visible); }
         }
 
-        private static int GetSystemDpi()
+        /// <summary> Get System Dpi Function </summary>
+        public static Func<int> GetSystemDpi
         {
+            get => getSystemDpi ?? (getSystemDpi = GetSystemDpiDefault);
```
H1Gdev: I think this property should be kept simple.

```cs
get => getSystemDpi;
```

inforithmics:
What is the Build Error (I built it with Visual Studio 2019 with Xamarin and UWP Tools installed)

Source/SvgDocument.cs
inforithmics: No Problem, I can make it simple

H1Gdev:
@inforithmics 

>  (I built it with Visual Studio 2019 with Xamarin and UWP Tools installed)

I have not installed this tools.
Is there a way to prevent build errors even if it is not installed as before?
(Or guide you to install tools.)

inforithmics:
I will change the behavior so that it works with Xamarin tools without needing the tools installed.

inforithmics:
Currently I didn't find a solution enable the targets when they are installed
As workaround the targetframeworks can be manually changed 
from
```
<TargetFrameworks>
     netstandard2.0;netcoreapp2.2;net452;Xamarin.iOS10;MonoAndroid10.0
</TargetFrameworks>
<TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">
    $(TargetFrameworks);uap10.0.16299
</TargetFrameworks>
```
to
```
<TargetFrameworks>
     netstandard2.0;netcoreapp2.2;net452
</TargetFrameworks>
```
Mobile Development and Developement for UWP must be installed for Xamarin Forms Development
image

inforithmics:
I fixed the Project so that it compiles in Visual Studio 2017

Source/SvgDocument.cs
```diff
@@ -14,6 +14,7 @@
 using System.Globalization;
 using System.Runtime.InteropServices;
 using Svg.Exceptions;
+using SizeF = System.Drawing.SizeF;
```
H1Gdev: Why is this needed ?

inforithmics: It isn't needed, I Will remove it

inforithmics:
I have found a solution to detect on which framework i'm running

https://stackoverflow.com/questions/48259320/how-to-determine-the-net-platform-that-runs-the-net-standard-class-library

I will try to use this approach, so no new targets are needed and I can remove the project changes

inforithmics:
I didn't suceed in my approach to detect the framework at runtime and use reflection to get the dpi so I use now the targeting capability of MSBuild.Sdk.Extras which is althoug used by Microsoft.

inforithmics:
I made a pull request so that it doesn't crash in UWP Release Mode anymore, 
with minimal changes. So this pull request can be retired

https://github.com/vvvv/SVG/pull/720

mrbean-bremen:
Close as requested.

PR #686, opened 2020-03-04
Fix: 'none' does not work at clip-path and filter.

H1Gdev:
#### Reference Issue

Fix issue that exception occurs when `none` is set.


H1Gdev:
Author...👻

Issue #693, opened 2020-03-20
Replace XmlTextWriter with the more general XmlWriter.


bbsgfalconer:
#### Description

I want to be able to serialize an SVG using a specialized XmlWriter that is not an XmlTextWriter. The specific use case is to write out the SVG as part of an Asp.Net Blazor application.

Could SvgDocument.Write(...) and associated methods be modified to accept an XmlWriter?

PR #694, opened 2020-03-20
Issue #2786 Amend SVG library to use XmlWriter instead of XmlTextWriter.

bbsgfalconer:
#### Reference Issue
Enhancement #693

Generalize XML output to use an XmlWriter instead of the more specific XmlTextWriter. In particular, my use case is that I want to use my own custom XmlWriter to send output as part of an Asp.Net Blazor application.

The changes to the various WriteXXX(...) methods are trivial. However, the XmlWriter class performs some stricter checks than the XmlTextWriter and so I had to change the handling of namespaces. The existing code specifies namespace simply by writing an "xmlns" attribute. However, with XmlWriter, this causes an exception to be thrown (similar to [this question on StackOverflow](https://stackoverflow.com/questions/23698767/the-prefix-cannot-be-redefined-from-to-url-within-the-same-start-element-t) )

Note also the Microsoft documentation for [XmlTextWriter](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmltextwriter?view=netframework-4.8):

> Starting with the .NET Framework 2.0, we recommend that you use the XmlWriter class instead.

> Starting with the .NET Framework 2.0, we recommend that you create XmlWriter instances by using the XmlWriter.Create method and the XmlWriterSettings class to take advantage of new functionality.

See also conformance options based on [ConformanceLevel](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmlwritersettings.conformancelevel?view=netframework-4.8)

#### Any other comments?

The changes impacted SvgAttributeAttribute.cs such that:

* the SVG namespace (http://www.w3.org/2000/svg) is no longer included in the list of namespaces that would be written into the root SvgFragment element from this list. Instead, all SvgElements are explicitly using this namespace as per SvgElement.cs
* the 'NamespaceAndName' method is no longer needed. Instead, XmlWriter.WriteAttributeString is called with the namespace and prefix to handle this. (again, in SvgElement.cs)

Since the SvgNamespace is needed more widely across SvgElement and SvgAttributeAttribute, I extracted it into its own static class.

All unit tests pass.

### DIFF
--- a/Source/Basic Shapes/SvgVisualElementStyle.cs
+++ b/Source/Basic Shapes/SvgVisualElementStyle.cs
@@ -1,5 +1,4 @@
 using System;
-using System.ComponentModel;
 
 namespace Svg
 {
@@ -7,13 +6,13 @@ namespace Svg
     {
         public virtual bool Visible
         {
-            get { return string.Equals(Visibility, "visible", StringComparison.OrdinalIgnoreCase); }
+            get { return string.Equals(Visibility.Trim(), "visible", StringComparison.OrdinalIgnoreCase); }
         }
 
         // Displayable - false if attribute display="none", true otherwise
         protected virtual bool Displayable
         {
-            get { return !string.Equals(Display, "none", StringComparison.OrdinalIgnoreCase); }
+            get { return !string.Equals(Display.Trim(), "none", StringComparison.OrdinalIgnoreCase); }
         }
 
         /// <summary>


mrbean-bremen:
Looks good to me. @H1Gdev, @wieslawsoltes - can one of you have another look please?

H1Gdev:
@mrbean-bremen 

I will check this PR.

Source/Document Structure/SvgFragment.cs
```diff
@@ -298,16 +293,14 @@ public override SvgElement DeepCopy<T>()
         }
 
         // Override the default behavior, writing out the namespaces.
-        protected override void WriteStartElement(XmlTextWriter writer)
+        protected override void WriteStartElement(XmlWriter writer)
```
H1Gdev: It also adds namespace to inner SVG.(existing issue...)
This process should be in `SvgDocument`.

Source/SvgElementFactory.cs
```diff
@@ -113,7 +113,7 @@ private SvgElement CreateElement<T>(XmlReader reader, bool fragmentIsDocument, S
             else
             {
                 // All non svg element (html, ...)
-                createdElement = new NonSvgElement(elementName);
+                createdElement = new NonSvgElement(elementName, elementNS);
```
H1Gdev: Does SVG embedded in HTML have an unnecessary namespace ?

H1Gdev:
What is "Issue #2786" ?

Source/Document Structure/SvgFragment.cs
H1Gdev: Fix #702.

H1Gdev:
Please separate PRs for each issue.

bbsgfalconer:
Oops - tidying up my branches. Well separate into smaller changes and resubmit.

Issue #696, opened 2020-03-23
Compile constant `Net4` missing


oerpli:
There's the `RegisterEvents` method which is only compiled if the target is Net4 ([see here](https://github.com/vvvv/SVG/blob/d7a767942de86a0263941dd3e7cd4f939709f036/Source/SvgElement.cs#L1066)) but the compile constant is not set ([see here](https://github.com/vvvv/SVG/blob/master/Source/Svg.csproj#L56)).

A few other methods are also affected: [Link](https://github.com/vvvv/SVG/search?q=net4&unscoped_q=net4)

Should I open a PR that adds the compile constant to the `.csproj`?

mrbean-bremen:
> Should I open a PR that adds the compile constant to the .csproj?

Sure, go ahead! Thanks!

PR #697, opened 2020-03-24
Added Net4 CompileConstant

oerpli:

Addresses #696 - a few functions that are only useful in the .net Framework version are gated with a compile constant (`Net4`) that is not defined in the current version. 

#### Any other comments?

Alternatively, the `#if`s could use the "correct" compile constants but I considered that doing it this way is nicer.

### DIFF
--- a/Source/Svg.csproj
+++ b/Source/Svg.csproj
@@ -53,7 +53,7 @@
 
   <PropertyGroup Condition="'$(TargetFramework)'=='net452'">
     <Title>Svg for .Net Framework 4.5.2</Title>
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
@H1Gdev , @wieslawsoltes - can one of you please review this? I'm a bit out of the cycle at the moment (still hoping for the new repo where you guys will have write access).

oerpli:
I also just checked the source with ripgrep for possible other undefined compilation constants but didn't find any:
```
rg "#if !?(.+)" -o -r '$1'  --no-line-number -I | sort | Get-Unique
```
>Net4
> NETCORE
> NETFULL
> NETSTANDARD || NETCORE
> NETSTANDARD20

PR #706, opened 2020-04-07
Supports localized family names.

H1Gdev:

- Localized family names

https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#font-family-prop


Source/SvgFontManager.cs
```diff
@@ -12,11 +12,19 @@ namespace Svg
     /// When a font is requested in the render process, if the font is not found as an embedded SvgFont, the render
     /// process will SvgFontManager.FindFont method.
     /// </summary>
-
     public static class SvgFontManager
     {
+        private static readonly string[][] LocalizedFamilyNames = new string[][]
+        {
+            // Japanese
```
H1Gdev, line 19: Add other localization if needed.

mrbean-bremen: Wouldn't it make sense to allow to add more localized names via the API? Collecting them all here will be a bit cumbersome. That being said, I agree that this certainly is an improvement :)

PR #713, opened 2020-04-17
Fix: #707 'Problems with entities/styles' issue.

### LINKED ISSUE #707: Problems with entities/styles
Hallo,

I have a problem with styles, for texts and rectangles, which are defined as entities in documenttype.
How can I get the refenced name of a style from a svgelement (text, rectangle).

I get for this text the attributes without style.

EntitiesStyles

H1Gdev:
#### Reference Issue

Fixes #707

Ref. #90

### DIFF
--- a/Source/SvgNodeReader.cs
+++ b/Source/SvgNodeReader.cs
@@ -130,7 +130,8 @@ namespace Svg
                 }
                 else
                 {
-                    _value = string.Empty;
+                    base.ResolveEntity();
+                    _value = ReadAttributeValue() ? base.Value : string.Empty;
                 }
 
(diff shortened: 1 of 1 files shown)


H1Gdev:
I fixed these issues, but I think that `SvgNodeReader` and `SvgTextReader` should be obsoleted.

These inherited classes have side effects and cause issues like this one.

These classes are needed for `SvgDocument` class `Open` method with `entities`.
But I don't think it's necessary because # 651 didn't work until it was resolved, and `Open(XmlDocument document, , Dictionary<string, string> entities)` doesn't exist.

PR #457, opened 2019-05-17
Sample design code for BaseUri.

H1Gdev:
I am considering relative paths in SVG.

Rough flow is below.

1. `Open`
2. `FlushStyles`
3. In `Open` method if exists path, then set `BaseUri` property
4. set `BaseUri` property
5. `Draw`

In 2. `FlushStyles`, `SvgPaintServerFactory` tries to resolve link values, but can not because it does not have base URI for relative path.

This PR is an implementation to set `BaseUri` in 1. `Open` and changes from clients are **not acceptable**.
This is better way, I think.

However, compatibility is broken because access attribute of `BaseUri` set property changes.

Need to keep `BaseUri` set property public ?

PR #721, opened 2020-04-29
Adding support for <link> css inclusion

FabioFerrettiMoodys:

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -260,32 +260,35 @@ namespace Svg
         /// Opens the document at the specified path and loads the SVG contents.
         /// </summary>
         /// <param name="path">A <see cref="string"/> containing the path of the file to open.</param>
+        /// <param name="baseUri"> base uri for linked file</param>
         /// <returns>An <see cref="SvgDocument"/> with the contents loaded.</returns>
         /// <exception cref="FileNotFoundException">The document at the specified <paramref name="path"/> cannot be found.</exception>
-        public static SvgDocument Open(string path)
+        public static SvgDocument Open(string path, Uri baseUri = null)
         {
-            return Open<SvgDocument>(path, null);
+            return Open<SvgDocument>(path, null, baseUri == null ? new Uri(path) : baseUri);
         }
 
         /// <summary>
         /// Opens the document at the specified path and loads the SVG contents.
         /// </summary>
         /// <param name="path">A <see cref="string"/> containing the path of the file to open.</param>
+        /// <param name="baseUri"> base uri for linked file</param>
         /// <returns>An <see cref="SvgDocument"/> with the contents loaded.</returns>
         /// <exception cref="FileNotFoundException">The document at the specified <paramref name="path"/> cannot be found.</exception>
-        public static T Open<T>(string path) where T : SvgDocument, new()
+        public static T Open<T>(string path, Uri baseUri = null) where T : SvgDocument, new()
         {
-            return Open<T>(path, null);
+            return Open<T>(path, null, baseUri);
         }
 
         /// <summary>
         /// Opens the document at the specified path and loads the SVG contents.
         /// </summary>
         /// <param name="path">A <see cref="string"/> containing the path of the file to open.</param>
+        /// <param name="baseUri"> base uri for linked file</param>
         /// <param name="entities">A dictionary of custom entity definitions to be used when resolving XML entities within the document.</param>
         /// <returns>An <see cref="SvgDocument"/> with the contents loaded.</returns>
         /// <exception cref="FileNotFoundException">The document at the specified <paramref name="path"/> cannot be found.</exception>
-        public static T Open<T>(string path, Dictionary<string, string> entities) where T : SvgDocument, new()
+        public static T Open<T>(string path, Dictionary<string, string> entities, Uri baseUri = null) where T : SvgDocument, new()
         {
             if (string.IsNullOrEmpty(path))
             {
@@ -299,8 +302,9 @@ namespace Svg
 
             using (var stream = File.OpenRead(path))
             {
-                var doc = Open<T>(stream, entities);
-                doc.BaseUri = new Uri(System.IO.Path.GetFullPath(path));
+                var doc = Open<T>(stream, entities, baseUri == null ? new Uri(path) : baseUri);
+                if (baseUri == null)
+                    doc.BaseUri = new Uri(System.IO.Path.GetFullPath(path));
                 return doc;
             }
         }
@@ -309,9 +313,10 @@ namespace Svg
         /// Attempts to open an SVG document from the specified <see cref="Stream"/>.
         /// </summary>
         /// <param name="stream">The <see cref="Stream"/> containing the SVG document to open.</param>
-        public static T Open<T>(Stream stream) where T : SvgDocument, new()
+        /// <param name="baseUri"> base uri for linked file</param>
+        public static T Open<T>(Stream stream, Uri baseUri = null) where T : SvgDocument, new()
         {
-            return Open<T>(stream, null);
+            return Open<T>(stream, null, baseUri );
         }
 
 
@@ -342,8 +347,9 @@ namespace Svg
         /// </summary>
         /// <param name="stream">The <see cref="Stream"/> containing the SVG document to open.</param>
         /// <param name="entities">Custom entity definitions.</param>
+        /// <param name="baseUri"> base uri for linked file</param>
         /// <exception cref="ArgumentNullException">The <paramref name="stream"/> parameter cannot be <c>null</c>.</exception>
-        public static T Open<T>(Stream stream, Dictionary<string, string> entities) where T : SvgDocument, new()
+        public static T Open<T>(Stream stream, Dictionary<string, string> entities, Uri baseUri = null) where T : SvgDocument, new()
         {
             if (stream == null)
             {
@@ -356,10 +362,10 @@ namespace Svg
                 XmlResolver = new SvgDtdResolver(),
                 WhitespaceHandling = WhitespaceHandling.None
             };
-            return Open<T>(reader);
+            return Open<T>(reader, baseUri);
         }
 
-        private static T Open<T>(XmlReader reader) where T : SvgDocument, new()
+        private static T Open<T>(XmlReader reader, Uri baseUri = null) where T : SvgDocument, new()
         {
             if (!SkipGdiPlusCapabilityCheck)
             {
@@ -435,6 +441,19 @@ namespace Svg
                             {
                                 styles.Add(unknown);
                             }
+                            else if (element.ElementName.ToLower() == "link")
+                            {
+                                if (element.CustomAttributes["rel"] == "stylesheet")
+                                {
+                                    var href = element.CustomAttributes["href"];
+                                    Uri uri = baseUri == null ? new Uri(href) : new Uri(baseUri, href);
+                                    var content = File.ReadAllText(uri.AbsolutePath);
+                                    SvgUnknownElement elem = new SvgUnknownElement("style");
+                                    element.Content = content;
+                                    styles.Add(element);
+
+                                }
+                            }
                             break;
                         case XmlNodeType.CDATA:
                         case XmlNodeType.Text:
@@ -491,8 +510,9 @@ namespace Svg
         /// Opens an SVG document from the specified <see cref="XmlDocument"/>.
         /// </summary>
         /// <param name="document">The <see cref="XmlDocument"/> containing the SVG document XML.</param>
+        /// <param name="baseUri"> base uri for linked file</param>
         /// <exception cref="ArgumentNullException">The <paramref name="document"/> parameter cannot be <c>null</c>.</exception>
-        public static SvgDocument Open(XmlDocument document)
+        public static SvgDocument Open(XmlDocument document, Uri baseUri = null)
         {
             if (document == null)
             {
@@ -500,7 +520,7 @@ namespace Svg
             }
 
             var reader = new SvgNodeReader(document.DocumentElement, null);
-            return Open<SvgDocument>(reader);
+            return Open<SvgDocument>(reader, baseUri);
         }
 
         public static Bitmap OpenAsBitmap(string path)


FabioFerrettiMoodys:
here is why the test are important ;)

i got a error in Open of svgDocument : i forgot the else ;)

       public static T Open<T>(string path, Dictionary<string, string> entities, Uri baseUri = null) where T : SvgDocument, new()
        {
            if (string.IsNullOrEmpty(path))
            {
                throw new ArgumentNullException("path");
            }

            if (!File.Exists(path))
            {
                throw new FileNotFoundException("The specified document cannot be found.", path);
            }

            using (var stream = File.OpenRead(path))
            {
                var doc = Open<T>(stream, entities, baseUri == null ? new Uri(System.IO.Path.GetFullPath(path)) : baseUri);
                if (baseUri == null)
                    doc.BaseUri = new Uri(System.IO.Path.GetFullPath(path));
                else
                    doc.BaseUri = baseUri;
                return doc;
            }
        }

Source/SvgDocument.cs
```diff
@@ -260,32 +260,35 @@ private static bool ExceptionCaughtIsGdiPlusRelated(Exception e)
         /// Opens the document at the specified path and loads the SVG contents.
         /// </summary>
         /// <param name="path">A <see cref="string"/> containing the path of the file to open.</param>
+        /// <param name="baseUri"> base uri for linked file</param>
         /// <returns>An <see cref="SvgDocument"/> with the contents loaded.</returns>
         /// <exception cref="FileNotFoundException">The document at the specified <paramref name="path"/> cannot be found.</exception>
-        public static SvgDocument Open(string path)
+        public static SvgDocument Open(string path, Uri baseUri = null)
```
H1Gdev, line 266: Adding `Open` argument doesn't seem to be active. (ref. #457)

Source/SvgDocument.cs
```diff
@@ -435,6 +443,19 @@ private static T Open<T>(XmlReader reader) where T : SvgDocument, new()
...
                                 styles.Add(unknown);
                             }
+                            else if (element.ElementName.ToLower() == "link")
+                            {
+                                if (element.CustomAttributes["rel"] == "stylesheet")
+                                {
+                                    var href = element.CustomAttributes["href"];
+                                    Uri uri = baseUri == null ? new Uri(href) : new Uri(baseUri, href);
+                                    var content = File.ReadAllText(uri.AbsolutePath);
+                                    SvgUnknownElement elem = new SvgUnknownElement("style");
```
H1Gdev, line 453: use `var`.

Source/SvgDocument.cs
FabioFerrettiMoodys: i don't understand.

mrbean-bremen:
@H1Gdev - are there open review issues here?

H1Gdev:
There is no `<link>` element in the SVG specifications.

- [SVG](https://www.w3.org/TR/SVG11/eltindex.html)
- [SVG2](https://www.w3.org/TR/SVG2/eltindex.html)

@FabioFerrettiMoodys 

Please add target SVG files to your test.
Please tell me supported software (major browser etc.).

FabioFerrettiMoodys:
Hello

https://www.w3.org/TR/SVG2/styling.html

```
6.3. External style sheets: the effect of the HTML ‘link’ element
An HTML ‘link’ element in an SVG document (that is, an element in the HTML namespace with local name "link") with its ‘rel’ attribute set to 'stylesheet' must be processed as defined in the HTML specification and cause external style sheets to be loaded and applied to the document. Such elements in HTML documents outside of an inline SVG fragment must also apply to the SVG content.
```

all major browser seems to support it 
i've test this sample with chrome, ie11 and firefox.
http://www.schepers.cc/svg/style/external-link-xhtml.svg

H1Gdev:
This PR has the following issues.
- File output by `Draw` is not supported.
- Do not support the case if `href` is specified in URL.

First, make the following changes.
- Create an element for \<link\>.
- `BaseUri` is a user-settable property, so please resolve `link` `href` later.

H1Gdev:
### FYI

Output SVG with this PR is as follows.
```svg
<svg width="100%" height="100%" viewBox="0, 0, 50, 50" xhtml="http://www.w3.org/1999/xhtml" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1">
  <title>Styling Test</title>
  <desc>
    Tests linking to external stylesheet with "link" element in XHML namespace.  
    Seems to work in Firefox, Opera, and Safari, as of 18-01-2009
  </desc>
  <metadata>author: schepers, created: 18-01-2009</metadata>
  <link href="http://www.schepers.cc/svg/style/svgstyle.css" rel="stylesheet" type="text/css" />
  <circle cx="25" cy="25" r="20" />
</svg>
```

FabioFerrettiMoodys:
> * `BaseUri` is a user-settable property, so please resolve `link` `href` later.

i don't know how to do it, you collect all the styles in the function Open<T>(XmlReader reader... 
You create a "resulting css" named "cssTotal" and assign the styles to te elements at the end of these function : before we can set the BaseUri.
line 480 of svgDocument.cs
` var cssTotal = styles.Select((s) => s.Content).Aggregate((p, c) => p + Environment.NewLine + c);
`

H1Gdev:
@mrbean-bremen 
@FabioFerrettiMoodys 

By #90, `BaseUri` property can be set...
We plan to ensure backward compatibility as much as possible.

This is modification to review style application sequence by supporting the SVG2 specification.

FabioFerrettiMoodys:
@H1Gdev 

> By #90, `BaseUri` property can be set...
> We plan to ensure backward compatibility as much as possible.
> 
> This is modification to review style application sequence by supporting the SVG2 specification.

I'm not sure to follow you : it can be set but only after the Open. but the styling is done in the Open. (SvgDocument.cs line 507 : svgDocument?.FlushStyles(true); ).
 i can make a link element and resolve the href later, but if a user set it, it will not work as all the styling is done at the open of the file. (if it's a new element, i will also need to redo the aggregate of styles at the end of the Open function. As the aggregate of the ISvGNode.Content will not work anymore.)

ps: sorry but i'm struggling to understand your comments and what and how you want me to do. I don't know your library as deep as you. Can you be more specific ? 

for exemple what do you mean by "File output by Draw is not supported." 
```
var svg = SvgDocument.Open(Source);
var image = svg.Draw();
```
this code work, do you mean something else ?

I also don't understand your "Output SVG with this PR is as follows."
how do you generate it like this ? i didn't alter the svgdocument structure, i've only added the css content to the list of styles used for the styling at the end of the Open.

H1Gdev:
> for exemple what do you mean by "File output by Draw is not supported."

It's a typo.

File output by `Write` is not supported.

H1Gdev:
>  it can be set but only after the Open. but the styling is done in the Open. (SvgDocument.cs line 507 : svgDocument?.FlushStyles(true); ).

This is because it doesn't support \<link\>.
This is because we was able to solve style here.

FabioFerrettiMoodys:
As the link element wasn't in the svg namespace i've added a NonSvgElementAttribute so the SvgElementFactory can create NonSvgElement by refraction like the SvgElement.

i've assigned the BaseUri just after the creation of the svgdocument in the open function : the svglink use the OwnerDocument.BaseUri when the content of the link is requested. (svgLink.GetLinkContentAsText)

i've corrected the svgDocument.Write : it write the <link> element, but i don't know if it is what you want as the styles are already added on each elements (by the flushstyles function).

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+            Preload,
+            Prerender,
+            Prev,
+            Search,
+            Stylesheet
+        }
+
+         public override SvgElement DeepCopy()
+        {
+            return DeepCopy<SvgTitle>();
```
H1Gdev, line 39: SvgLink

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+
+        [SvgAttribute("href")]
+        public string Href
+        {
+            get { return GetAttribute<string>("href", false,string.Empty); }
```
H1Gdev, line 46: add space.
`false, string.Empty`

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Text;
+
+namespace Svg
+{
+    [NonSvgElementAttribute("link")]
+    public class SvgLink : NonSvgElement
+    {
+        public SvgLink()
+          :base("link")
+		{
```
H1Gdev, line 15: Not use TAB.

Source/SvgDocument.cs
```diff
@@ -392,6 +400,8 @@ private static T Open<T>(XmlReader reader) where T : SvgDocument, new()
                             else
                             {
                                 svgDocument = elementFactory.CreateDocument<T>(reader);
+                                if(baseUri!=null)
```
H1Gdev, line 403: Add space.

Source/SvgElementFactory.cs
```diff
@@ -14,8 +14,9 @@ namespace Svg
     /// </summary>
     internal class SvgElementFactory
     {
-        private List<ElementInfo> availableElements;
-        private Parser cssParser = new Parser();
+		private List<ElementInfo> availableElements;
```
H1Gdev, line 17: What changed ?

FabioFerrettiMoodys: another automatic tabs/space remplacement by my texteditor configuration at work. sorry.

Source/Document Structure/SvgLink.cs
FabioFerrettiMoodys: should be ok

Source/Document Structure/SvgLink.cs
FabioFerrettiMoodys: should be ok

Source/Document Structure/SvgLink.cs
FabioFerrettiMoodys: should be ok

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+                content = sr.ReadToEnd();
+            }
+            stream.Dispose();
+            return content;
+        }
+
+
+        public Stream GetLinkContentAsStream(RelativeValue rel = RelativeValue.Unknown)
+        {
+            if (Rel != rel && rel != RelativeValue.Unknown)
```
H1Gdev, line 77: If you're doing a comparison with `Rel` property, why not use the value of `Rel` property value ??

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+            set { Attributes["href"] = value; }
+        }
+
+        [SvgAttribute("rel")]
+        public RelativeValue Rel
+        {
+            get { return GetAttribute<RelativeValue>("rel", false, RelativeValue.Unknown); }
+            set { Attributes["rel"] = value; }
+        }
+
```
H1Gdev, line 56: There is one blank line.

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+            Prev,
+            Search,
+            Stylesheet
+        }
+
+        public override SvgElement DeepCopy()
+        {
+            return DeepCopy<SvgLink>();
+        }
+
```
H1Gdev, line 41: There is one blank line.

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+namespace Svg
+{
+    [NonSvgElementAttribute("link")]
+    public class SvgLink : NonSvgElement
+    {
+        public SvgLink()
+          : base("link")
+        {
+
```
H1Gdev, line 16: No blank lines are needed.

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            get { return GetAttribute<RelativeValue>("rel", false, RelativeValue.Unknown); }
+            set { Attributes["rel"] = value; }
+        }
+
+
+
+        public string GetLinkContentAsText(RelativeValue rel)
+        {
+            var stream = GetLinkContentAsStream(rel);
```
H1Gdev, line 61: use `using`

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+            if (stream == null)
+                return null;
+
+            string content = null;
+            using (StreamReader sr = new StreamReader(stream))
```
H1Gdev, line 66: use `var`.

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,107 @@
...
+                var uri = new Uri(safeUriString, UriKind.RelativeOrAbsolute);
+
+                if (!uri.IsAbsoluteUri)
+                    uri = new Uri(OwnerDocument.BaseUri, uri);
+
+                // should work with http: and file: protocol urls
+                var httpRequest = WebRequest.Create(uri);
+
+                var webResponse = httpRequest.GetResponse();
+                var stream = webResponse.GetResponseStream();
```
H1Gdev, line 94: Please call `Dispose` when exception occurs.

Source/Document Structure/SvgLink.cs
FabioFerrettiMoodys: it is to avoid as much as possible the use of the content with a incorrect rel type.

if you call GetLinkContentAsStream(stylesheet) on a link  with rel="icon" for exemple, you will not get the content and cannot use it as a stylesheet by mistake (and make the code crash).

assuming you are absolutely sure of what you do and what you have, you can just call GetLinkContentAsStream(), you will get the content independently of the rel 

 these technique avoid to check everywhere if the Rel is correct every where you use the link content, here it is done in only one place.

I can remove if if you prefer.

mrbean-bremen:
@H1Gdev - can you please check if the current state can be merged? You did some review that has been addressed (admittetly, this was quite some time ago, but I hadn't been active in this project since then).

kimsey0:
(If #872 is merged and this pull request is resumed, it should check that `SvgDocument.ResolveExternalResources == true` before creating a `WebRequest` in `GetLinkContentAsStream`.)

paulushub:
@FabioFerrettiMoodys It looks like your `link` requirement is non-standard, or please provide some standard use cases.

paulushub:
@FabioFerrettiMoodys Sorry for the delay. Please update your project for review.

Source/Document Structure/SvgLink.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            try
+            {
+                var uri = new Uri(safeUriString, UriKind.RelativeOrAbsolute);
+
+                if (!uri.IsAbsoluteUri)
+                    uri = new Uri(OwnerDocument.BaseUri, uri);
+
+                // should work with http: and file: protocol urls
+                var httpRequest = WebRequest.Create(uri);
```
kimsey0, line 89: This should check that `SvgDocument.ResolveExternalResources` is true before creating a web request to avoid XSS vulnerabilities. See #873.

paulushub: @kimsey0 Can you fix it by editing the PR source?

kimsey0: It's been too long since I worked on this project, so I can't really remember the details, nor how tests etc. are structured. I can see that #873 split `SvgDocument.ResolveExternalResources` into three properties: `ResolveExternalXmlEntites`, `ResolveExternalImages`, and `ResolveExternalElements`, probably to give more fine-grained control over which external resources are allowed to be resolved. You may want to introduce a new `ExternalType ResolveExternalStylesheets` property, somehow pass it in or make it available to this `GetLinkContentAsStream` method, and then trace a warning and return null or an empty string if `!ResolveExternalStylesheets.AllowsResolving(uri)`.

paulushub: @kimsey0 Thank for the update. Just do not know the best way to handle these staled PR.
- Take over (clone PR reposition and fix issues), the original owner looses his/her contribution
- Close the PR and see how to reimplement the issue, a waste but may be worth it

mrbean-bremen: - adapt/finish the original PR, so it will just have 2 authors

paulushub: > * adapt/finish the original PR, so it will just have 2 authors

Never done, can you outline the procedure and give me some hints?

mrbean-bremen: If you are asking about the git workflow, something like:
> git remote add FabioFerrettiMoodys https://github.com/FabioFerrettiMoodys/SVG
> git checkout FabioFerrettiMoodys/master (probably under some other name)
> git rebase origin/main -> resolve conflicts as usual
> git push --force-with-lease
(or the equivalents if you use a git GUI client(
- make your changes, commit, push...

Is this what you are asking?

paulushub: > Is this what you are asking?

I think so, thanks - will look into it this weekend.

kimsey0: Depending on the policies of this project (the contribution guide doesn't say), you can also just avoid squashing the commits, leaving the original author for each one and making it clear who's done what.

mrbean-bremen: > the contribution guide doesn't say

Something to be added...

The current (undocumented) convention is that commits are usually squashed and only rebased if the individual commits are independent changes to be preserved, and have sensible commit comments (which is not the case here). Though it is possible to rebase the PR respectively before the merge, if a rebase is wanted.

Issue #726, opened 2020-05-07
Fontweight Bolder == Bold when Bounds are computed?


foxmichaels:
#### Description
Not sure if this is a known issue or not. Using the latest (as of now, 3.1.1), `Bounds` seems to treat "Bolder" (W900) the same as "Bold" (W700). This is despite them showing correctly (Bolder is thicker than Bold) in browsers and Illustrator and online SVG viewers.

#### Example data
```
SvgText txt1 = new SvgText
{
	FontFamily = "Helvetica",
	FontSize = new SvgUnit(SvgUnitType.Pixel, 12f),
	FontWeight = SvgFontWeight.W900
};
SvgTextSpan txts1 = new SvgTextSpan { Text = "TEST-FONT" };
txt1.Children.Add(txts1);
txt1.Bounds.Height = <value>
```
In the above code, `<value>` will be the same if `FontWeight` is set to W900 and W700. It will adjust properly when `FontWeight` is set to W400.

Have tested this on a handful of fonts (to rule out a font issue) - including:
HelveticaNeue, Helvetica, Times New Roman, Courier New, Arial.

#### Used Versions
SVG 3.1.1, .NET 4.8

I didn't see this listed as an issue, so thought I would mention it.

Thanks!

H1Gdev:
`font-weight` is displayed using `System.Drawing.FontStyle`(`Regular` or `Bold`).

It is assigned as follows.

|`font-weight`|`System.Drawing.FontStyle`|
|:-:|:-:|
|100 - 500|Regular|
|600 - 900|Bold|

So, 900 same as 700.

foxmichaels:
Interesting, thank you. Looking into it, I see that yeah, they're the same in System.Drawing.FontStyle.

Would it be possible to implement something using the `Font.FromHfont`, like:

```
IntPtr hFont = CreateFont(-13, 0, 0, 0, 100, 0, 0, 0, 1, 0, 0, 0, 0, "Arial\0");
Font.FromHfont(hFont)
```
I haven't yet clones the repository so I need to see first how the fonts are being generated. If it's possible, I may clone and try giving it a shot. :)

Thanks!
```

**Comment by foxmichaels** [comment]
Update - 
I tried it, just to rule it out as a possibility. This code:
```
[DllImport("gdi32.dll")]
static extern IntPtr CreateFont(int nHeight, int nWidth, int nEscapement,
   int nOrientation, int fnWeight, uint fdwItalic, uint fdwUnderline, uint
[...]

PR #727, opened 2020-05-09
Implement font-weight bolder and lighter.

H1Gdev:
#### Reference Issue

Ref. #726 

#### Any other comments?

Border between `Regular` and `Bold` is based on major browsers.


H1Gdev:
I will update release note after #728 is merged.

PR #737, opened 2020-06-04
Added UWP Target and UWP Dpi

inforithmics:
#### Reference Issue
https://github.com/vvvv/SVG/issues/736

Svg.Skia issue
https://github.com/wieslawsoltes/Svg.Skia/issues/16

Fixes Crash in UWP on Startup

used https://www.nuget.org/packages/MSBuild.Sdk.Extras/
for targeting UWP


mrbean-bremen:
@wieslawsoltes - can you please have a look?

Source/SvgDocument.cs
```diff
@@ -42,8 +42,11 @@ public override SvgOverflow Overflow
 
         private static int GetSystemDpi()
         {
+#if WINDOWS_UWP
+            var di = Windows.Graphics.Display.DisplayInformation.GetForCurrentView();
```
inforithmics, line 46: This needs to run on UI Thread or else it crashes.

inforithmics:
Before this can be merged I have to make it UI Thread save.

inforithmics:
So I made it Thread Save and it should work now.

FoggyFinder:
any updates?

inforithmics:
Another possibility to achieve this would be to Reference Xamarin.Essentials and get the information from there.

I think the Main question is which way you want to go supporting with this library.

a) Implement platform support from externally. (like svg.skia for rendering on different platforms, then platform specific things like dpi need to be set externally Then this pull request  https://github.com/vvvv/SVG/pull/739 is needed).

b) Implement platform support internally than this library needs to be compiled for every platform.
UWP, Android, iOs, Linux, Mac (I could make other pull requests to support it).

c) Implementing a and b than both pull requests are needed

inforithmics:
Another way fixing this would be to detect if i'm on UWP and return the Default 96 DPI

Code for detecting uwp platform

https://github.com/dotnet/corefx/blob/aa731cb68cd60128cc4b1efcd8b1a8e859750755/src/CoreFx.Private.TestUtilities/src/System/PlatformDetection.cs#L23

Probably it would do this with a third pull request. What's your opnion about that?

mrbean-bremen:
@H1Gdev , @wieslawsoltes : can you comment on this, please? Is this ready to be merged?

wieslawsoltes:
I prefer solution from https://github.com/vvvv/SVG/pull/739 PR.

mrbean-bremen:
Closing as discussed.

PR #739, opened 2020-06-07
Make Points Per Inch setable and only calculate on First Access

inforithmics:
#### Reference Issue
https://github.com/vvvv/SVG/issues/736

PointsPerInch is now a Property that can be set from outside, If it is not set from outside it is calculated by the default Implementation on first access


mrbean-bremen:
@wieslawsoltes, @gvheertum, @H1Gdev - I have no experience with UWP, can someone please review?

Tests/Svg.UnitTests/SvgDocumentsTests.cs
```diff
@@ -0,0 +1,22 @@
...
+            var testPointsPerInch = SvgDocument.PointsPerInch;
+            Assert.AreEqual(1, testPointsPerInch, "Test sets the Dpi to 1");
+
+            // resets the System Dpi
+            SvgDocument.PointsPerInch = pointsPerInch;
+            var comparePointsPerInch = SvgDocument.PointsPerInch;
+            Assert.AreEqual(pointsPerInch, comparePointsPerInch, "After setting to null the default Implementation should be taken");
+        }
+    }
+}
```
H1Gdev, line 22: Please apply [EditorConfig](https://github.com/vvvv/SVG/blob/nuget-3.1.1/Source/.editorconfig#L10) first.

H1Gdev:
If PR #737 can calculate the correct value, this PR seems unnecessary.
Why do you need this PR ?

inforithmics:
> If PR #737 can calculate the correct value, this PR seems unnecessary.
> Why do you need this PR ?

PR #737 Is accepted this Pull request isn't needed and can be closed.

inforithmics:
The idea behind this is that like 
SkipGdiPlusCapabilityCheck

it can be set from an external library like skia.svg and is only used on first access.

here the code in svk.skia
``
 static SKSvg()
        {
            SvgDocument.SkipGdiPlusCapabilityCheck = true;
        }
``

This avoids the problematic Code path that crashes the library in uwp release mode.

Long Term maybe this needs to be changed because the dpi settings are per screen and now only a global variable is set that 
doesn't change.
But beeing a property it could be changed from externally.

So this pull request is in my opinion complementary to my other pull request where the correct dpi for uwp is calculated

inforithmics:
Additional Information:

The reason the api needs to be changed is that accessing the class Initializes it so it needs to be a property to be able to be set.
Only exception handling wouldn't do it because in UWP this is an illegal operation that terminates the programm.

mrbean-bremen:
@wieslawsoltes - can me merge this one then?

inforithmics:
In my opinion yes, then the other https://github.com/vvvv/SVG/pull/737 pull request can be closed. I tried the new svg.skia version and it doesn't crash anymore in Release Mode on UWP

PR #753, opened 2020-07-27
Move SvgDocumentMetadata from "Document Structure" to "Metadata".

H1Gdev:

\<metadata\> element is classified as Metadata.
https://www.w3.org/TR/SVG11/metadata.html

So, move `SvgDocumentMetadata` from "Document Structure" to "Metadata".


Source/Metadata/SvgDocumentMetadata.cs
```diff
@@ -54,9 +47,8 @@ public override void InitialiseFromXML(XmlTextReader reader, SvgDocument documen
         {
```
H1Gdev, line 47: This method is not called from anywhere in this project...

I don't know how it was added from commit logs.
It doesn't seem to work.

PR #764, opened 2020-10-04
Added ElementInfo caching for faster SvgElement Creation

inforithmics:
This causes the Performance Test to finish in 10 seconds instead of 22 seconds without the optimizations.

#### Reference Issue
https://github.com/vvvv/SVG/issues/767

I profiled a Mobile App that uses this svg library and the profiler showed that a large part of the parsing of Svg is spent in 
creating the ElementInfo, I cache it statically, because it can't change during loading.

#### Any other comments?
Added an unit test for performance that times out after 25 seconds. Maybe the timeout can be adjusted downwards.
On my machine this Unit Test runs in 10 seconds.

### DIFF
--- a/Source/Css/SvgElementOps.cs
+++ b/Source/Css/SvgElementOps.cs
@@ -16,8 +16,7 @@ namespace Svg.Css
 
         public Selector<SvgElement> Type(NamespacePrefix prefix, string name)
         {
-            var types = _elementFactory.AvailableElements.Where(e => e.ElementName.Equals(name)).Select(e => e.ElementType);
-            if (types.Any())
+            if (_elementFactory.AvailableElementsDictionary.TryGetValue(name, out var types))
             {
                 return nodes => nodes.Where(n => types.Contains(n.GetType()));
             }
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
@H1Gdev, anyone else interested: this and the other performance PR seem to be ready to be merged, is there anything preventing the merge from your perspective?

wieslawsoltes:
I am working on similar perf improvement in #772, Source Generators have advantage of being compile time operation, also remove usage of `Activator.CreateInstance` which should add even more performance.

mrbean-bremen:
> I am working on similar perf improvement in #772

Would this replace this PR, or add to it?

mrbean-bremen:
@wieslawsoltes: if you are adding to it, it would make sense to merge it (and the other performance PR) now, what do you think?

wieslawsoltes:
> @wieslawsoltes: if you are adding to it, it would make sense to merge it (and the other performance PR) now, what do you think?

You can merge this PR, I can port my changes, no problem.

wieslawsoltes:
> > I am working on similar perf improvement in #772
> 
> Would this replace this PR, or add to it?

I can add my changes after this PR

mrbean-bremen:
Ok, I merged the two performance PRs (and also a small fix I made).

wieslawsoltes:
> Ok, I merged the two performance PRs (and also a small fix I made).

Great, will update my PR

PR #770, opened 2020-12-30
Make sure that paths with single moveto don't display line caps

### LINKED ISSUE #634: zero-length sub path are rendered with a line-cap with only single moveto
#### Description
An SVG path with only a single moveTo renders a line-cap (round dark circle in the example below).

[SVG1.1](https://www.w3.org/TR/SVG11/painting.html#StrokeProperties) says:
> A subpath (see Paths) consisting of a single moveto shall not be stroked.

[SVG2](https://svgwg.org/svg-next/painting.html#SpecifyingStrokePaint) says:
> Note that when stroking a ‘path’ element, any subpath consisting of a moveto but no following line drawing command will not be stroked.

#### Example data
````svg
<svg version="1.0" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" width="80px"
	 height="80px" viewBox="0 0 80 80" enable-background="new 0 0 80 80" xml:space="preserve">
	<path stroke="#1A171B" stroke-width="2" stroke-linecap="round" d="M10,10"/>
</svg>

````

mrbean-bremen:
- set stroke to None
- fixes #634


mrbean-bremen:
@H1Gdev, whoever is around: please have a look. This is a minor fix to a workaround added for another issue.

PR #780, opened 2021-01-03
Use Span in SvgPathBuilder and update Samples

wieslawsoltes:

* Uses Span for parsing svg path data to reduce string allocations and improve performance.

Below changes enable `dotnet build ./Source/Svg.sln` - at least on Windows (WindowsForms are windows only).
* Added netcoreapp3.0 target to enable testing with SvgW3CTestRunner.
* Updated Samples to be compatible with dotnet command-line tooling for building.
* Update Nerdbank.GitVersioning package to fix dotnet build errors.


inforithmics:
Good work, I would have done the same changes.

Source/Paths/SvgPathBuilder.cs
```diff
@@ -32,8 +32,13 @@ public static SvgPathSegmentList Parse(string path)
...
             try
             {
-                foreach (var commandSet in SplitCommands(path.TrimEnd(null)))
-                    CreatePathSegment(commandSet[0], segments, new CoordinateParser(commandSet.Trim()));
+                var pathTrimmed = path.AsSpan().TrimEnd();
+                var commands = SplitCommands(ref pathTrimmed);
+                foreach (var commandSet in commands)
+                {
+                    var commandSetTrimmed = pathTrimmed.Slice(commandSet.Start, commandSet.Length).Trim();
+                    CreatePathSegment(commandSetTrimmed[0], segments, new CoordinateParser(ref commandSetTrimmed), ref commandSetTrimmed);
```
inforithmics, line 40: new CoordinateParser(ref commandSetTrimmed),

I think the Parser could be kept in a variable var parser = new CoordinateParser();
and then in the iteration this could be called.
parser.Init(ref commmandSetTrimmed);

In the Init method the all would be initialized like in the Constructor

wieslawsoltes: Add Init method to CoordinateParser

wieslawsoltes:
@mrbean-bremen This is ready for review, would be nice to have merged if possible before continuing my work on #772

mrbean-bremen:
I'm reviewing this right now, but it would be nice if @inforithmics could also have a look. I haven't use `Span`, so I may not be the best reviewer here (actually I haven't done much in C# at all in the last time).

mrbean-bremen (approved):
Looks good to me - as I wrote, would like to have input from @inforithmics before merging.

Source/Paths/CoordinateParser.cs
```diff
@@ -17,19 +18,25 @@ private enum NumState
...
-            if (char.IsLetter(coords[0])) ++i;
+        }
+
+        public void Init(ref ReadOnlySpan<char> chars)
+        {
+            _currState = NumState.separator;
+            _newState = NumState.separator;
+            i = 0;
+            Position = 0;
+            HasMore = chars.Length <= 0 ? false : true;
```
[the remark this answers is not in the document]
wieslawsoltes: I had issue with that check so I made sure it's very explicit if.

Samples/Entities/Entities.csproj
```diff
@@ -1,73 +1,12 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
wieslawsoltes: It's new csproj format, I will double check with VS, but worked fine with dotnet command line.

inforithmics:
I testet this pull request with UWP and it worked without problems. Then I profiled the Code and saw following

image

i think here it should be assigned to a variable like var currentChar = path[i]; and then the comparisions should be made

**f (char.IsLetter(path[i]) && path[i] != 'e' && path[i] != 'E')**

And this line is althoug several times used the length of the span doesn't change so it can be replaced with a variable.
var length = path.length;

path.Length 

I did the Performance Profiling in UWP, where Span is not optimized only with .NET 6 it will be optimized. So this could be a temporary Problem.

Source/DataTypes/SvgPointCollection.cs
```diff
@@ -60,10 +60,12 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
         {
             if (value is string)
             {
-                var parser = new CoordinateParser(((string)value).Trim());
+                var coords = ((string) value).AsSpan().Trim();
+                var parser = new CoordinateParser();
```
inforithmics, line 64: Is this Method often called, when yes the parser Could be an instance variable and if it is null on the First access initialized

Source/Paths/SvgPathBuilder.cs
```diff
@@ -217,31 +224,38 @@ private static PointF ToAbsolute(float x, float y, SvgPathSegmentList segments,
             return point;
         }
 
-        private static IEnumerable<string> SplitCommands(string path)
+        private static List<(int Start, int Length)> SplitCommands(ref ReadOnlySpan<char> path)
```
inforithmics, line 227: Is there a reason why the return type is not List<(ReadOnlySpan<char>) where the slices are returned

Samples/Entities/Entities.csproj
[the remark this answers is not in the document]
wieslawsoltes: Tested in Visual Studio and all projects work.

Source/DataTypes/SvgPointCollection.cs
wieslawsoltes: Made changes as suggested.

Source/Paths/SvgPathBuilder.cs
wieslawsoltes: You can use ReadOnlySpan<char> as type argument.

wieslawsoltes: The best solution is to inline the SplitCommands so no List or enumerator allocations are made.

wieslawsoltes:
> I testet this pull request with UWP and it worked without problems. Then I profiled the Code and saw following
> 
> image
> 
> i think here it should be assigned to a variable like var currentChar = path[i]; and then the comparisions should be made
> 
> **f (char.IsLetter(path[i]) && path[i] != 'e' && path[i] != 'E')**
> 
Done.

> And this line is althoug several times used the length of the span doesn't change so it can be replaced with a variable.
> var length = path.length;
> 
> path.Length
> 
Done.

> I did the Performance Profiling in UWP, where Span is not optimized only with .NET 6 it will be optimized. So this could be a temporary Problem.

Yes this is target runtime dependent, I suspect .NET 5 will have better perf for example.

Source/Paths/CoordinateParser.cs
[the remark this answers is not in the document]
wieslawsoltes: Done

Source/Paths/SvgPathBuilder.cs
inforithmics: Or the return type could be an enumerable and the slices could be yielded.

wieslawsoltes: Not possible, can not return slices.

wieslawsoltes: I will just inline SplitCommands as it waste too much resources by allocating list or enumerable.

wieslawsoltes:
@inforithmics I think I have addressed all your review comments, this can be tested again.

inforithmics:
image

So this version is now faster on UWP too. (And uses a lot less memory). So good work. This can now be merged in my opinion

inforithmics (approved):

Issue #769, opened 2020-12-04
text-anchor middle not rendering correctly


Stoffelche:
#### Description
I have a simple SVG, where the text with text-anchor middle is rendered incorrectly 
#### Example data
```
<svg xmlns="http://www.w3.org/2000/svg" >
	<rect fill="#0669b2" stroke="none" stroke-width="0" width="120" height="60"  />
	<svg width="120" height="60">
		<text font-size="11pt" fill="white" text-anchor="middle" transform="matrix(1 0 0 1 60 30)" >
			<tspan dy="0.3em">Some text</tspan>
		</text>
	</svg>
</svg>
```

SVG.Net renders: 
![grafik](https://user-images.githubusercontent.com/8353574/101196108-9cc74f00-3660-11eb-8615-38be81b01d9e.png)
but the correct rendering should be:
![grafik](https://user-images.githubusercontent.com/8353574/101196281-d0a27480-3660-11eb-8844-b5dc7404cebd.png)

(don't look at the image paddings they are just screenshots, one from your SVG Viewer the other from a browser).
The point is that, SVG.Net starts the text in the middle position but does not position the the text horizontally centered.

#### Used Versions
Master

Stoffelche:
I found, that when I initialize _xAnchor to 0 in the second ctor of TextDrawingState<br/>
` public TextDrawingState(TextDrawingState parent, SvgTextBase element)`<br/>
then my sample renders correctly. Then I tried another sample, which has more nodes and connections inbetween (the samples are from svg rendered by joint.js). And with those I experienced the following strange behavior:<br>
When I format th xml with line breaks and tab-indents, the problem with the text not being aligned to middle occurs again.<br/> 
However, if i delete all these \r, \n, \t, then the text is aligned correctly, after my change mentioned above.<br/>
Also, when I tried to debug the formatted xml, I found that the white chars are meddled somehow with the text-attributes of the tspan elements. <br/>
The formatting does not have any effect on the rendering in browsers, so I suspect, there is something wrong with respect to handing of it, in the svg.net code as well.

tebjan:
Thanks for the detailed description!

Stoffelche:
If you want to look into the problem with the formatted xml:<br/>
[TestSVG.zip](https://github.com/vvvv/SVG/files/5714496/TestSVG.zip)
I have uploaded a zip with the sources for the svg with and without formatting

mrbean-bremen:
May also be related to #742.

[...]

PR #782, opened 2021-01-04
Fixed rendering of text-anchor middle

mrbean-bremen:
- fix provided by Stefan Zschocke (@stoffelche)
- see #769

Make sure that both constructors use the same code.


Stoffelche:
looks good to me

PR #783, opened 2021-01-05
Add Property to Disable Dtd Processing

inforithmics:
Disable Dtd Processing, is not necessary.
1. Xml is not validated.
2. some svg's don't have a dtd
3. Designed Svgs have a dtd, this causes to parse the dtd which takes a long time.

Performance improvement

image

Here the Read of SvgTextReader takes 2.2 sec the reason is that the dtds are parsed but not needed 
This pull request reduces this to almos 0


wieslawsoltes:
I would make this a static flag on SvgDocument
```C#
public static bool DisableDtdProcessing { get; set; } = true;
```

inforithmics:
> I would make this a static flag on SvgDocument
> 
> ```cs
> public static bool DisableDtdProcessing { get; set; }
> ```

Good idea, because it seems it is needed.

wieslawsoltes:
Looking good.

wieslawsoltes:
@inforithmics I would add also:
```C#
XmlResolver = SvgDocument.DisableDtdProcessing ? new XmlUrlResolver() : new SvgDtdResolver(),
```

mrbean-bremen (approved):
@wieslawsoltes - this looks good to me, I would merge it now, if you don't mind.

wieslawsoltes:
> @wieslawsoltes - this looks good to me, I would merge it now, if you don't mind.

Go ahead look good.

PR #787, opened 2021-01-08
Migrate to supported .NET Core versions

wieslawsoltes:

1. Migrates to supported .NET Core versions:
    * from .NET Core 2.2 (Out of support version) to .NET Core 2.1
    * from .NET Core 3.0 (Out of support version) to .NET Core 3.1
2. Adds support for .NET 5.0 as it's is currently [recommended](https://dotnet.microsoft.com/download) .NET SDK

.NET Core 2.1 and 3.1 are LTS (Long-term support) releases, .NET 5.0 is basically basically new version of netstandard.

https://dotnet.microsoft.com/download/dotnet-core


wieslawsoltes:
I think adding `net5.0` target would also be good, we are already using it on AppVeyor for building and Source Generators require new compiler toolchain so this would not introduce any new dependency.

wieslawsoltes:
I will fix later samples, there are some breaking changes in `netcoreapp3.1` regarding Windows Forms.

mrbean-bremen (approved):
Looks good to me, and makes sense in my opinion, though I would like to hear from @tebjan about this (regarding removing support for unsupported versions).
I got an initial GH Actions workflow running based on this branch, though [one test is failing](https://github.com/mrbean-bremen/SVG/runs/1673428737?check_suite_focus=true) both in .NetCore 3.1 and .Net 5 (`__issue-084-01` comparison test). I have not looked into this, but would probably handle this in a separate issue, if this is not a false positive (which happens with the comparison tests).

wieslawsoltes:
> Looks good to me, and makes sense in my opinion, though I would like to hear from @tebjan about this (regarding removing support for unsupported versions).

Well I am and moving netcoreapp2.2 to netcoreapp2.1 so it's is event better, netcoreapp3.0 to netcoreapp3.1 is not that bad as we have netcoreapp2.1, so I think in terms of compatibility we are pretty good. The net5.0 is pretty important as it is becoming new standard target.  

> I got an initial GH Actions workflow running based on this branch, though [one test is failing](https://github.com/mrbean-bremen/SVG/runs/1673428737?check_suite_focus=true) both in .NetCore 3.1 and .Net 5 (`__issue-084-01` comparison test). I have not looked into this, but would probably handle this in a separate issue, if this is not a false positive (which happens with the comparison tests).

I have checked the test (quickly) and it seems that issue is with saving and reloading the svg, not sure what is wrong, maybe I need to update System.Drawing package.

mrbean-bremen:
I think it makes sense for @tebjan to make a release with the old supported versions, and merge this afterwards. I would make a PR for GitHub actions after this is merged (with the current version, the tests for .NetCore 2.2 fail because it is not supported). As for the failing test - if you don't find the problem right away, I would create a new issue for this and temproraily remove this image from the tests. It still works with the older versions, so this is not a regression.

wieslawsoltes:
> I think it makes sense for @tebjan to make a release with the old supported versions, and merge this afterwards.

That's fine with me.

> I would make a PR for GitHub actions after this is merged (with the current version, the tests for .NetCore 2.2 fail because it is not supported). 

You need to install:
-  .NET Core sdk 2.2.207
-  .NET Core sdk 3.0.103
to make it work.

> As for the failing test - if you don't find the problem right away, I would create a new issue for this and temproraily remove this image from the tests. It still works with the older versions, so this is not a regression.

The new issue would be good.

wieslawsoltes:
Testing the failing test:

```C#
       [Test]
        public void Test_Issue_084_01()
        {
            var svgPath = Path.Combine(AssemblyDirectory, "..", "..", "..", "..", "W3CTestSuite", "svg");
            var file = Directory.GetFiles(svgPath, "__issue-084-01.svg").FirstOrDefault(); 
            var doc1 = SvgDocument.Open<SvgDocument>(file);

            using (var memStream = new MemoryStream())
            {
                var targetFrameworkAttribute = Assembly.GetExecutingAssembly()
                    .GetCustomAttributes(typeof(System.Runtime.Versioning.TargetFrameworkAttribute), false)
                    .SingleOrDefault() as System.Runtime.Versioning.TargetFrameworkAttribute;
                
                
                doc1.Write(memStream);
                memStream.Position = 0;

                var streamReader = new StreamReader(memStream);
                var str = streamReader.ReadToEnd();
                File.WriteAllText($"c:\\DOWNLOADS\\__issue-084-01_{targetFrameworkAttribute.FrameworkName}.svg", str);

                memStream.Position = 0;

                var baseUri = doc1.BaseUri;
                var doc2 = SvgDocument.Open<SvgDocument>(memStream);
                doc2.BaseUri = baseUri;
            }
```

[__issue-084-01.zip](https://github.com/svg-net/SVG/files/5791144/__issue-084-01.zip)

wieslawsoltes:
Looks like the `.NETCoreApp,Version=v3.1` and `.NETCoreApp,Version=v5.0` produce different floating points for some reason.

wieslawsoltes:
Seems to be related https://devblogs.microsoft.com/dotnet/floating-point-parsing-and-formatting-improvements-in-net-core-3-0/

mrbean-bremen:
> __issue-084-01.zip

Cannot download this file, as it supposedly contains a virus.

wieslawsoltes:
> > __issue-084-01.zip
> 
> Cannot download this file, as it supposedly contains a virus.

It contains only svg's.

[__issue-084-01.svg.txt](https://github.com/svg-net/SVG/files/5791226/__issue-084-01.svg.txt)
[__issue-084-01_.NETCoreApp,Version=v2.1.svg.txt](https://github.com/svg-net/SVG/files/5791227/__issue-084-01_.NETCoreApp.Version.v2.1.svg.txt)
[__issue-084-01_.NETCoreApp,Version=v3.1.svg.txt](https://github.com/svg-net/SVG/files/5791228/__issue-084-01_.NETCoreApp.Version.v3.1.svg.txt)
[__issue-084-01_.NETCoreApp,Version=v5.0.svg.txt](https://github.com/svg-net/SVG/files/5791229/__issue-084-01_.NETCoreApp.Version.v5.0.svg.txt)
[__issue-084-01_.NETFramework,Version=v4.5.2.svg.txt](https://github.com/svg-net/SVG/files/5791230/__issue-084-01_.NETFramework.Version.v4.5.2.svg.txt)
[__issue-084-01_.NETFramework,Version=v4.6.1.svg.txt](https://github.com/svg-net/SVG/files/5791231/__issue-084-01_.NETFramework.Version.v4.6.1.svg.txt)

wieslawsoltes:
@mrbean-bremen I will add float extension method for ToString, so we have one place for all conversion from float to string and can control output.

Source/DataTypes/SvgViewBox.cs
```diff
@@ -242,8 +242,8 @@ public override object ConvertTo(ITypeDescriptorContext context, CultureInfo cul
                 var viewBox = (SvgViewBox)value;
 
                 return string.Format("{0}, {1}, {2}, {3}",
-                    viewBox.MinX.ToString(CultureInfo.InvariantCulture), viewBox.MinY.ToString(CultureInfo.InvariantCulture),
-                    viewBox.Width.ToString(CultureInfo.InvariantCulture), viewBox.Height.ToString(CultureInfo.InvariantCulture));
+                    viewBox.MinX.ToString(CultureInfo.InvariantCulture), viewBox.MinY.ToSvgString(),
+                    viewBox.Width.ToString(CultureInfo.InvariantCulture), viewBox.Height.ToSvgString());
```
mrbean-bremen, line 246: Have you checked if this impacts the performance?

wieslawsoltes: No, will create benchmark.

wieslawsoltes: Performance is the same.

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-4790K CPU 4.00GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.101
  [Host]     : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT
  Job-AOWSNB : .NET Core 3.1.10 (CoreCLR 4.700.20.51601, CoreFX 4.700.20.51901), X64 RyuJIT

Runtime=.NET Core 3.1

|            Method |     Mean |   Error |  StdDev |        Op/s | Ratio | Rank |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------ |---------:|--------:|--------:|------------:|------:|-----:|-------:|------:|------:|----------:|
|    Float_ToString | 638.4 ns | 1.68 ns | 1.49 ns | 1,566,483.8 |  1.00 |    1 | 0.0381 |     - |     - |     160 B |
| Float_ToSvgString | 640.9 ns | 2.20 ns | 1.95 ns | 1,560,194.7 |  1.00 |    1 | 0.0381 |     - |     - |     160 B |

mrbean-bremen:
[Here is the test result] for thid branch (with the failing test, removed, Windows only).

wieslawsoltes:
> [Here is the test result](https://github.com/mrbean-bremen/SVG/actions/runs/474560451) for this branch (with the failing test removed, Windows only).

This is great, I think I have smaller repro for failing issue.

mrbean-bremen:
Do you think it makes sense to add all or some benchmark tests to the CI build? Currently only the unittests are run.

wieslawsoltes:
> Do you think it makes sense to add all or some benchmark tests to the CI build? Currently only the unittests are run.

I don't think they will be comparable, probably by limited CI machine resources and will increase the build times.

Maybe we can add separate job only for benchmarks, the cpu performance will vary but memory usage can be useful.

mrbean-bremen:
> Maybe we can add separate job only for benchmarks

Yes, this is what I have been thinking.

wieslawsoltes:
> > Do you think it makes sense to add all or some benchmark tests to the CI build? Currently only the unittests are run.
> 
> I don't think they will be comparable, probably by limited CI machine resources and will increase the build times.
> 
> Maybe we can add separate job only for benchmarks, the cpu performance will vary but memory usage can be useful.

I would start we this benchmark:
```
dotnet run -c Release -f netcoreapp3.1 -- -f '*SvgDocument_*'
```

mrbean-bremen:
Yeah, this [increases the build times](https://github.com/mrbean-bremen/SVG/runs/1677492355?check_suite_focus=true) a bit, but that is still acceptable. Could further parallize the tests, if needed, and probably add an option to show less output.

wieslawsoltes:
> Yeah, this [increases the build times](https://github.com/mrbean-bremen/SVG/runs/1677492355?check_suite_focus=true) a bit, but that is still acceptable. Could further parallize the tests, if needed, and probably add an option to show less output.

Yes that looks acceptable.

mrbean-bremen:
@wieslawsoltes - I'm going to merge this now, if you don't object - the release is out!

wieslawsoltes:
> @wieslawsoltes - I'm going to merge this now, if you don't object - the release is out!

@mrbean-bremen No objections, ready to merge.

PR #793, opened 2021-01-11
Make up for PR #564.

H1Gdev:
#### What does this implement/fix? Explain your changes.

- Make up for PR #564.
- Pass pservers-pattern-03-f.svg.

PR #794, opened 2021-01-11
Fix size zero pattern issue

H1Gdev:
- If either width or height is zero.

- Corrects PR #793.


mrbean-bremen (approved):
Thanks - overlooked this, too..

PR #797, opened 2021-01-12
Add GitHub actions status badges

wieslawsoltes:


mrbean-bremen (approved):
Ha! I added exactly the same change a minute ago - takes yours in this case :)

PR #786, opened 2021-01-07
Parse performance optimizations

wieslawsoltes:
Optimizes performance and memory usage for Parse operations.

This PR has been split and summary moved to tracking issue #802

Split status:
- [x] Source/DataTypes/SvgNumberCollection.cs #807
- [x] Source/DataTypes/SvgPointCollection.cs #808
- [x] Source/DataTypes/SvgUnitCollection.cs #805
- [x] Source/DataTypes/SvgUnitConverter.cs #805
- [x] Source/Helpers/FloatParser.cs #801
- [x] Source/Helpers/StringSplitEnumerator.cs #801
- [x] Source/Painting/SvgColourConverter.cs #806
- [x] Source/Paths/CoordinateParser.cs #804
- [x] Source/Paths/SvgPathBuilder.cs #810
- [x] Source/Transforms/SvgTransformConverter.cs #809
- [x] Tests/Svg.Benchmark/CoordinateParserBenchmarks.cs #804
- [x] Tests/Svg.Benchmark/Program.cs #803
- [x] Tests/Svg.Benchmark/README.md #803
- [x] Tests/Svg.Benchmark/Svg.Benchmark.csproj #803
- [x] Tests/Svg.Benchmark/SvgColourConverterBenchmarks.cs #806
- [x] Tests/Svg.Benchmark/SvgDocumentBenchmarks.cs #803
- [x] Tests/Svg.Benchmark/SvgNumberCollectionConverterBenchmarks.cs #807
- [x] Tests/Svg.Benchmark/SvgPathBuilderBenchmarks.cs #810
- [x] Tests/Svg.Benchmark/SvgPointCollectionConverterBenchmarks.cs #808
- [x] Tests/Svg.Benchmark/SvgTransformConverter.cs (only rename) #803
- [x] Tests/Svg.Benchmark/SvgTransformConverterBenchmarks.cs #809
- [x] Tests/Svg.Benchmark/SvgUnitCollectionConverterBenchmarks.cs #805
- [x] Tests/Svg.Benchmark/SvgUnitConverterBenchmarks.cs #805
- [x] Tests/Svg.UnitTests/MarkerEndTest.cs #810
- [x] Tests/Svg.UnitTests/PerformanceTest.cs

PR #801, opened 2021-01-14
Adds common string parser class with parse helper methods

wieslawsoltes:
#### Reference Issue

Split from #786

Adds common string parser class with parse helper methods.

### DIFF
--- a/Source/Helpers/StringParser.cs
+++ b/Source/Helpers/StringParser.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Globalization;
+
+namespace Svg.Helpers
+{
+    internal static class StringParser
+    {
+        private static readonly CultureInfo Format = CultureInfo.InvariantCulture;
+
+        public static float ToFloat(ref ReadOnlySpan<char> value)
+        {
+#if NETSTANDARD2_1 || NETCORE || NETCOREAPP2_1 || NETCOREAPP3_1 || NET5_0
+            return float.Parse(value, NumberStyles.Float, Format);
+#else
+            return float.Parse(value.ToString(), NumberStyles.Float, Format);
+#endif
+        }
+
+        public static float ToFloatAny(ref ReadOnlySpan<char> value)
+        {
+#if NETSTANDARD2_1 || NETCORE || NETCOREAPP2_1 || NETCOREAPP3_1 || NET5_0
+            return float.Parse(value, NumberStyles.Any, Format);
+#else
+            return float.Parse(value.ToString(), NumberStyles.Any, Format);
+#endif
+        }
+
+        public static double ToDouble(ref ReadOnlySpan<char> value)
+        {
+#if NETSTANDARD2_1 || NETCORE || NETCOREAPP2_1 || NETCOREAPP3_1 || NET5_0
+            return double.Parse(value, NumberStyles.Any, Format);
+#else
+            return double.Parse(value.ToString(), NumberStyles.Any, Format);
+#endif
+        }
+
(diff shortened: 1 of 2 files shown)


wieslawsoltes:
@mrbean-bremen Can we do quick review of this PR, this is necessary to split the rest of #786 PR

mrbean-bremen (approved):

H1Gdev:
@wieslawsoltes 
@mrbean-bremen 
What are rules for using these helper methods ?

Source/Helpers/StringParser.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Svg.Helpers
+{
+    internal static class StringParser
+    {
+        private static readonly CultureInfo Format = CultureInfo.InvariantCulture;
+
+        public static float ToFloat(ref ReadOnlySpan<char> value)
+        {
+#if NETSTANDARD2_1 || NETCORE || NETCOREAPP2_1 || NETCOREAPP3_1 || NET5_0
```
H1Gdev, line 12: Considering the future, I think that the following will improve maintainability.
```cs
#if Net4
    return float.Parse(value.ToString(), NumberStyles.Float, Format);
#else
    return float.Parse(value, NumberStyles.Float, Format);
#endif
```

H1Gdev:
@wieslawsoltes 
What do we need to be aware of in terms of performance ? (Method ? Arguments ?)

wieslawsoltes:
> @wieslawsoltes
> @mrbean-bremen
> What are rules for using these helper methods ?

See the usage examples https://github.com/svg-net/SVG/pull/786/files

They follow original code usage.

wieslawsoltes:
> @wieslawsoltes
> What do we need to be aware of in terms of performance ? (Method ? Arguments ?)

Do not allocate new strings e.g." "string".Trim() allocates, use "string".AsSpan() etc. see the #786 PR

Source/Helpers/StringParser.cs
wieslawsoltes: NETSTANDARD2_0 does not support that API

H1Gdev: @wieslawsoltes 
Add NETSTANDARD2_0, too.
The newly added API will not be removed in the future.
However, as it is, it has to be added to `#if` every time the version of Framework goes up.

wieslawsoltes: It requires netstandard2.1

H1Gdev:
```cs
#if Net4 || NETSTANDARD2_0
    return float.Parse(value.ToString(), NumberStyles.Float, Format);
#else
    return float.Parse(value, NumberStyles.Float, Format);
#endif
```

wieslawsoltes: Now yes.

Source/Helpers/StringSplitEnumerator.cs
```diff
@@ -35,7 +35,7 @@ public bool MoveNext()
                     return true;
                 }
 
-                var slice = span.Slice(0, index).Trim();
+                var slice = span.Slice(0, index);
                 _str = span.Slice(index + 1);
                 if (slice.Length == 0)
```
H1Gdev, line 40: https://github.com/svg-net/SVG/blob/5dc4460eaf9819909b8a3beb993f3ba4eb7dbf6e/Source/Helpers/StringSplitEnumerator.cs#L38-L44

The following code is more efficient.
```cs
    _str = span.Slice(index + 1);
    if (index == 0)
        continue;
    Current = new StringPart(span.Slice(0, index));
```

PR #809, opened 2021-01-14
Refactor SvgTransformConverter to have Parse method and use StringParser

wieslawsoltes:
#### Reference Issue

Split from #786

Refactor SvgTransformConverter to have Parse method and use common StringParser

We want to have all converters the same Parse method signature for string conversion to unify API and this can be in feature used to optimize property setter to avoid type converters.


wieslawsoltes:
@mrbean-bremen This is ready for review and merge, this is just refactoring to unify parse api

mrbean-bremen (approved):

PR #810, opened 2021-01-14
Refactor SvgPathBuilder to have Parse  that accepts ReadOnlySpan<char>

wieslawsoltes:
#### Reference Issue

Split from #786

Refactor SvgPathBuilder to have Parse  that accepts ReadOnlySpan<char>

We want to have all converters the same `Parse` method signature (that accepts  `ReadOnlySpan<char>` instead of string for performance reason - less allocations, and unify Parse method API)


wieslawsoltes:
@mrbean-bremen This is ready for review and merge, this is just refactoring to unify parse api

mrbean-bremen (approved):

Source/Paths/SvgPathBuilder.cs
```diff
@@ -322,7 +319,7 @@ private static PointF ToAbsolute(float x, float y, SvgPathSegmentList segments,
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
             if (value is string s)
-                return Parse(s);
+                return Parse(s.AsSpan());
```
inforithmics, line 322: Could it be that converters operate on a span (I don't know the exact architecture of vvv svg, so that's why I'm asking)

```
if (value is string s)
               return Parse(s.AsSpan());
```
change to

```
if (value is string s)
                return Parse(s.AsSpan());
if (value is ReadOnlySpan<char> span)
                return Parse(span);	             
```

wieslawsoltes: I don't know either, if this will be reported as issue we can add string Parse method.

For new frameworks there is implicit conversion available

https://docs.microsoft.com/en-us/dotnet/api/system.string.op_implicit?view=net-5.0

Issue #175, opened 2015-07-28
Stroke width not considered for Boundary


ubbn:
SVG basic elements' width is not considered for boundary. When SVG basic elements' stroke width becomes wider, it looks more apparent and half of stroke width is trimmed in other words full stroke is not rendered fully. The issue is for all basic shapes. Attached image is example for circle, its stroke width is 10. 

![circle](https://cloud.githubusercontent.com/assets/1505531/8929915/622333c0-3529-11e5-93c8-fba3b1fa75d8.png)

And when stroke width is 0, still svg element is rendered as with stroke width 1

ubbn:
Rendered image with this fix is below:

![circle](https://cloud.githubusercontent.com/assets/1505531/8930292/fbdcbbfa-352c-11e5-88d1-2837d2077936.png)

PR #816, opened 2021-01-20
Fix filled polyline not displayed with stroke-width=0

### LINKED ISSUE #785: Invalid bmp conversion rendering
#### Description

Svg file is not being properly rendered when converted to bmp.  Svg file produces desired results in chrome/edge.  The image should have 4 stacked squares.  The top square should be brown and the rest should be blue.  When using vvvv/svg to render as bmp, the top brown square is not included.  The space is transparent instead.  I can see where the brown square should be, but it's like that component is being completely ignored.  The results are the same at any size from 16x16 to 256x256

#### Example data

SVG source file (github wouldnt let me upload an svg, so its a zip):
[Plywood_Logo_SuperSimple.zip](https://github.com/vvvv/SVG/files/5779980/Plywood_Logo_SuperSimple.zip)

Bmp output file:
[256.zip](https://github.com/vvvv/SVG/files/5779982/256.zip)

#### Used Versions
svg 3.1.1
.net 5.0.1
Win10

mrbean-bremen:
- remove unneeded check to account for filled polyline
- add test
- fixes #785

This has been introduced with the fix for #175, I added images with stroke width 5 as a regression test for #175.


mrbean-bremen:
@H1Gdev - please have a look.

H1Gdev:
@mrbean-bremen 
I will check this PR.

Tests/W3CTestSuite/svg/__issue-785-01.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
+  <polyline style="fill: red; stroke:black; stroke-width: 5px;" points="10 20 35 35 60 20 35 10 10 20"/>
+  <polyline style="fill: red; stroke:black; stroke-width: 0px;" points="110 20 135 35 160 20 135 10 110 20"/>
+  <circle style="fill: blue; stroke:black; stroke-width: 5px;" cx="35" cy="70" r="20"/>
+  <circle style="fill: blue; stroke:black; stroke-width: 0px;" cx="135" cy="70" r="20"/>
+<rect style="fill: green; stroke:black; stroke-width: 5px;" x="20" y="105" width="30" height="30"/>
```
H1Gdev, line 7: Please insert indent \<rect\> as well.

Source/Basic Shapes/SvgPolyline.cs
```diff
@@ -15,7 +15,7 @@ public partial class SvgPolyline : SvgPolygon
 
         public override GraphicsPath Path(ISvgRenderer renderer)
         {
-            if ((_Path == null || this.IsPathDirty) && base.StrokeWidth > 0)
```
H1Gdev, line 18: Here also has the same condition.
https://github.com/svg-net/SVG/blob/c4e83167c6adfb8d6770eb109d1a03dd656096c2/Source/Basic%20Shapes/SvgLine.cs#L90

Probably don't need this either.

doc/ReleaseNotes.md
```diff
@@ -3,6 +3,11 @@ The release versions are NuGet releases.
 
 ## Unreleased (master)
 
+### Fixes
+* fixed filled polyline not displayed with stroke-width=0 (see [#785](https://github.com/svg-net/SVG/issues/785)
```
H1Gdev, line 7: append ` and PR #816`.

Tests/W3CTestSuite/svg/__issue-785-01.svg
mrbean-bremen: I actually had tested it, but forgot to add it to the file - good point.

Source/Basic Shapes/SvgPolyline.cs
mrbean-bremen: This is a line that cannot be filled, so this may avoid calling unnecessary code.

doc/ReleaseNotes.md
mrbean-bremen: The issue is linked with the PR anyway - I usually link with the PR only if there is no issue for it. This is the way it is done in other projects, too.

H1Gdev (approved):

Issue #817, opened 2021-01-21
GetXML does not output added child elements


JohnGoldsmith:
Hi,

I'm having trouble getting back the correct XML after adding sub elements to an `SvgDocument`

Given the following svg in a file "Min.svg":

```xml
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:ev="http://www.w3.org/2001/xml-events" xml:space="preserve">
  <g transform="translate(5, 5)">
    <rect width="25" height="50" />
  </g>
</svg>
```

and the following code:

```csharp
void Main()
{
	var path = @"C:\Users\USERNAME\Desktop\Min.svg";
	var svgDoc = SvgDocument.Open(path);
	var grp = svgDoc.Children.OfType<SvgGroup>().FirstOrDefault();
	if (grp != null)
	{
		for (int i = 1; i < 3; i++)
		{
			grp.Children.Add(new SvgCircle
			{
				CenterX = 10 * i,
				CenterY = 10 * i,
				Radius = 50
			});
		}
	}
	svgDoc.GetXML().Dump();
}
```
This (LINQPad) sample outputs the following text (without the SvgCircle elements being displayed):
```xml
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg d1p1:space="preserve" ev="http://www.w3.org/2001/xml-events" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1" xmlns:d1p1="http://www.w3.org/XML/1998/namespace" xmlns="http://www.w3.org/2000/svg">
  <g transform="translate(5, 5)">
    <rect width="100" height="50" />
  </g>
</svg>
```
My main question is where are the added elements?

Other questions are why does the `d1p1` alias get added and, also is it correct that the `xmlns:ev="http://www.w3.org/2001/xml-events"` gets picked up as a custom attribute (SvgDocument.CustomAttributes)?

Many thanks for any insight

John 

[...]

Issue #604, opened 2019-11-02
How to keep the namespaces of elements and attributes from inkscape when saving?


hakakou:
#### Description
Doing this simple test on a file saved by inkscape:

`var doc = SvgDocument.Open<SvgDocument>(@"inkscape_guides_layers.svg");`
`doc.Write(@"inkscape_guides_layers_resaved.svg");`

strips the inkscape namespace of non-svg elements, so the file cannot be reopened in inkscape preserving the guide and layer metainfo.

#### Example data
`  <sodipodi:guide
       id="guide2071"
       orientation="0,-1"
       position="278.19047,240.30358" />`
`<g
     style=""
     id="layer1"
     inkscape:groupmode="layer"
     inkscape:label="Layer A" />`

Becomes:

`<guide id="guide2071" orientation="0,-1" position="278.19047,240.30358" />`
`<g id="layer1" groupmode="layer" label="Layer A" />`

See [test.zip](https://github.com/vvvv/SVG/files/3800997/test.zip) file for the original svg.

I have tried declaring the namespaces manually, but this does not help:

    SvgAttributeAttribute.Namespaces.Add(new KeyValuePair<string, string>(
      "inkscape", "http://www.inkscape.org/namespaces/inkscape"));

    SvgAttributeAttribute.Namespaces.Add(new KeyValuePair<string, string>(
      "sodipodi", "http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"));

#### Used Versions
3.0.49

H1Gdev:
@mrbean-bremen 

`SvgCustomAttributeCollection` has no Namespace prefix.

H1Gdev:
This issue needs to be changed in various functions(Reader, AttributeCollection, Write, ...).

mrbean-bremen:
Reopen - has been accidentally closed via GitHub link.

PR #824, opened 2021-02-03
Use XmlWriter rather than XmlTextWriter

bbsgfalconer:
#### Reference Issue
Implements enhancement #693 

Generalize XML output to use an XmlWriter instead of the more specific XmlTextWriter. 

#### Any other comments?
All unit test pass.


H1Gdev:
@bbsgfalconer 

There is also `XmlTextWriter` here.
https://github.com/svg-net/SVG/blob/ef55e3c2e5436bc37ca328416c2003ce71b020f6/Source/Metadata/SvgDocumentMetadata.cs#L38

Source/SvgExtentions.cs
```diff
@@ -50,10 +50,15 @@ public static string GetXML(this SvgElement elem)
...
             {
                 Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
+
+                var writerSettings = new XmlWriterSettings { Encoding = System.Text.Encoding.UTF8 };
+
                 using (var str = new StringWriter())
-                using (var xml = new XmlTextWriter(str))
                 {
-                    elem.Write(xml);
+                    using (var xml = XmlWriter.Create(str, writerSettings))
```
H1Gdev, line 58: There seems to be no need to separate code block.

bbsgfalconer:
> 
> 
> @bbsgfalconer
> 
> There is also `XmlTextWriter` here.
> 
> https://github.com/svg-net/SVG/blob/ef55e3c2e5436bc37ca328416c2003ce71b020f6/Source/Metadata/SvgDocumentMetadata.cs#L38

That's not an XmlTextWriter. It's a reader.

Source/SvgExtentions.cs
bbsgfalconer: Need to either close the XmlWriter or explicitly call XmlWriter.Flush() before reading str.
Closing the using block is a convenient way to force the writer to flush.

Previous code did not flush XmlTextWriter but seemed to get away with it.

H1Gdev:
@bbsgfalconer 

Oh...sorry.
This PR was only for `Writer`.

Source/SvgExtentions.cs
H1Gdev: Both look equivalent in code.
I think it's less misleading to call the `Flush` method explicitly.

Similarly, in order to complete write process, please call the Flush method in the following location as well.
https://github.com/svg-net/SVG/blob/ef55e3c2e5436bc37ca328416c2003ce71b020f6/Source/SvgElement.cs#L965

H1Gdev:
@bbsgfalconer 

By the way, does this PR solve https://github.com/svg-net/SVG/issues/817#issuecomment-765041138 ?

Source/SvgExtentions.cs
bbsgfalconer: Changed SvgExtentions.cs to explicitly call Flush().

I don't see any need or benefit in changing SvgElement.cs... it will be inefficient to flush each element individually.

bbsgfalconer:
> 
> 
> @bbsgfalconer
> 
> By the way, does this PR solve issue [#817 (comment)](https://github.com/svg-net/SVG/issues/817#issuecomment-765041138) ?

This PR solves the issue with the prefix for xml:space in [#817 (comment)](https://github.com/svg-net/SVG/issues/817#issuecomment-765041138).
However, the xmlns:ev declaration still becomes an ev= attribute. I don't understand why.

When I test the corrected code in #817 with this PR, I get output as below:
```xml
<svg xml:space="preserve" ev="http://www.w3.org/2001/xml-events" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" version="1.1" xmlns="http://www.w3.org/2000/svg">
  <g transform="translate(5, 5)">
    <rect width="25" height="50" />
    <circle cx="10" cy="10" r="50" />
    <circle cx="20" cy="20" r="50" />
  </g>
</svg>
```

Source/SvgExtentions.cs
H1Gdev: > I don't see any need or benefit in changing SvgElement.cs...

In the following case.
```cs
var element = new SvgDescription();
var str = new StringWriter();
var writer = new XmlTextWriter(str);
element.Write(writer);
```

In this case, `XmlTextWriter` can get string immediately after, but `XmlWriter` cannot get string until `Flush` method is called.

> it will be inefficient to flush each element individually.

I think it's inefficient when called from `SvgDocument`.
It seems that need to decide whether to flush.

bbsgfalconer: It should not be the responsibility of `SvgElement` to flush the writer.

It is the calling client's responsibility to ensure the writer is flushed. This happens when the writer is closed, so is not normally a problem. If a different policy is required for some specific use case, then a custom XmlWriter class can be created to implement the desired policy.

H1Gdev: > It should not be the responsibility of SvgElement to flush the writer.

Please make it clear.
If there is a guideline etc. in MS, it is better to put the link in xmldoc.

At the very least, this PR will make the above code incompatible.

H1Gdev:
> However, the xmlns:ev declaration still becomes an ev= attribute. I don't understand why.

This is [known issue](https://github.com/svg-net/SVG/issues/604#issuecomment-551379435).

H1Gdev (changes requested):
Until this change is.
https://github.com/svg-net/SVG/pull/824#discussion_r570623370

Source/SvgExtentions.cs
bbsgfalconer: I don't understand what you are asking for.

I have changed the API of this library to accept an `XmlWriter` where previously it would only accept an `XmlTextWriter`.
However, if you pass in an `XmlTextWriter`, it will behave exactly as before. The only change is that you _may_ now pass another kind of `XmlWriter`.

So:
```csharp
var element = new SvgDescription();
var str = new StringWriter();
var writer = new XmlTextWriter(str);
element.Write(writer);
```
works as before.

The behaviour will only change if you choose to do:
```csharp
var element = new SvgDescription();
var str = new StringWriter();
var writer = XmlWriter.Create(str);    // factory method for abstract XmlWriter
element.Write(writer);
```
in which case you are passing in an abstract `XmlWriter` instead.

H1Gdev: > I have changed the API of this library to accept an `XmlWriter` where previously it would only accept an `XmlTextWriter`.

This PR is so.
But .Net recommends using `XmlWriter`.

https://docs.microsoft.com/dotnet/api/system.xml.xmltextreader
> Starting with the .NET Framework 2.0, we recommend that you use the `XmlWriter` class instead.

Until now, you could only use `XmlTextWriter`.
But This PR will allow to use `XmlWriter`.

Users can take this opportunity to use `XmlWriter`.

It would be nice if `XmlWriter` could get the same result as `XmlTextWriter`, but that would be [inefficient](https://github.com/svg-net/SVG/pull/824#discussion_r570074069), so I thought it would be nice to specify in xmldoc(remarks ??) who would `Flush`.

bbsgfalconer: Added XmlDoc commentary with links to MS docs.

Source/SvgElement.cs
```diff
@@ -956,7 +956,20 @@ private bool TryResolveParentAttributeValue(string attributeKey, out object pare
...
 
-        public virtual void Write(XmlTextWriter writer)
+        /// <summary>
+        /// Write this SvgElement out using a given XmlWriter.
+        /// </summary>
+        /// <param name="writer">The XmlWriter to use.</param>
+        /// <remarks>
+        ///   Recommendation is to create an XmlWriter by calling a factory method,<br/>
+        ///   e.g. <see cref="XmlWriter.Create(System.IO.Stream)"/>,
+        ///   as per <a href="https://docs.microsoft.com/en-gb/dotnet/api/system.xml.xmltextreader?view=net-5.0#remarks">Microsoft documentation</a>.<br/>
```
H1Gdev, line 966: change URL to 
`https://docs.microsoft.com/dotnet/api/system.xml.xmltextreader#remarks`.

Source/SvgElement.cs
```diff
@@ -956,7 +956,20 @@ private bool TryResolveParentAttributeValue(string attributeKey, out object pare
             return resolved;
         }
 
-        public virtual void Write(XmlTextWriter writer)
+        /// <summary>
+        /// Write this SvgElement out using a given XmlWriter.
+        /// </summary>
+        /// <param name="writer">The XmlWriter to use.</param>
+        /// <remarks>
+        ///   Recommendation is to create an XmlWriter by calling a factory method,<br/>
```
H1Gdev, line 964: 1 space. (like \<summary\>.)

H1Gdev (commented):
Thanks !!
Please change only these.

Source/SvgElement.cs
bbsgfalconer: changed

Source/SvgElement.cs
bbsgfalconer: Removed additional spaces.

Source/SvgElement.cs
H1Gdev: Please delete language(`en-gb/`) as well.

bbsgfalconer: done

H1Gdev (approved):
This looks good to me.

Issue #482, opened 2019-06-06
Masks not rendered properly


thibnes:
Using this file as example 
[testSvgMasks.zip](https://github.com/vvvv/SVG/files/3261803/testSvgMasks.zip)

the generated jpeg looks like this 
![output](https://user-images.githubusercontent.com/11884292/59037667-0602b200-8872-11e9-9dd2-a6cf548fdb86.jpg)

Is there a special setup to handle masks?

mrbean-bremen:
I will have a look later, but this is most certainly not supported or a bug.

mrbean-bremen:
Yes, masks are unsupported - the related W3C examples also fail. There is an empty class `SvgMask`, so I guess an implementation has been considered some time...
So I think you are out of luck here, as I doubt that mask implementation is something trivial enouogh to be added soon.

thibnes:
> Yes, masks are unsupported - the related W3C examples also fail. There is an empty class `SvgMask`, so I guess an implementation has been considered some time...
> So I think you are out of luck here, as I doubt that mask implementation is something trivial enough to be added soon.

Alright, thank you for your quick response.

Zakkon:
I found a simple workaround, for anyone interested. To get masks to atleast be written into the .svg files, create a class like this:

`[SvgElement("mask")]
public class SvgMask2 : SvgMask { }`
And use that instead of `SvgMask`. Remember to give it an ID so you can reference it later.
 
And to make a `SvgElement` use the said mask, simply add a custom attribute to it, using the ID you gave the mask:
`element.CustomAttributes.Add("mask", "url(#maskID)");`

This will obviously not render when converting to bitmap, but it won't throw an error either. Useful if all you care about is the .svg file.

mrbean-bremen:
Yes, I also added the attribute to the (empty) SvgMask class, and this fixes at least the writing (though not the functionality, of course). I thought it not worth a separate commit without the mask working, but that was obviously wrong. I will add this tomorrow.

mrbean-bremen:
Thanks for the hint, I committed the change now!

PR #825, opened 2021-02-03
#482 - Masks are now rendered

thenkles:


mrbean-bremen:
If someone wants to review this, please go ahead! I was going to review it for some time now, but didn't get to it due to time constraints, so I would appreciate if someone else could have a look. At a first glance, it looked quite good.

Source/ExtensionMethods/RectangleFExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using System.Drawing;
+using System.Drawing.Drawing2D;
+using System.Text;
+
+namespace Svg.ExtensionMethods
+{
+    internal static class RectangleFExtensions
+    {
+        internal static Rectangle Round(this RectangleF rectangle)
```
H1Gdev, line 11: Why not use [Rectangle.Round](https://docs.microsoft.com/dotnet/api/system.drawing.rectangle.round) ?

H1Gdev (commented):
It does not seem to dispose `Bitmap`.

Source/Svg.csproj
```diff
@@ -144,6 +144,62 @@ Many performance improvements by using SourceGenerators instad of reflection.
   <PropertyGroup Condition="'$(TargetFramework)' != 'net452'">
     <DefineConstants>$(DefineConstants);USE_SOURCE_GENERATORS</DefineConstants>
   </PropertyGroup>
+    
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|netstandard2.0|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
H1Gdev, line 149: Is this property required for each ?

Source/Rendering/SvgRenderer.cs
```diff
@@ -162,5 +178,75 @@ public static ISvgRenderer FromNull()
...
+            _innerGraphics.DrawImageUnscaled(buffer, new Point(renderedBounds.X, renderedBounds.Y));
+            _innerGraphics.Transform = previousTransform;
+        }
+
+        private void ApplyAlphaMask(Bitmap buffer, Bitmap mask, Rectangle renderedBounds)
+        {
+            var bufferData = buffer.LockBits(new Rectangle(0, 0, renderedBounds.Width, renderedBounds.Height), ImageLockMode.ReadWrite, PixelFormat.Format32bppArgb);
+            var maskData = mask.LockBits(renderedBounds, ImageLockMode.ReadOnly, PixelFormat.Format32bppArgb);
+
+            unsafe
```
H1Gdev, line 231: Is there implementation that doesn't use `unsafe` ?

thenkles: There are two I'm aware of - _GetPixel()_/_SetPixel()_ that's really slow or using the _Marshal_ class to copy the bitmap data but it uses a lot of memory.

Source/Svg.csproj
thenkles: I suppose so? That was auto-generated so I'm not 100% sure but please let me know what you think the better approach would be.

Source/ExtensionMethods/RectangleFExtensions.cs
thenkles: Good point, I wasn't aware it existed.

thenkles:
Regarding disposing bitmaps, I'll have a look into it later.

tebjan:
I think we cannot introduce the unsafe flag, many bigger enterprise projects don't allow to reference assemblies with that flag. 

New .NET versions have many new safe memory operations. Probably worth looking at them.

And for bitmap disposal, you should use `using` statements, if possible.

thenkles:
> I think we cannot introduce the unsafe flag, many bigger enterprise projects don't allow to reference assemblies with that flag.
> 
> New .NET versions have many new safe memory operations. Probably worth looking at them.

Ok, that's a fair point. Any details on these new memory operations I could read?

H1Gdev:
How about using [Span](https://docs.microsoft.com/dotnet/api/system.span-1) ?
I don't know if can completely remove `unsafe`.

mrbean-bremen:
@tebjan , @H1Gdev - did you have another look after the latest changes? It's been a while...

H1Gdev:
@mrbean-bremen 

I will check source files, but it will take some time because there are many changes...

Tests/Svg.UnitTests/MaskingTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            float equalPercentage;
+            Bitmap difference;
+
+            ImagesAreEqual(renderedDocument, expectedImage, 1, out equalPercentage, out difference);
+
+            Assert.Greater(equalPercentage, 99);
+        }
+    }
+}
```
H1Gdev, line 79: Add CRLF.

Source/Rendering/SvgRenderer.cs
```diff
@@ -162,5 +179,90 @@ public static ISvgRenderer FromNull()
             var g = CreateGraphics(img);
             return new SvgRenderer(g, img);
         }
+
+        public void SetMask(Bitmap mask)
+        {
+            _mask = mask;
```
H1Gdev, line 185:
```cs
_mask?.Dispose();
_mask = mask;
```

Source/Rendering/SvgRenderer.cs
```diff
@@ -162,5 +179,90 @@ public static ISvgRenderer FromNull()
...
+        }
+
+        public Bitmap GetMask()
+        {
+            return _mask;
+        }
+
+        public void DisposeMask()
```
H1Gdev, line 193: Delete this method.
Use `SetMask(null);`

Source/Rendering/SvgRenderer.cs
H1Gdev: And dispose `_mask` in Dispose method.

Source/Rendering/SvgRenderer.cs
```diff
@@ -35,23 +39,31 @@ public float DpiY
...
+
         /// <summary>
         /// Initializes a new instance of the <see cref="ISvgRenderer"/> class.
         /// </summary>
-        private SvgRenderer(Graphics graphics, bool disposable = true)
+        private SvgRenderer(Graphics graphics, SizeF renderSize, bool disposable = true)
         {
             _innerGraphics = graphics;
             _disposable = disposable;
+            this.RenderSize = renderSize;
```
H1Gdev, line 55: remove 'this'.
Please omit this as much as possible.

H1Gdev (commented):
Please clarify dispose responsibility of mask to be set in `ISvgRenderer.SetMask()`.
Do you use stack for mask bitmaps ?

Source/ExtensionMethods/RectangleFExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Svg.ExtensionMethods
+{
+    internal static class RectangleFExtensions
+    {
+        internal static PointF[] GetPoints(this RectangleF rectangle)
+        {
+            return new[]
+            {
+                rectangle.Location,
+                rectangle.Location + rectangle.Size
```
jasonliuplanetart, line 16: here should condiser all four point like this:
``` c#
            return new[]
            {
                rectangle.Location,
                rectangle.Location + rectangle.Size,
                new PointF(rectangle.X,rectangle.Y+rectangle.Height),
                new PointF(rectangle.X+rectangle.Width,rectangle.Y)
            };
```

liuzhenyulive, line 16: here should consider five points like this, If there are only two points, there will have issue with the bounds when the transform contains rotation
``` c#
 return new[]
            {
                rectangle.Location,
                rectangle.Location + rectangle.Size,
                new PointF(rectangle.X,rectangle.Y+rectangle.Height),
                new PointF(rectangle.X+rectangle.Width,rectangle.Y)
            };
```

liuzhenyulive: you can verify this by this transform : "matrix(0.51 0.51 -0.51 0.51 776.90 96.03)"

paulushub:
@thenkles Sorry for the delay is resolving this PR. Please can you rework and repost this for consideration?

paulushub:
@thenkles Sorry for the delay, please can you update this project for consideration?

Issue #755, opened 2020-07-28
StackOverFlowException


msasso69:
#### Description

With the svg image at the url https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/photos.svg I get a StackOverFlowException

#### Example data

Here is a simple code to reproduce the issue:
```
using System;
using System.IO;
using System.Net;

namespace VisualSEO.EtoFormGui
{
	public class Program
	{
		[STAThread]
		public static void Main(string[] args)
		{
			var url = "https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/photos.svg";
			Uri reqUri = new Uri(url);

			byte[] bytes;

			HttpWebRequest req = WebRequest.Create(reqUri) as HttpWebRequest;
			using (HttpWebResponse webResp = req.GetResponse() as HttpWebResponse)
			{
				byte[] buffer = new byte[32768];
				using (Stream respStream = webResp.GetResponseStream())
				{
					using (MemoryStream ms1 = new MemoryStream())
					{
						int chunkLen = respStream.Read(buffer, 0, buffer.Length);
						ms1.Write(buffer, 0, chunkLen);
						ms1.Position = 0;
						bytes = ms1.ToArray();

						var svg = System.Text.Encoding.Default.GetString(bytes);
						var doc = Svg.SvgDocument.FromSvg<Svg.SvgDocument>(svg);
						var img = doc.Draw();
					}
				}
			}
		}
	}
}
```

#### Used Versions

svg ver.3.3.1, OS: macOS 10.15, mono: 6.10

H1Gdev:
[This SVG file](https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/photos.svg) does not specify `width` and `height` in root \<svg\>.
At this time, `width` and `height` are `100%`.

In browsers it will be entire `canvas` but this library cannot determine canvas size.(Currently, users cannot specify default canvas size.)

If this SVG file is editable, specifying `width` and `height` is easiest solution.

Another issue is that this SVG file is loading images with `JavaScript`.
However, this library does not support script execution with \<script\>.

msasso69:
Thanks @H1Gdev ,
unfortunately I cannot edit the svg file,
in fact I bumped into it because my application is a crawler for SEO site analysis.
IMHO the best would be that if the library does not support something of an svg file it would not crash :D

msasso69:
found another image that is causing the library to crash:

```
[...]

PR #827, opened 2021-02-05
Fix: 'Error occurs with empty SVG' issue.

H1Gdev:
#### Reference Issue

Ref. #755

Fixes issue that occurred error with empty SVG such as:

```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="100%" height="100%">
</svg>
```

Error
```
{"Cannot process SVG file, cannot allocate the required memory"} | Svg.Exceptions.SvgMemoryException
```

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -605,13 +605,16 @@ namespace Svg
         {
             //Trace.TraceInformation("Begin Render");
 
+            var size = Size.Round(GetDimensions());
+            if (size.Width <= 0 || size.Height <= 0)
+                return null;
+
             Bitmap bitmap = null;
             try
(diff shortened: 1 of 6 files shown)


Source/SvgDocument.cs
```diff
@@ -605,13 +605,16 @@ public virtual Bitmap Draw()
         {
             //Trace.TraceInformation("Begin Render");
 
+            var size = Size.Round(GetDimensions());
+            if (size.Width <= 0 || size.Height <= 0)
```
H1Gdev, line 609: Add size check for Bitmap.

mrbean-bremen (approved):
LGTM

PR #832, opened 2021-02-28
SvgViewer outputs SVG file.

H1Gdev:

SvgViewer outputs SVG file.


H1Gdev:
@mrbean-bremen 

I check both image and SVG.
It is inefficient to enable it every time I check.

PR #870, opened 2021-05-06
XXE Mitigation

### LINKED ISSUE #869: Security: vulnerable to XXE attacks
#### Description
The library currently does not protect against [XXE attacks](https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing).
e.g. `SvgDocument.DisableDtdProcessing` defaults to false, which leads to DTDs being accepted in general.
But even when setting this to true, it is still possible to use `xrefs` to reach external resources.

Please:
1. Set default behaviour to be the most secure
2. Provide documentation on the risks and toggles.

#### Example data

```xml
<?xml version="1.0" standalone="yes"?>
<!DOCTYPE foo [ <!ENTITY xxe SYSTEM "file://etc/hosts" > ]>
<svg>....</svg>
```

```xml
<image x="90" y="-65" width="128" height="146"
     xlink:href="https://malicious.net"/>
```

#### Used Versions
latest.

gmwalker:
#### Reference Issue
Fixes #869 

This address the issue of XXE Vulnerability by changing `SvgDtdResolver` so it only resolves external resources if `SvgDocument.ResolveExternalResources` is `true`.  `SvgDocument` will no longer resolve external resources by default, to be on the safe side as suggested in the issue.

#### Any other comments?
In `SvgDtdResolver` I narrowed the check around whether to serve up the embedded svg11.dtd to URIs that look like DTD requests.  Only checking for 'svg' could match a directory named 'svg' containing a different resource altogether - which was a problem I encountered making the unit tests for this pull request.

This is my first contribution to this project and I'm looking forward to collaborating with you all.


mrbean-bremen (approved):
Thank you for the contribution, this looks good! Can you please add an entry in the [release notes](https://github.com/svg-net/SVG/blob/master/doc/ReleaseNotes.md)?

Tests/Svg.UnitTests/XxeVulnerabilityTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            writer.Write(s);
+            writer.Flush();
+            stream.Position = 0;
+            return stream;
+        }
+
+        private const string SvgTextId = "secretText";
+        private const string Secret = "This is a secret!";
+    }
+}
```
H1Gdev, line 103: Please add a newline.

gmwalker: Ok, done

gmwalker:
Added "change" to release notes, as requested.

mrbean-bremen:
Thank you!

Issue #872, opened 2021-05-11
Additional XXE vulnerabilities


kimsey0:
#### Description
Despite the work in #870, there seem to still be a few ways to cause SVG.NET to load external resources, even when `SvgDocument.ResolveExternalResources` is set to `false.`

#### Example data
Via image (same as the second example in #869):
```svg
<svg xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
	<image height="71" width="71" xlink:href="https://www.w3.org/Icons/SVG/svg-logo-v.svg" />
</svg>
```
Via text reference:
```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <tref xlink:href="https://www.w3.org/Icons/SVG/svg-logo-v.svg#nonexistent"/>
</svg>

```

#### Used Versions
Master branch.

mrbean-bremen:
Fixed with #873.

PR #873, opened 2021-05-11
Additional XXE mitigation

kimsey0:
#### Reference Issue
#872 

This prevents loading external images and elements when `SvgDocument.ResolveExternalResources` is `false`.

#### Any other comments?
This is just a draft pull request to get the ball rolling.

1. Should loading external images and text content like this be disabled by default, or should this behavior be behind a separate flag from `ResolveExternalResources`? The property name seems to indicate that it controls the loading of all kinds of external resources, but there might be users who expect to be able to include external images?
2. Should `SvgImage` and `SvgElementIdManager` access the static `SvgDocument.ResolveExternalResources` like this or should it somehow be passed down to them, like for `SvgDtdResolver`?
3. It's missing unit tests. I'll add some, but I'm just trying to figure out what external URI's I can use that will always be available and that will allow distinguishing behaviors from each other. Is it and should it be possible to run the tests without an internet connection?


kimsey0:
@gmwalker: Since this builds on your work, perhaps you have some comments, specifically about the use of the static property?

@H1Gdev, @mrbean-bremen: You were both part of reviewing #870 a few days ago. (Sorry if tagging you like this is unnecessary.)

kimsey0:
It looks like quite a few (190) of the W3C test suite SVG's use external references, specifically for fonts, for example: https://github.com/svg-net/SVG/blob/062dc7c91cadf5607cceca1103f38628c0e065f8/Tests/W3CTestSuite/svg/color-prop-01-b.svg#L40

I have very little information about how SVG.NET is used elsewhere. It might be common and expected to load and render SVG's with external font references. However, in our use-case, we render user-uploaded SVG's, and would like to prevent them from both loading local files and making web requests. That also seems to be the concern of @gmwalker in https://github.com/svg-net/SVG/issues/869#issuecomment-831729398.

It could be that these W3C suite tests should be run with `SvgDocument.ResolveExternalResources = true`, or it could be that the behavior of loading external images and fonts should be controlled by another flag that's enabled by default.

kimsey0:
For now, I enabled `SvgDocument.ResolveExternalResources` in the rendering tests.

Tests/Svg.UnitTests/ImageComparisonTest.cs
```diff
@@ -44,6 +44,9 @@ public void CompareSvgImageWithReference(ImageTestDataSource.TestData testData)
         [TestCaseSource(typeof(ImageTestDataSource), "PassingTests")]
         public void CompareSvgImageWithReference(ImageTestDataSource.TestData testData)
         {
+            // W3C Test Suites use external references to local fonts
+            SvgDocument.ResolveExternalResources = true;
```
H1Gdev, line 48: It seems better to review the settings here as well.

https://github.com/svg-net/SVG/blob/062dc7c91cadf5607cceca1103f38628c0e065f8/Tests/Svg.UnitTests/ImageComparisonTest.cs#L94

kimsey0: Well spotted. I've removed the assignment in `CompareSvgImageWithReference` and moved the assignment in `CompareSvgImageWithReferenceImpl` to the top of the method instead.

mrbean-bremen (commented):
FWIW, the changes look good to me, but as I wrote -- I have no experience here and probably wouldn't find any missing checks, so I'll wait a bit for someone else to have a look.
I also think that having `ResolveExternalSources` default to `false` is the correct way, we just have to make sure that this is made clear in the release notes. As this is a breaking change, we'll probably need a new major release version, if we want to follow semantic versioning -- @tebjan , what do you think?

kimsey0:
I've added image comparison tests that render an SVG with an external image reference and one with a text reference, and check that the image and text is only shown if `ResolveExternalSources` is `true`. I put the tests in `XxeVulnerabilityTests`, but made the `CompareSvgImageWithReferenceImpl` from `ImageComparisonTest` public and static and used it. Let me know if you'd rather have the tests moved to `ImageComparisonTest`, have the rendering and comparison logic duplicated in `XxeVulnerabilityTests`, or think this could be tested better with a different kind of test.

> I also think that having ResolveExternalSources default to false is the correct way, we just have to make sure that this is made clear in the release notes. As this is a breaking change, we'll probably need a new major release version, if we want to follow semantic versioning

I agree. If loading external images and other references is disabled by default, it's a breaking change and should be a new major version.

Tests/Svg.UnitTests/ImageComparisonTest.cs
```diff
@@ -28,6 +28,9 @@ public class ImageComparisonTest
         [TestCaseSource(typeof(ImageTestDataSource), nameof(ImageTestDataSource.PassingTests))]
         public void CompareSvgImageWithReference(ImageTestDataSource.TestData testData)
         {
+            // W3C Test Suites use external references to local fonts
+            SvgDocument.ResolveExternalResources = true;
```
H1Gdev, line 32: Is it possible to move this line to head of `CompareSvgImageWithReferenceImpl` method ?

kimsey0: I did so in an earlier commit, but moved it back here because I now also use `CompareSvgImageWithReferenceImpl` from `XxeVulnerabilityTests`, which need `XxeVulnerabilityTests` to not be overwritten. I've added a few suggestions for alternative organization of the test here https://github.com/svg-net/SVG/pull/873#issuecomment-841139774.

I can also make the value for `ResolveExternalResources` a parameter of `CompareSvgImageWithReferenceImpl` and pass it in from `CompareSvgImageWithReference` and `XxeResolveExternalImage`? However you'd prefer.

H1Gdev: In `XxeVulnerabilityTests`, please evaluate DOM value etc. instead of comparing the images.

Now in `ImageComparisonTest`, I think that I do is only in case of `SvgDocument.ResolveExternalResources = true`.

kimsey0: For both `<image xlink:href="..." />` and `<tref xlink:href="..."/>`, resolving the external reference only seems to happen during rendering, and even after calling `Draw`, I can't see anything that has changed in the DOM which a test could look for. Am I overlooking something?

H1Gdev: Please add property or method to evaluate.

H1Gdev (commented):
I think setting this feature with `Boolean` value will significantly reduce usability.
In most cases, we want to disable `external file reference`, but we want to display images.

What do you want to disable in this PR ?

kimsey0:
@H1Gdev: I'm not sure I quite understand your questions.

> I think setting this feature with Boolean value will significantly reduce usability.

Do you mean to suggest having multiple different Boolean properties controlling the different types of external resources that can be loaded:

``` csharp
public bool ResolveExternalXmlEntites { get; set }
public bool ResolveExternalImages { get; set; }
public bool ResolveExternalElements { get; set }
```

Or something similar with an enum or flags enum?

> In most cases, we want to disable external file reference, but we want to display images.

The current implementation in this pull request allows displaying images via `data:` URI's, but neither via local file system references nor web references:

``` svg
<!-- Displayed when SvgDocument.ResolveExternalResources == false -->
<image href="data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7" />

<!-- Not displayed when SvgDocument.ResolveExternalResources == false -->
<image href="../../MySecretImages/Secret.png" />
<image href="https://example.org/External.png" />
```

Are you saying that, in most cases, we want to allow loading external images from the local file system and the web?

> What do you want to disable in this PR ?

Our use-case is to allow users to upload SVG files and to generate previews of them. When rendering these previews, we want to:
1. Disable loading files from the local file system, since that could allow including confidential information from the local system in the previews.
2. Disable loading files from the web, since that exposes us to Server Side Request Forgery, where a user has our server make malicious web requests.

> Please add property or method to evaluate.

I'm not sure where I should add a property or method to evaluate what?

tebjan:
@mrbean-bremen sorry for the late reply. Yes, I agree with what you said about the versioning, since it's a breaking change. 

Regarding code, it's out of my comfort zone a bit. But from what I've seen, it's definitely an improvement, even if it isn't mitigating all possible vulnerabilities.

mrbean-bremen:
@tebjan - thanks! I also think that the changes are good, and we may just wait a bit for any other comments and ideas before merging them in. 

@H1Gdev - maybe you can show in a code example what you mean? I also didn't quite get your review comments.

H1Gdev:
@kimsey0 

> Or something similar with an enum or flags enum?

I think `flags enum` is easy to use.

However, we haven't been able to identify use cases, so I'm wondering what kind of interfaces should be defined.

- design 1
  ```cs
  [Flags]
  enum ExternalType
  {
    None = 0x0,
    Local = 0x1,
    Remote = 0x2,
  }
  public ExternalType ResolveExternalXmlEntites { get; set } = ExternalType.None;
  public ExternalType ResolveExternalImages { get; set; } = ExternalType.Local;
  public ExternalType ResolveExternalElements { get; set } = ExternalType.Local | ExternalType.Remote;
  ```
- design 2
  ```cs
  [Flags]
  enum ExternalType
  {
    Local_XML = 0x1,
    Remote_XML = 0x2,
    Local_Image = 0x4,
    Remote_Image = 0x8,
    Local_Element = 0x10,
    Remote_Element = 0x20,
  }
  public ExternalType ResolveExternalResources { get; set } = ExternalType.Local_Image | ExternalType.Remote_Image;
  ```

> Are you saying that, in most cases, we want to allow loading external images from the local file system and the web?

This depends on intended use of users.
(I mainly use it for SVG image output.)

> Our use-case is to allow users to upload SVG files and to generate previews of them. When rendering these previews, we want to:

It seems that default values are different on server side or client side...
It may be kind to define default values in each case.

> I'm not sure where I should add a property or method to evaluate what ?

Please add it to corresponding element.
(For example `SvgImage`)
```cs
public ExternalType ResolveExternalResources { get; set }
```

```cs
                if (ResolveExternalResources.HasFlag(ExternalType.Local))
                {
                }
                if (ResolveExternalResources.HasFlag(ExternalType.Remote))
                {
                }
```

kimsey0:
@H1Gdev 

> I think flags enum is easy to use.

OK. I went with your design 1, since I think it makes both setup and usage of the properties more clear and makes it harder to accidentally overwrite flags for other resource types. Have a look!

> It seems that default values are different on server side or client side...
> It may be kind to define default values in each case.

I'm not exactly sure how this should be accomplished. We can't reliably determine which environment SVG.NET is being used in, so it would have to be configurable. We could add additional properties or configuration methods like (shorthand):

```csharp
public static void ConfigureServerSecurity() => ResolveExternalXmlEntites = ResolveExternalImages = ResolveExternalElements = ExternalType.None;
```

But I'm not sure that would be any more discoverable or understandable than just setting the properties manually.

I've set `ResolveExternalXmlEntites` to `ExternalType.None` and `ResolveExternalImages` and `ResolveExternalElements` to `ExternalType.Local | ExternalType.Remote` for now. That keeps the current behavior on `master`, but the defaults changes from #870 may still be considered breaking and necessitate a new major version.

> Please add it to corresponding element.
> (For example SvgImage)
I can add this property to `SvgImage`, but I'm not sure how to get it initialized. `SvgImage` seems to be instantiated by `SvgElementFactory`, which is source generated by `AvailableElementsGenerator`. I can't find any functionality to initialize this kind of property which isn't bound to anything in the SVG file. Can you point me in the right direction?

Tests/Svg.UnitTests/XxeVulnerabilityTests.cs
```diff
@@ -39,6 +41,35 @@ public void XxeResolveExternalResources(bool resolveExternalResources)
                 DeleteSecretFile(secretFilePath);
             }
         }
+        
+        [TestCase(ExternalType.None, "__pull_request-873-01", "__pull_request-873-01-ExternalReferencesDisabled")]
```
H1Gdev, line 45: Include this test in `ImageComparisonTest`.

https://github.com/svg-net/SVG/pull/873#discussion_r632527485

Tests/Svg.UnitTests/ImageComparisonTest.cs
```diff
@@ -62,8 +72,10 @@ public void CompareSvgImageWithReference(ImageTestDataSource.TestData testData)
             CompareSvgImageWithReferenceImpl(baseName, svgPath, pngPath, testSaveLoad);
         }
 #endif
+        
+        
 
-        private void CompareSvgImageWithReferenceImpl(string baseName,
+        public static void CompareSvgImageWithReferenceImpl(string baseName,
```
H1Gdev, line 78: Please revert this line.

H1Gdev:
@kimsey0 

>> Please add it to corresponding element.
>> (For example SvgImage)

> I can add this property to SvgImage, but I'm not sure how to get it initialized. SvgImage seems to be instantiated by SvgElementFactory, which is source generated by AvailableElementsGenerator. I can't find any functionality to initialize this kind of property which isn't bound to anything in the SVG file. Can you point me in the right direction?

At this time, there is no need to initialize.

A place that has access to direct `SvgDocument` from `SvgImage`, ensure that you access through the property or method of `SvgImage`.

It may be better to have it in `SvgElement`.

H1Gdev:
> But I'm not sure that would be any more discoverable or understandable than just setting the properties manually.

If we want to make it stronger, you can force users to set it.
If not set, an exception will be thrown and it will not work.

mrbean-bremen:
Sorry for the long delay - I missed the question, and I didn't follow up here, as I was busy with other stuff. 
I just had another look, and I actually think this is sufficient as is. It does the main task to make the library usage safer by default, and allows to configure it easily.
I will wait a couple of days for further comments and merge if nothing substantial comes up. It also probably makes sense to make a new release soon.

H1Gdev:
It looks like some code hasn't been fixed yet.

mrbean-bremen:
> It looks like some code hasn't been fixed yet.

You refer to the private method made public, and the reference to the test? Yes, these could be changed, though these are minor things. Anything else?

H1Gdev:
Maybe that's it.

Tests/Svg.UnitTests/XxeVulnerabilityTests.cs
mrbean-bremen: There is no need for this, as the image comparison tests are already executed here.

Tests/Svg.UnitTests/ImageComparisonTest.cs
mrbean-bremen: This is used in the other test, so it cannot be private. Can be made internal instead.

mrbean-bremen (approved):

Tests/Svg.UnitTests/XxeVulnerabilityTests.cs
H1Gdev: I commented that DOM test should be done here because images comparison test is uncertain...

H1Gdev: If it looks like it will be in time for this release, I will create a PR.

mrbean-bremen: Go ahead 👍

MarvinPogoda:
Hey @mrbean-bremen ,
is it possible to mark the nuget-versions before this fix as vulnerable, so using the older version will know that there might be security issues?

PR #875, opened 2021-05-17
Manage namespaces and prefixes.

### LINKED ISSUE #604: How to keep the namespaces of elements and attributes from inkscape when saving?

H1Gdev:
#### Reference Issue

Fixes #604 and #817

- Manage namespaces and prefixes in `SvgElement`.
- `SvgCustomAttributeCollection` has namespace.

### DIFF
--- a/Source/SvgNamespace.cs
+++ b/Source/SvgNamespace.cs
@@ -1,9 +1,0 @@
-﻿
-
-namespace Svg
-{
-    public static class SvgNamespace
-    {
-        public const string UriString = "http://www.w3.org/2000/svg";
-    }
-}
--- a/Generators/AvailableElementsGenerator.cs
+++ b/Generators/AvailableElementsGenerator.cs
@@ -255,7 +255,7 @@ namespace Svg
                     }
                 }
 
-                if (string.Empty.Equals(elementName))
+                if (elementName?.Length == 0)
                     continue;
 
                 if (elementName is not null && items.TryGetValue(elementName, out var element))
--- a/Source/Document Structure/SvgFragment.cs
+++ b/Source/Document Structure/SvgFragment.cs
@@ -16,7 +16,7 @@ namespace Svg
         /// <summary>
         /// Gets the SVG namespace string.
         /// </summary>
-        public static readonly Uri Namespace = new Uri(SvgNamespace.UriString);
+        public static readonly Uri Namespace = new Uri(SvgNamespaces.SvgNamespace);
 
         PointF ISvgBoundable.Location
         {
--- a/Source/SvgAttributeAttribute.cs
+++ b/Source/SvgAttributeAttribute.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 
 namespace Svg
 {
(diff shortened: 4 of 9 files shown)


H1Gdev:
@mrbean-bremen 

Will this be merged ?

mrbean-bremen:
> Will this be merged ?

I missed that, apparently - will have a look tonight. Is this a breaking change (at a first glance, there seems to be a namespace change)?

H1Gdev:
I change to parse XML namespaces.

> Is this a breaking change ?

This PR considers compatibility.

Source/SvgNamespace.cs
```diff
@@ -1,9 +0,0 @@
-﻿
-
-namespace Svg
-{
-    public static class SvgNamespace
-    {
-        public const string UriString = "http://www.w3.org/2000/svg";
```
mrbean-bremen, line 7: Removing this changes the public API. I have no idea if this is used somewhere, but if it is, it will break the code. I'm a bit reluctant to introduce breaking changes for a dot release. 
On the other hand, the XXE mitigation also changes the behavior and will break existing applications, so maybe we should at least change the minor version for the next release. Strictly speaking, if following semantic versioning, each such change would mean a new major version, but I'm not sure about that... @tebjan - what do you think?

Regardless, any change in the public API also needs to be described in the release notes under `Changes`.

H1Gdev: In PR #760, public API `SvgAttributeAttribute.SvgNamespace` was removed when `SvgNamespace` class was created.
It's possible to leave `SvgNamespace` class, but I've removed it because having this class makes code dirty.

mrbean-bremen: Thanks, you are right - I wasn't aware that this has been added only after the last release!

mrbean-bremen (approved):

PR #890, opened 2021-09-01
Make SvgExtentions.Traverse/TraverseDepthFirst internal

### LINKED ISSUE #889: Traverse<T>() and TraverseDepthFirst<T>() show as extension methods on everything
I use the library as a small part of a project I work on, for processing SVG images downloaded from the internet, and it has been bothering me for awhile that these two methods show up as extensions methods for *everything,* including things that have absolutely nothing to do with this library.

Would it be possible to add a constraint on these methods or otherwise make them internal? Having a no-constraint generic extension method on a public API for a project dedicated to a specific line of work feels like a potential problem and at the very least, an eyesore for consumers.

This would be a breaking change to do either, and I have no clue how many people use these extensions outside of their internal use by the library, but it would be greatly appreciated.

I know this is totally splitting hairs for some, but for others like me it's a genuine annoyance when trying to work with other objects and seeing these extension methods on everything.

If absolutely desired, one could simply slap `[EditorBrowsable(EditorBrowsableState.Never)]` on them to not introduce a breaking change, but some IDEs like Rider still ignore this for some reason.

mrbean-bremen:
- avoids polluting the external API
- fixes #889

### DIFF
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -4,24 +4,28 @@ The release versions are NuGet releases.
 ## Unreleased (master)
 
 ### Changes
-* change namespace of SvgSymbol from Svg.Document_Structure to Svg. (see [PR #556](https://github.com/svg-net/SVG/pull/556))
-* changed default behavior of DTD resolution so external references are not resolved by default; to mitigate XXE vulnerability. (see [PR #870](https://github.com/svg-net/SVG/pull/870))
-* changed default behavior so external references to images, text definitions, and other resources are not resolved by default; to improve safety of rendering untrusted files. (see [PR #873](https://github.com/svg-net/SVG/pull/873))
-* fixed ISvgRenderer.SmoothingMode reset in method SvgVisualElement.RenderFillAndStroke. (see [PR #883](https://github.com/svg-net/SVG/pull/883))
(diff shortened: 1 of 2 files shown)


doc/ReleaseNotes.md
```diff
@@ -4,24 +4,28 @@ The release versions are NuGet releases.
...
+* changed namespace of `SvgSymbol` from `Svg.Document_Structure` to `Svg` (see [PR #556](https://github.com/svg-net/SVG/pull/556))
+* mitigated XXE vulnerability: changed default behavior of DTD resolution so external references are not resolved by default
+  (see [#869](https://github.com/svg-net/SVG/issues/869))
+* improved safety of rendering untrusted files: changed default behavior so external references to images, text definitions, 
+  and other resources are not resolved by default (see [#872](https://github.com/svg-net/SVG/issues/872))
+* made `SvgExtentions.Traverse` and `SvgExtentions.TraverseDepthFirst` internal to avoid polluting the API
+  (see [#889](https://github.com/svg-net/SVG/issues/889))
 
 ### Enhancements
-* minimize XmlTextReader customization (see [PR #836](https://github.com/svg-net/SVG/pull/836))
```
H1Gdev, line 13: This is `pull`.

PR #897, opened 2021-09-20
Expose Open(XmlReader) method.

### LINKED ISSUE #877: Is there a reason SvgDocument.Open(XmlReader) is not exposed?
`SvgDocument.Open` only accepts a `string`, `Stream` or `XmlDocument`, but there is no public overload for reading out of a generic XML, which should be done via `XmlReader`. It seems the overload is there, but it is not public. Is there a reason for it, or could it be exposed?

This would make reading SVG files from other sources much easier. The data could be produced from a text stream or from a possible serialization of the XML infoset other than the standard one, from `XDocument`, `XElement` (and `XmlElement` as well) or `XPathNavigator`, all without having to store the whole document in memory at once. Additionally a custom entity resolver could be used as well.

H1Gdev:
#### Reference Issue

Fixes #877

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -390,7 +390,7 @@ namespace Svg
                 WhitespaceHandling = WhitespaceHandling.Significant,
                 DtdProcessing = DisableDtdProcessing ? DtdProcessing.Ignore : DtdProcessing.Parse,
             };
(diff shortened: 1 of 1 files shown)


mrbean-bremen (approved):
Thank you again!

PR #905, opened 2021-10-07
Added  tests concerning svg's with xml:space="preserve"

hakakou:
Added 2 tests concerning svg's with xml:space="preserve" enabled. 
Also added test for adding SvgText with null Text.

#### Reference Issue
Based on #817 

#### Any other comments?
Added 3 tests. 1 passes and 2 fails.


paulushub:
The following W3C Tests already provide sample files:
- text-ws-01-t.svg (Test for viewer correct handling of whitespace and the 'xml:space' attribute. )
- text-ws-02-t.svg
- text-ws-03-t.svg

Also, available are
- text-align-04-b.svg
- text-deco-01-b.svg
- text-tspan-02-b.svg

PR #911, opened 2021-10-12
Fix: #909 'Very high memory usage (and allocations)' issue.

### LINKED ISSUE #909: Very high memory usage (and allocations)
Due to precalculating Content property of SvgDocument (and all nodes below) there is multiple string concatenation which cause multiple GC

https://github.com/svg-net/SVG/blob/master/Source/SvgDocument.cs#L504

Processing of one file (5MB)

image

H1Gdev:
#### Reference Issue

Fixes #909

Ref. #89, #90, #93

### DIFF
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -3,7 +3,7 @@ The release versions are NuGet releases.
 
 ## Unreleased (master)
 
-### Performances
+### Performance
 * reduce very high memory usage (see [#909](https://github.com/svg-net/SVG/issues/909))
 
 ### Fixes


Materix:
Give me a sec. I have some hard time in job

Materix:
I am really curious why there is no allocation at all for returned new string.
Internally `Concat` with array param make some magic with char arrays

Materix:
@H1Gdev @AraHaan okey, there is correct results.  Base is with string `+` operator and `Aggregate` to comparison. As you can see using `ToArray`  make almost 40% profit with only 10% cost in memory term. Difference between `Join` and `Concat` is really small but I think that using `Concat` will be clearer

PR #920, opened 2021-10-30
Fix: #917 'SVG to PNG renders inline images incorrectly' issue.

### LINKED ISSUE #917: SVG to PNG renders inline images incorrectly
#### Description
I'm trying to convert an online Meteogram from SVG to PNG. It looks like it doesn't scale the inline (base64) images down correctly.

#### Example data
Url: https://developer.yr.no/doc/assets/meteogram.svg

Expected:
image

Result:
image

Example code:
```csharp
using Svg;
using System.Drawing;
using System.Drawing.Imaging;
using System.IO;
using System.Net.Http;
using System.Threading.Tasks;

namespace SvgConvert
{
    public class Program
    {
        public static async Task Main()
        {
            HttpClient httpClient = new HttpClient();
            Stream httpSvgStream = await httpClient.GetStreamAsync("https://developer.yr.no/doc/assets/meteogram.svg");
            SvgDocument svgDocument = SvgDocument.Open<SvgDocument>(httpSvgStream);
            Bitmap bitmap = svgDocument.Draw();
            bitmap.Save("Meteogram.png", ImageFormat.Png);
        }
    }
}
```

Maybe the issue is only scaling and it could be an easy fix. Or there is more info in the <style> tags which isn't parsed/respected.

#### Used Versions
Windows 10 Pro 21H1 19043.1288
Svg 3.3.0
.NET 6.0 (Also tried .NET Core 3.1 and .NET 5.0)

H1Gdev:
#### Reference Issue

Fixes #917


Source/Document Structure/SvgFragment.cs
```diff
@@ -171,9 +171,9 @@ protected override void Render(ISvgRenderer renderer)
                     var prevClip = renderer.GetClip();
                     try
                     {
-                        var size = this.Parent == null ? renderer.GetBoundable().Bounds.Size : GetDimensions();
-                        var clip = new RectangleF(this.X.ToDeviceValue(renderer, UnitRenderingType.Horizontal, this),
-                                                  this.Y.ToDeviceValue(renderer, UnitRenderingType.Vertical, this),
+                        var size = this is SvgDocument ? renderer.GetBoundable().Bounds.Size : GetDimensions(renderer);
```
mrbean-bremen, line 174: Is there a difference between `Parent == null` and `this is SvgDocument`?

mrbean-bremen (approved):
Looks good to me, thanks!

Source/Document Structure/SvgFragment.cs
H1Gdev: Outermost `<svg>` may also have `Parent`.(`Parent` is required to apply styles.)
So I changed condition.

mrbean-bremen: Ah, thanks - didn't see that!

PR #925, opened 2021-12-02
Support relative commands in Paths

H1Gdev:

- Save `Paths` as relative commands.
- Corresponds to dynamic changes in code.


Source/Paths/SvgArcSegment.cs
```diff
@@ -43,24 +42,26 @@ private static double CalculateVectorAngle(double ux, double uy, double vx, doub
             return DoublePI - (ta - tb);
         }
 
-        public override void AddToPath(GraphicsPath graphicsPath)
+        public override PointF AddToPath(GraphicsPath graphicsPath, PointF start, SvgPathSegmentList parent)
```
mrbean-bremen, line 45: same here

Source/Paths/SvgClosePathSegment.cs
```diff
@@ -1,31 +1,32 @@
+﻿using System.Drawing;
 using System.Drawing.Drawing2D;
 
 namespace Svg.Pathing
 {
     public sealed class SvgClosePathSegment : SvgPathSegment
     {
-        public override void AddToPath(GraphicsPath graphicsPath)
+        public SvgClosePathSegment(bool isRelative)
```
mrbean-bremen, line 8: Same here - make it default to true.

mrbean-bremen (commented):
As far as I can see, this all looks good, and is a helpful addition, You also added a lot of tests. The only thing is that we don't want to change the public API without announcing it first - as I wrote, add the previous API back and mark it as obsolete, so we can remove it in a later version.

H1Gdev:
I understand idea of wanting to keep previous API  for compatibility.
But that doesn't work perfectly correctly.(I didn't want to change public API if I could...)
Is that okay ??

Source/Paths/SvgClosePathSegment.cs
```diff
@@ -1,31 +1,43 @@
...
 
         public override string ToString()
         {
-            return "z";
+            return IsRelative ? "z" : "Z";
+        }
+
+        [System.Obsolete("Use new constructor.")]
+        public SvgClosePathSegment()
+            : this(true)
```
H1Gdev, line 34: We don't know why this command is only relative path...

mrbean-bremen (approved):
Thanks!

PR #931, opened 2022-01-02
Enable building code without System.Drawing.Common dependency

wieslawsoltes:

Enabled building library without System.Drawing.Common dependency by defining `NO_SDC`.

### DIFF
--- a/Source/Paths/SvgQuadraticCurveSegment.cs
+++ b/Source/Paths/SvgQuadraticCurveSegment.cs
@@ -1,5 +1,7 @@
 ﻿using System.Drawing;
+#if !NO_SDC
 using System.Drawing.Drawing2D;
+#endif
 
 namespace Svg.Pathing
 {
@@ -34,6 +36,7 @@ namespace Svg.Pathing
             return new PointF(x2, y2);
         }
 
+#if !NO_SDC
         private static PointF CalculateControlPoint(PointF start, PointF firstControlPoint)
         {
             var x1 = (firstControlPoint.X * 3 - start.X) / 2;
@@ -66,6 +69,13 @@ namespace Svg.Pathing
             return end;
         }
 
+        [System.Obsolete("Use new AddToPath.")]
+        public override void AddToPath(GraphicsPath graphicsPath)
+        {
+            AddToPath(graphicsPath, Start, null);
+        }
+#endif
+
         public override string ToString()
         {
             if (float.IsNaN(ControlPoint.X) || float.IsNaN(ControlPoint.Y))
@@ -79,11 +89,6 @@ namespace Svg.Pathing
             : this(false, controlPoint, end)
         {
             Start = start;
-        }
-        [System.Obsolete("Use new AddToPath.")]
-        public override void AddToPath(GraphicsPath graphicsPath)
-        {
-            AddToPath(graphicsPath, Start, null);
         }
     }
 }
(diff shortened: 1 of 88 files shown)


H1Gdev:
I don't know how to surround `NO_SDC`.
Are there any rules ?

wieslawsoltes:
If you remove`System.Drawing.Common` package reference anything that does not compile. All the code that uses types  from System.Drawing.Common are surrounded by NO_SDC if/def.

wieslawsoltes:
> If you remove`System.Drawing.Common` package reference anything that does not compile. All the code that uses types from System.Drawing.Common are surrounded by NO_SDC if/def.

In Svg.Skia I substituted the System.Drawing.Common with SkiaSharp alternative.

mrbean-bremen (approved):

mrbean-bremen:
@wieslawsoltes - I have rebased this with master, please check if all still works with Svg.Skia and I haven't messed something up while resolving conflicts. Will merge if you give the ok.

wieslawsoltes:
> @wieslawsoltes - I have rebased this with master, please check if all still works with Svg.Skia and I haven't messed something up while resolving conflicts. Will merge if you give the ok.

@mrbean-bremen Works fine with Svg.Skia. Thanks!

H1Gdev:
@wieslawsoltes 

The questions are listed below.
- Where is `NO_SDC` defined ?
- How does SVG behave when `NO_SDC` is defined?

wieslawsoltes:
> @wieslawsoltes
> 
> The questions are listed below.
> 
> * Where is `NO_SDC` defined ?
> * How does SVG behave when `NO_SDC` is defined?

In my Svg.Custom project https://github.com/wieslawsoltes/Svg.Skia/blob/407a032aeaaf9a3716959e52e63aca6953f50c6d/src/Svg.Custom/Svg.Custom.csproj#L14

When it's defined I replace all removed functionality with my code.

H1Gdev:
@wieslawsoltes 

Thanks !

@mrbean-bremen 

This PR depends on `Svg.Skia`.
How about creating another branch ?

mrbean-bremen:
@H1Gdev - it is the other way around, `Svg.Skia` depends on this. It is not ideal, but I see this as a first step to factor out the drawing functionality as mentioned in #590. Anyway we need a way to handle the dropped support for `Common.Drawing`, and `Svg.Skia` already does a lot here. I would like to work both repos to be able to work together more seemlessly, and this change might be a step on this way.

Source/Paths/SvgQuadraticCurveSegment.cs
```diff
@@ -34,6 +36,7 @@ private static PointF CalculateSecondControlPoint(PointF controlPoint, PointF en
             return new PointF(x2, y2);
         }
 
+#if !NO_SDC
```
H1Gdev, line 39: @wieslawsoltes 

I think it surrounds `AddToPath`.

mrbean-bremen: It surrounds `AddToPath` and the helper function `CalculateControlPoint` used in `AddToPath`. Is that wrong?

wieslawsoltes: I think I will submit PR with splitting of all drawing code inside NO_SDC if/def into separate files. This should be easier to read and maintain also make next step closer for https://github.com/svg-net/SVG/issues/590

wieslawsoltes: Started work here https://github.com/svg-net/SVG/pull/934 @H1Gdev @mrbean-bremen your input would be appreciated

mrbean-bremen: That sounds good, thanks! I will have a look later today.

PR #935, opened 2022-01-10
HttpMaster added as a tool using the library

joxi:


mrbean-bremen (approved):
Thanks!
@tebjan - are there any rules for including links here?

PR #940, opened 2022-01-12
Take into account of parent transforms

Happypig375:


Happypig375:
@mrbean-bremen Re-run?

H1Gdev:
What kind of `SVG` will this PR improve ?

Happypig375:
@H1Gdev 
```svg
<svg viewBox="0 0 100 100">
  <text transform="translate(10 30)">
    <tspan id="target">ABC</tspan>
  </text>
</svg>
```
Bounds of `tspan` from view box should match `text`

H1Gdev:
Thanks.

What happens with this SVG ?
```svg
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="100" height="100">
  <text>
    <tspan transform="translate(10 30)">ABC</tspan>
  </text>
</svg>
```

Many major browsers don't show anything for this SVG. (`transform` in \<tspan\> seems to be ignored.)

Happypig375:
<tspan> doesn't support transforms by itself, correct. But its bounds should use its parent transforms to result in same bounds for the <text> and <tspan> elements since they do represent the same text at the same location.

H1Gdev:
It doesn't seem to work properly with latest master, can it be fixed by merging this PR ?

Happypig375:
Yes, this is what I aim to fix.

H1Gdev:
It doesn't seem to be fixed...
This SVG should not show anything because \<tspan\> `transform` is invalid.

Happypig375:
In this case, it's the problem of the parser, and Bounds just takes information from the object tree. Out of scope.

Happypig375:
```svg
<svg viewBox="0 0 100 100">
  <text transform="translate(10 30)">
    <tspan id="target">ABC</tspan>
  </text>
</svg>
```
This PR fixes this, and this only.

Happypig375:
Hmm, actually this issue may exist across different element types, not just \<tspan>.

Happypig375:
Superceded by #945

PR #942, opened 2022-01-13
Fix: #941 'Multiple TextDecoration flags result in commas when serializing to .svg' issue.

### LINKED ISSUE #941: Multiple TextDecoration flags result in commas when serializing to .svg
#### Description
When using multiple Text Decorations, the serialized .svg file has commas in between each text decoration flag. This causes renderers (the .svg file as well as .bmp) to not render the text decorations.

#### Example data

```C#
var svg = new SvgDocument
{
    Width = 300,
    Height = 100,
    ViewBox = new SvgViewBox(0, 0, 300, 100),
    Children =
    {
        new SvgText("Hello World!")
        {
            Y = {50},
            FontSize = 50,
            TextDecoration = SvgTextDecoration.LineThrough | SvgTextDecoration.Underline |
                             SvgTextDecoration.Overline
        }
    }
};

using var bitmap = svg.Draw();
bitmap.Save("card.bmp", ImageFormat.Bmp);
svg.Write("card.svg");
```

Resulting SVG:

```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" width="300" height="100" viewBox="0, 0, 300, 100">
  <text y="50" font-size="50" text-decoration="underline, overline, line-through">Hello World!</text>
</svg>
```

`text-decoration="underline, overline, line-through"` should read `text-decoration="underline overline line-through"`

#### Used Versions
NuGet version 3.4.0 (released 3 days ago), Windows 10, Target Framework .NET 5.0

H1Gdev:
#### Reference Issue

Fixes #941


Source/DataTypes/EnumConverters.cs
```diff
@@ -126,6 +126,24 @@ public sealed class SvgGradientSpreadMethodConverter : EnumBaseConverter<SvgGrad
...
+
+            return base.ConvertFrom(context, culture, value);
+        }
+
+        public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
+        {
+            var destination = base.ConvertTo(context, culture, value, destinationType);
+
+            if (destination is string && value is SvgTextDecoration)
+                destination = ((string)destination).Replace(",", string.Empty);
```
mrbean-bremen, line 143: Are we sure that always have an additional space here? Maybe it would be more to secure to replace it with a space?

mrbean-bremen (approved):
Looks good, thanks! Will you resolve the conflict yourself?

Source/DataTypes/EnumConverters.cs
H1Gdev: `Enum` separator(", ") hasn't changed since long ago, so I think it's okay.
Simple replacement will result in duplicate spaces. However, I think regular expressions are expensive.
Simple replacement will result in duplicate spaces. However, I think using RegExp is cost to performance.

H1Gdev:
@mrbean-bremen 

Would you like to branch out(`feature/drawing`, etc.) commits of drawing separation from master once ?
If such another commit like this is mixed in between, the history will be difficult to understand.

H1Gdev:
@mrbean-bremen 

I'm not saying that you should operate it on `GitHub Flow`, but the current master is very halfway...
Is there anything wrong with keeping `master` branch in good shape ?

mrbean-bremen:
> I'm not saying that you should operate it on GitHub Flow, but the cur-rent master is very halfway... 
Is there anything wrong with keeping master branch in good shape ?

I think that separating the drawing as is now in master, while it is not used for anything yet, is not a bad design per se. My point is that if we use the current state (or the state before the breaking changes will be integrated) for a later 3.x branch, it will be easier to merge fixes back and forth between branches (independent of what we actually branch - the same would be true id we would branch the 4.x branch).

So my currrent preference is to remain on the master branch as long as possible (e.g. we can still make bugfixes here). From experience I would say that branching as late as possible is always preferrable to avoid a later merge hell. 
We can decide to merge the changes @wieslawsoltes is working on into a separate branch first, but my personal preference would be to make a 3.x branch at that point (after making another release) and continue with the 4.x changes on the master branch.

CC @wieslawsoltes

H1Gdev:
> From experience I would say that branching as late as possible is always preferrable to avoid a later merge hell.

I think we can put merge into both branches at any time. (I think it's not so much trouble because you can cherry-pick to some extent.)

I think you disagree because of disagreements about the current state of master branch...
So I try to resolve this conflict.

mrbean-bremen:
> I think we can put merge into both branches at any time. (I think it's not so much trouble because you can cherry-pick to some extent.)

My point was that the separation of drawing was a disruptive change in regard to marges - it is easy to make merge errors if code has been split out and put into separate files, at least this is my experience. Cherry-picking will have the same problems. I may be wrong here, that's why I asked for other opinions.

PR #945, opened 2022-01-22
Add BoundsRelativeToTop

Happypig375:
There is currently no way to obtain the bounds of an element relative to the whole document instead of relative to the parent. This PR adds that.

### DIFF
--- a/Source/Basic Shapes/SvgPathBasedElement.Drawing.cs
+++ b/Source/Basic Shapes/SvgPathBasedElement.Drawing.cs
@@ -1,4 +1,4 @@
-#if !NO_SDC
+﻿#if !NO_SDC
 using System.Drawing;
 using System.Drawing.Drawing2D;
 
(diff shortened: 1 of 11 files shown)


H1Gdev:
What benefits does this PR have for users ?

Happypig375:
It can determine where to overlay HTML elements over SVG (input boxes), or determine collisions between elements given changing transforms and coordinates.

Happypig375:
This is a behavior I have to emulate myself by applying parent transforms to the bounds myself, by the way.

H1Gdev:
Can you give us concrete examples ?
We can not determine if this PR needs to be merged...

paulushub:
@Happypig375 Please explain the need for this feature. 
It is mostly `GraphicsPath` based operations, can this implemented as extension method?

mrbean-bremen:
@paulushub - I would rely on @H1Gdev to review this - I don't really have the inside to say if this is helpful.

paulushub:
> @paulushub - I would rely on @H1Gdev to review this - I don't really have the inside to say if this is helpful.

Testing it now, having merged it into `develop branch` of my fork. You can checkout my branch, flip between master/develop
and see the difference. So far, it is a useful contribution.
Using the quick test-run dialog I submitted in my PR, this PR already resolved 3 of the 8 exceptions still in the W3C Test Suite (those failing).
So the enhancements to the bounding box computation is useful, with no breaking change so-far, that was why I requested your review.

In fact, I am currently improving the quick test-run dialog to better compare before and after PR changes in the tests.

mrbean-bremen:
Well, that looks like you have sufficiently reviewed it yourself! Just go ahead in this case...

Happypig375:
I originally wanted to implement clickable buttons given an area inside the SVG. I later wrote this as an extension method in my code.

paulushub:
> I originally wanted to implement clickable buttons given an area inside the SVG. I later wrote this as an extension method in my code.

Thanks so much for the updates. The PR will still be around as we learn what to include.

PR #955, opened 2022-02-17
Fix: #954 'Invalid Cast Exception when updating SvgText Coordinates' issue.

### LINKED ISSUE #954: Invalid Cast Exception when updating SvgText Coordinates
The issue is happening when you want to update SvgText.X property value with a new one.

if you run this unit test you will see the error:

```csharp
[Test]
public void TestUpdateCoordinatesForCollectionChange()
{
    SvgText text = new SvgText()
    {
        Text = "Test coordinates",
        X = { 10 },
        Y = { 30 },
        Dx = { 40 },
        Dy = { 50 }
    };

    text.X = new SvgUnitCollection() { 20 };  // <- This line throws an Invalid Cast Exception

    var xml = text.GetXML();
    Assert.IsTrue(xml.Contains("x=\"20\""));
    Assert.IsTrue(xml.Contains("y=\"30\""));
    Assert.IsTrue(xml.Contains("dx=\"40\""));
    Assert.IsTrue(xml.Contains("dy=\"50\""));
}
```

Exception detail

```
System.InvalidCastException
  HResult=0x80004002
  Message=Unable to cast object of type 'Svg.SvgUnitCollection' to type 'Svg.SvgUnit'.
  Source=Svg
  StackTrace:
   at Svg.SvgAttributeCollection.UnboxAndCheck[T](Object a, Object b) in C:\Projects\Svg-Net\SVG\Source\SvgAttributeCollection.cs:line 136
   at Svg.SvgAttributeCollection.TryUnboxedCheck(Object a, Object b) in C:\Projects\Svg-Net\SVG\Source\SvgAttributeCollection.cs:line 118
   at Svg.SvgAttributeCollection.set_Item(String attributeName, Object value) in C:\Projects\Svg-Net\SVG\Source\SvgAttributeCollection.cs:line 99
   at Svg.SvgTextBase.set_X(SvgUnitCollection value) in C:\Projects\Svg-Net\SVG\Source\Text\SvgTextBase.cs:line 70
   at Svg.UnitTests.SvgTextTests.TestWritesCoordinatesForCollectionChange2() in C:\Projects\Svg-Net\SVG\Tests\Svg.UnitTests\SvgTextTests.cs:line 102
```

> From the code reference below you can see this happens in line 70

https://github.com/svg-net/SVG/blob/e37e4441641806205b6aeb055864a899de1adc36/Source/Text/SvgTextBase.cs#L56-L78

## We can see the following
From the 
* _x -> Type = SvgUnitCollection
* `incoming value` -> Type = SvgUnitCollection
* Attributes["x"] -> Type = SvgUnit

So when line 70 is executed it throws an invalid case exception `Attributes["x"] = value; `.

## My observations

+ It's valid for a SvgText element to have multiple values for X property as described in [MDN page](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/x#text), so passing a collection is valid
+ From the current code, We can see that it only stores the first element in the collection, see line 76 from referenced code. So I assume multiple values are not supported yet.
+ As coding practice we may want or not (it depends on what we want to do) overwrite a property that it has a collection, I guess most of the times we don't want to allow that, but in this case I think it is a good behavior, since you can update a set of values (in the furture) without the need of empty the current collection and then populate by using Collection.Add property, it is not too friendly in this case in my opinion

## Proposal
Having said this, I think the SvgText.X property should allow direct collection assignment following the current implementation, that is only supporting one single value for X property by now.

Based on what we have on line 76 I updated line 70 and it worked.
```csharp
[SvgAttribute("x")]
public virtual SvgUnitCollection X
{
    get { return _x; }
    set
    {
        if (_x != value)
        {
            if (_x != null) _x.CollectionChanged -= OnXChanged;
            _x = value;
            if (_x != null) _x.CollectionChanged += OnXChanged;

            IsPathDirty = true;
        }
        Attributes["x"] = value.FirstOrDefault(); // <- Adding FirstOrDefault sends the SVGUnit value and no cast exception happens
    }
}
```

## Notes
* Let me know if you agree with my approach, so I can create the PR.
Btw **The same behavior is in Y, Dx & Dy properties**

H1Gdev:
#### Reference Issue

Fixes #954


H1Gdev:
There are many anomalies to implementation of `SvgTextBase` class...
If we have problems with this PR, we need to set it to `Attributes`, not temporary fields.

Source/Text/SvgTextBase.cs
```diff
@@ -61,22 +70,16 @@ public virtual SvgUnitCollection X
...
 
                     IsPathDirty = true;
                 }
                 Attributes["x"] = value;
             }
         }
 
-        private void OnXChanged(object sender, NotifyCollectionChangedEventArgs e)
-        {
-            Attributes["x"] = X.FirstOrDefault();
```
H1Gdev, line 76: Cause of exception is here.

Source/Text/SvgTextBase.cs
```diff
@@ -19,10 +19,19 @@ public abstract partial class SvgTextBase : SvgVisualElement
 
         public SvgTextBase()
         {
-            _x.CollectionChanged += OnXChanged;
-            _dx.CollectionChanged += OnDxChanged;
-            _y.CollectionChanged += OnYChanged;
-            _dy.CollectionChanged += OnDyChanged;
+            NotifyCollectionChangedEventHandler xHandler = null;
```
H1Gdev, line 22: It is enough to set `Attributes` only once, which is changed from default value.

mrbean-bremen (approved):
Thanks!

Issue #960, opened 2022-03-16
RequiresSmoothRendering is always false for an SvgRectangle if CornerRadiusX and CornerRadiusY are 0


supranowitz:
#### Description
Anti-aliasing is never used for a simple rectangle without rounded corners, but forcing anti-aliasing to be enabled improves the rendering.  I do not quite understand the reasoning for the logic in SvgRectangle.  

Currently, RequiresSmoothRendering is implemented like this:

```
        protected override bool RequiresSmoothRendering
        {
            get
            {
                if (base.RequiresSmoothRendering)
                    return (CornerRadiusX.Value > 0.0f || CornerRadiusY.Value > 0.0f);
                else
                    return false;
            }
        }
```

But this would make more sense to me (no matter what base.RequiresSmoothRendering returns, force it on if corners are rounded):
```
        protected override bool RequiresSmoothRendering
        {
            get
            {
                if (!base.RequiresSmoothRendering)
                    return (CornerRadiusX.Value > 0.0f || CornerRadiusY.Value > 0.0f);
                else
                    return true;
            }
        }
```

#### Example data

| SVG | svg-net rendering |
| ----------- | ----------- |
| [Rectangles](https://user-images.githubusercontent.com/64907346/158681119-35009cf9-012d-4a51-b711-c6efc8069bf4.svg)| <img width="21" alt="rectangles" src="https://user-images.githubusercontent.com/64907346/158680309-e3c41283-1344-43f3-985b-223ba2464cc4.png">|
| [Rectangles with corner radius](https://user-images.githubusercontent.com/64907346/158681166-796b0742-5346-48a2-9767-f045cc484b0f.svg)| <img width="21" alt="rectangles-rounded" src="https://user-images.githubusercontent.com/64907346/158680998-4be58e10-c7b6-4564-a7cd-55454173d47e.png">|

When the original svg is rendered to a small bitmap the spacing and bottom vertical alignment does not appear to be consistent, but the rendering is improved when anti-aliasing is forced to be on by adding a very small corner radius to the rectangles.

#### Used Versions

SVG 3.3.0
.NET Framework 4.8

H1Gdev:
Condition for anti-aliasing in `SvgRectangle` have been around since early days.

https://github.com/svg-net/SVG/blob/cf59fef87fbac69ed12f51123f85a7ac363b29d2/Source/Basic%20Shapes/SvgRectangle.cs#L103

mrbean-bremen:
[...]

PR #965, opened 2022-04-03
Consolidate anti-aliasing behavior with browsers

mrbean-bremen:
- geometricPrecision now uses anti-aliasing
- no special handling for rectangles with rounded edges
- conforms to rendering in major browsers
- see #960

@H1Gdev - I used your example as a test image.
I also had to adapt another rectangle test image which is now rendered anti-aliased (which is consistent withe browsers, but was intentionally not ant-aliased before). I'm not 100% that this is the correct thing to do...


H1Gdev:
@mrbean-bremen 
I will check this PR.

H1Gdev (approved):

Issue #966, opened 2022-04-04
System.ArgumentException: ColorBlend object that was set is not valid for svg with linear gradient


nickdekesel:
#### Description

I am getting the following exception when calling SvgDocument.Draw on a specific svg:
```
System.ArgumentException: ColorBlend object that was set is not valid.Position's last element must be equal to 1.0. ColorBlend objects must be constructed with the same number of positions and color values. Positions must be between 0.0 and 1.0, 1.0 indicating the last element in the array.
   at System.Drawing.Drawing2D.LinearGradientBrush.set_InterpolationColors(ColorBlend value)
   at Svg.SvgLinearGradientServer.GetBrush(SvgVisualElement renderingElement, ISvgRenderer renderer, Single opacity, Boolean forStroke)
   at Svg.SvgDeferredPaintServer.GetBrush(SvgVisualElement styleOwner, ISvgRenderer renderer, Single opacity, Boolean forStroke)
   at Svg.SvgVisualElement.RenderFill(ISvgRenderer renderer)
   at Svg.SvgVisualElement.RenderFillAndStroke(ISvgRenderer renderer)
   at Svg.SvgVisualElement.RenderInternal(ISvgRenderer renderer, Action`1 renderMethod)
   at Svg.SvgVisualElement.RenderInternal(ISvgRenderer renderer, Boolean renderFilter)
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer)
   at Svg.SvgElement.RenderChildren(ISvgRenderer renderer)
   at Svg.SvgVisualElement.RenderInternal(ISvgRenderer renderer, Action`1 renderMethod)
   at Svg.SvgVisualElement.RenderInternal(ISvgRenderer renderer, Boolean renderFilter)
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer)
   at Svg.SvgElement.RenderChildren(ISvgRenderer renderer)
   at Svg.SvgVisualElement.RenderInternal(ISvgRenderer renderer, Action`1 renderMethod)
   at Svg.SvgVisualElement.RenderInternal(ISvgRenderer renderer, Boolean renderFilter)
   at Svg.SvgVisualElement.Render(ISvgRenderer renderer)
   at Svg.SvgElement.RenderChildren(ISvgRenderer renderer)
   at Svg.SvgElement.Render(ISvgRenderer renderer)
   at Svg.SvgFragment.Render(ISvgRenderer renderer)
[...]

PR #961, opened 2022-03-17
Fix: precision issue in SvgUnit.

H1Gdev:
#### Reference Issue

Fixes https://github.com/svg-net/SVG/pull/903#issuecomment-1069408039

#### What does this implement/fix? Explain your changes.

Change order of operations.

PR #968, opened 2022-04-04
Account for rounding errors in gradient stop positions

mrbean-bremen:
- if the first/last value is not exactly 0/1,
  an exception is raised by Drawing2D
- could happen due to rounding errors
- see #966

This fixes only the crash. In this case the last stop value was 0.999994 instead of 1.0, and it is now always set to 1.
Note that cases where 0 and 1 stop values are missing are already handled by inserting them. This case only happens if the values in the svg are correct (100% in this case), and get slightly changed due to rounding errors for certain positions.

Also note that I added a test image, but did not add it to passing images, because the resulting image is still wrong (it has no gradient). This has also to be fixed before enabling the test.


H1Gdev:
PR #961 couldn't fix all situations. I will consider how it can be fixed.

mrbean-bremen:
Crash is fixed by the other PR.

PR #969, opened 2022-04-05
Fix: precision issue in SvgUnit.

H1Gdev:
#### Reference Issue

Fixes part of #966
Ref. #961


H1Gdev:
The issue that gradation is not correct will be fixed in another PR.

mrbean-bremen (approved):

PR #983, opened 2022-06-13
Update XMLOutputTester.

H1Gdev:

- Allow test code to be modified.


mrbean-bremen (approved):
Nice! Maybe you can adapt the form title from `Form1`  to something like `SVG Builder`.

mrbean-bremen (approved):
Thanks!

Issue #978, opened 2022-05-07
SVG 3.4.2 version SvgDocument Draw Out of memory


DebbieGao:
#### Description

When update to svg 3.4.2, the svgDocument.Draw() throw exception, out of memory.
If use version 3.4.0, it is OK. 
#### Used Versions

SVG 3.4.2
.net standard 2.0

mrbean-bremen:
Can you show the image that causes that? Or does this happen for all images?

DebbieGao:
> Can you show the image that causes that? Or does this happen for all images?

That is a really specific image, which is .svg and about 1100KB. I did not try other images.

H1Gdev:
@DebbieGao 

Can you publish SVG file ?

DebbieGao:
> 

sorry, I cannot publish it.

H1Gdev:
The tendency is that `gradation` seems to be out of memory.
Is it possible to create minimum SVG that will fail ?

wshiadai:
I have  the same problem
this is my demo code
` private void btnSVG2_Click(object sender, EventArgs e)
        {
            try
            {
                var svgDoc = SvgDocument.Open<SvgDocument>(path);
                Bitmap bmp = svgDoc.Draw();
                bmp.Save(@"D:\Liu\Project\DataPenetrationDemo\DataPenetrationDemo\Static\liu002.png");
                bmp.Dispose();
                MessageBox.Show("SVG已生成图片");
            }
            catch (Exception ex)
            {
                MessageBox.Show(ex.Message);
            }
           
        }`
result：
![image](https://user-images.githubusercontent.com/1819491/177458470-d406ae59-b19f-4ec0-83d0-bd671fb35e4c.png)

my svg file:
`<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 66.81 66.68">
    <defs>
        <linearGradient id="未命名的渐变_7" x1="21.39" y1="33.34" x2="45.3" y2="33.34" gradientUnits="userSpaceOnUse">
            <stop offset="0" stop-color="#32343d"/>
            <stop offset="0.13" stop-color="#18121c"/>
            <stop offset="0.55" stop-color="#434652"/>
            <stop offset="0.63" stop-color="#3f414d"/>
            <stop offset="0.74" stop-color="#32323d"/>
            <stop offset="0.86" stop-color="#1e1924"/>
[...]

PR #991, opened 2022-07-13
Fixes out of memory in Svg 3.4.2

### LINKED ISSUE #989: Rendering a gradient using SvgDocument.Draw() runs out of memory in Svg 3.4.2
#### Description
Rendering an SvgDocument containing a gradient to a bitmap runs out of memory if the document is repeatedly rendered.

#### Example data
Example project that repeatedly renders an SvgDocument to a Bitmap and displays the bitmap in a PictureBox.
[SvgGradientTest.zip](https://github.com/svg-net/SVG/files/9095733/SvgGradientTest.zip)

#### Used Versions
Svg 3.4.2 (problem does not occur in earlier versions)
Windows 11
.NET 6

#### Additional information
This problem may be related to issue #978, which also uses a gradient.

The problem appears to be caused by SvgGradientServer.LoadStops(), which was introduced in version 3.4.2:
```
        private void LoadStops(SvgVisualElement parent)
        {
            var gradient = this;
            while (gradient?.Stops.Count == 0)
                gradient = SvgDeferredPaintServer.TryGet<SvgGradientServer>(gradient.InheritGradient, parent);

            if (gradient != null)
                Stops.AddRange(gradient.Stops);
        }
```
If this.Stops.Count is nonzero on entry, Stops.AddRange(gradient.Stops) will append this.Stops to itself. If this is done repeatedly, the length of this.Stops increases geometrically, leading to an out-of-memory condition. One solution is to check for this condition:
```
        private void LoadStops(SvgVisualElement parent)
        {
            var gradient = this;
            while (gradient?.Stops.Count == 0)
                gradient = SvgDeferredPaintServer.TryGet<SvgGradientServer>(gradient.InheritGradient, parent);

            if (gradient != null && gradient != this)
                Stops.AddRange(gradient.Stops);
        }
```

H1Gdev:
#### Reference Issue

Fixes #989

#### Any other comments?

#978 may also be fixed.

### DIFF
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -9,6 +9,9 @@ The release versions are NuGet releases.
 ### Enhancements
 * update XMLOutputTester (rename to SVGBuilder) (see [PR #983](https://github.com/svg-net/SVG/pull/983))
 
+### Fixes
+* fixed out of memory in Svg 3.4.2 (see [#989](https://github.com/svg-net/SVG/issues/989))
+
 ## [Version 3.4.2](https://www.nuget.org/packages/Svg/3.4.2)  (2022-04-11)
 
 ### Changes


mrbean-bremen (approved):
Thanks - that is great! I think we have to make a patch release with this fix soon, maybe at the weekend.

PR #723, opened 2020-05-02
Redesign of SvgFontManager.

H1Gdev:
#### Reference Issue

Ref. #213, #414, #706, #711 

#### What does this implement/fix? Explain your changes.

- Manage `PrivateFontCollection`.
- Dispose `FontFamily`.
- Expand `LocalizedFamilyNames`.

PR #993, opened 2022-08-03
Fixes localized family names in SvgFontManager.

H1Gdev:
#### Reference Issue

Ref. #706, #723

#### Any other comments?

Issue that occurred in #723.


H1Gdev:
I modified files so that test would catch it.

H1Gdev:
@mrbean-bremen 

test is no longer passing...

Is there a way to install fonts in GitHub Workflow ?
If not, I will skip this test.

mrbean-bremen:
> Is there a way to install fonts in GitHub Workflow ?

I have no idea - better to skip the test for now, and maybe add an issue for installing a font in the CI environment.

H1Gdev:
@mrbean-bremen 

Changed test to use font installed in Workflow.

mrbean-bremen (approved):

PR #1030, opened 2022-12-18
Fix: Compile error in latest C#.

H1Gdev:
#### Reference Issue

Ref. [Instance method on ref struct may capture unscoped ref parameters](https://learn.microsoft.com/dotnet/csharp/whats-new/breaking-changes/compiler%20breaking%20changes%20-%20dotnet%207#instance-method-on-ref-struct-may-capture-unscoped-ref-parameters)

### DIFF
--- a/Source/Paths/CoordinateParser.cs
+++ b/Source/Paths/CoordinateParser.cs
@@ -24,7 +24,7 @@ namespace Svg
         public int Position;
         public bool HasMore;
 
-        public CoordinateParserState(scoped ref ReadOnlySpan<char> chars)
+        public CoordinateParserState(ref ReadOnlySpan<char> chars)
         {
             CurrNumState = NumState.Separator;
             NewNumState = NumState.Separator;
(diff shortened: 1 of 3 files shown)


Source/Paths/CoordinateParser.cs
```diff
@@ -24,7 +24,7 @@ public enum NumState
         public int Position;
         public bool HasMore;
 
-        public CoordinateParserState(ref ReadOnlySpan<char> chars)
+        public CoordinateParserState(scoped ref ReadOnlySpan<char> chars)
```
mrbean-bremen, line 27: As far as Im understand, the `scoped` keyword wa introduced in C#11. What about compiling for older versions?

H1Gdev: It will be a compilation error.
We can also pin language version to C#10 if we want that.

mrbean-bremen: At the moment I wouldn't want to restrict the compiler to C# 11, so pinning to C# 10 is indeed an option, at least for a time. If there is an option to make it compile under both versions (using #ifdef) it would be better, but I don't know if there is a possibility to do this.

mrbean-bremen (commented):
I think this is ok for the time being. Maybe you can also add a line in the release notes...

mrbean-bremen (approved):

PR #781, opened 2021-01-04
Update Nerdbank.GitVersioning package to fix dotnet build errors

wieslawsoltes:
#### What does this implement/fix? Explain your changes.

Fixed `dotnet build` as I am getting the following error:

```
/opt/dotnet/5.0.100/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.2' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [/home/codespace/workspace/SVG/Source/Svg.csproj]
/opt/dotnet/5.0.100/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.2' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [/home/codespace/workspace/SVG/Source/Svg.csproj]
/var/nuget/nerdbank.gitversioning/3.0.26/build/Nerdbank.GitVersioning.targets(63,5): error MSB4018: The "Nerdbank.GitVersioning.Tasks.GetBuildVersion" task failed unexpectedly. [/home/codespace/workspace/SVG/Source/Svg.csproj]
/var/nuget/nerdbank.gitversioning/3.0.26/build/Nerdbank.GitVersioning.targets(63,5): error MSB4018: System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. [/home/codespace/workspace/SVG/Source/Svg.csproj]
/var/nuget/nerdbank.gitversioning/3.0.26/build/Nerdbank.GitVersioning.targets(63,5): error MSB4018:  ---> System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.DotNet.PlatformAbstractions, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified. [/home/codespace/workspace/SVG/Source/Svg.csproj]
/var/nuget/nerdbank.gitversioning/3.0.26/build/Nerdbank.GitVersioning.targets(63,5): error MSB4018:  [/home/codespace/workspace/SVG/Source/Svg.csproj]
[...]

PR #503, opened 2019-06-27
Netcore merge

mrbean-bremen:
This is a cleaned up version of the .NetCore2.2 branch which only contains the new commits.
This is to avoid problems due to different base branches - the relevant commits are cherry-picked and adapted to avoid any conflicts. As far as I can see, this has all needed changes from that branch.

@tebjan - please have a look at the commits (especially yours). If this is ok, we can rebase and merge this.

PR #1036, opened 2023-02-06
Update Nerdbank.GitVersioning from version 3.3.37 to 3.5.119

Hurricane31337:
Currently, the project is not compilable with Visual Studio. In order to fix this, Nerdbank.GitVersioning needed to be updated from version 3.3.37 to 3.5.119.

#### Any other comments?
Additionally, I had to remove these lines in Samples\SvgConsole\SvgConsole.csproj:
```
    <IsTool>True</IsTool>
    <PackAsTool>True</PackAsTool>
```
I didn't commit this, but I would consider removing them to make it easier for the other devs (or add this info to CONTRIBUTING.md or README.md).

### DIFF
--- a/Source/Svg.csproj
+++ b/Source/Svg.csproj
@@ -130,7 +130,7 @@
     <PackageReference Include="System.Memory" Version="4.5.4" />
     <PackageReference Include="ExCSS" Version="4.1.4" />
     <PackageReference Include="Fizzler" Version="1.2.1" />
-    <PackageReference Include="Nerdbank.GitVersioning" Version="3.3.37">
+    <PackageReference Include="Nerdbank.GitVersioning" Version="3.5.119">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>


mrbean-bremen:
> Additionally, I had to remove these lines in Samples\SvgConsole\SvgConsole.csproj

Do you know the reason for this? Have these properties been removed in newer VS version?
I'm not sure is these are actually needed - if they are problematic, I would rather remove them.

@wieslawsoltes - you did add this example, what do you think?

H1Gdev:
@mrbean-bremen 

This package has been added in PR #503.
What do you need for versioning when you release to nuget ?

Hurricane31337:
Yes I got compile errors which were resolved after updating this library. Here is another pull request that suggested updating the library because of compile errors: https://github.com/svg-net/SVG/pull/781
Sorry, I'm currently not at my PC to revert the changes to paste the exact errors or investigate further. I would assume that updating the libary two minor versions should be fine.

mrbean-bremen:
Updating the library is fine, my question was just about the`IsTool` / `PackAsTool` properties. If they cause a compile error, I'm in favor of removing them also, but I first wanted to understand what the problem is...

paulushub:
@Hurricane31337 Please consider updating to the latest version.

paulushub:
We will be dropping support for `Nerdbank.GitVersioning`, using the GitHub Workflow tools for versioning.

PR #1037, opened 2023-02-06
Update .NET Framework from version 4.6.2 to 4.8

Hurricane31337:
This commit updates .NET Framework from version 4.6.2 to 4.8.

#### Any other comments?
If there is a specific reason to stay on 4.6.2, feel free to discard this pull request. It compiles and works just fine (I tested the DLL).

### DIFF
--- a/Samples/Entities/Entities.csproj
+++ b/Samples/Entities/Entities.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net462;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net48;netcoreapp3.1</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Samples/SVGBuilder/SVGBuilder.csproj
+++ b/Samples/SVGBuilder/SVGBuilder.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFrameworks>net462;net5.0-windows</TargetFrameworks>
+    <TargetFrameworks>net48;net5.0-windows</TargetFrameworks>
     <UseWindowsForms>true</UseWindowsForms>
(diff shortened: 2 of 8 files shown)


mrbean-bremen:
> If there is a specific reason to stay on 4.6.2, 

We try to support the oldest possible version that is still supported by MS. As long as no features from newer versions are needed, this should not be a problem, as they are are compatible with newer versions.
.NET 4.6.2 is [supported until 2027](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework), so I think we can leave it for a while.

>  It compiles and works just fine

Not for .NET 4.6.2, as you can see 😀

mrbean-bremen:
I'm closing this, as it won't be merged in the near future. What we would need though is updating the tests from the now unsupported net5 to a newer version...

charlesroddie:
> > It compiles and works just fine
> 
> Not for .NET 4.6.2, as you can see 😀

That's not correct.

Consider the change to [Source/Svg.csproj](https://github.com/svg-net/SVG/pull/1037/files#diff-6b81e81274156646b0e88c7221aa9e9cbc5f3b797878a6d0ac3e24d6c233f2b4)
```
<TargetFrameworks>netstandard2.0;netstandard2.1;netcoreapp3.1;net5.0;net462</TargetFrameworks>
<TargetFrameworks>netstandard2.0;netstandard2.1;netcoreapp3.1;net5.0;net48</TargetFrameworks>
```
Both before and after, the library is built for various frameworks, quite unnecessarily.

After this PR it appears that net462 is dropped but that's not true as netstandard2.0 supports net462 and the original net462 was not necessary.

In fact netstandard2.1, netcoreapp3.1, net5.0, net462 and net48 are all unnecessary as any platform that supports those libraries will support netstandard2.0 libraries.

mrbean-bremen:
I mostly agree, see my comments on the PR.

paulushub:
I will be closing this PR. Supports for both 4.6.2 and 4.8 will not be an issue. 
**NOTE**: Will wait for another 1 or 2 for a response.

PR #1045, opened 2023-02-23
Update from .NET 5 to .NET 6.

H1Gdev:

Update from .NET 5 to .NET 6.

https://dotnet.microsoft.com/platform/support/policy/dotnet-core

#### Any other comments?

Support for .NET 5 has already ended.


mrbean-bremen:
@H1Gdev - what is the state of this PR? Is there anything blocking this?

H1Gdev:
I am considering solution of [this warning](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1416)...

Samples/SVGViewer/SvgViewer.cs
```diff
@@ -69,15 +69,25 @@ private void TextBox1_KeyDown(object sender, KeyEventArgs e)
...
 
             //using (var render = new DebugRenderer())
             //    svgDoc.Draw(render);
             svgImage.Image = svgDoc.Draw();
 
             var baseUri = svgDoc.BaseUri;
             var outputDir = Path.GetDirectoryName(baseUri != null && baseUri.IsFile ? baseUri.LocalPath : Application.ExecutablePath);
+#if NET5_0_OR_GREATER
+            if (OperatingSystem.IsWindows())
+                svgImage.Image?.Save(Path.Combine(outputDir, "output.png"));
```
mrbean-bremen, line 87: So what woud happen on non-Windows platforms here? I'm not sure if we shall raise an exception, log a warning, or disable the whole function in this case, what do you think?

Tests/SvgW3CTestRunner/View.cs
```diff
@@ -1,41 +1,36 @@
...
     public partial class View : Form
     {
-        //DIRECTORY SEPARATOR: The value of this field is a slash ("/") on UNIX and on Mac OSX, and a backslash ("\") on the Windows operating systems.
-        static private string sprt = Path.DirectorySeparatorChar.ToString();
-
         //Data folders
-        private string _svgBasePath = @".." + sprt + ".." + sprt + ".." + sprt + ".." + sprt + "W3CTestSuite" + sprt + "svg" + sprt;
-        private string _pngBasePath = @".." + sprt + ".." + sprt + ".." + sprt + ".." + sprt + "W3CTestSuite" + sprt + "png" + sprt;
+        private string _svgBasePath = Path.Combine("..", "..", "..", "..", "W3CTestSuite", "svg");
+        private string _pngBasePath = Path.Combine("..", "..", "..", "..", "W3CTestSuite", "png");
```
mrbean-bremen, line 18: Much better!

Source/Document Structure/SvgImage.Drawing.cs
```diff
@@ -209,22 +209,28 @@ public object GetImage(string uriString)
...
+                    using (var stream = httpResponseMessage.Content.ReadAsStreamAsync().Result)
+                    {
+                        if (uri.LocalPath.EndsWith(".svg", StringComparison.InvariantCultureIgnoreCase) ||
+                            httpResponseMessage.Headers.Contains(MimeTypeSvg))
+                            return LoadSvg(stream, uri);
+                        else
+                            using (var image = Image.FromStream(stream))
+                                return new Bitmap(image);
+                    }
+                else
```
mrbean-bremen, line 232: Is this change related to the platform change?

H1Gdev: WebRequest is obsolete, so we use HttpClient instead.

https://learn.microsoft.com/dotnet/api/system.net.webrequest#remarks

mrbean-bremen: Ah ok, thanks!

Samples/SVGViewer/SvgViewer.cs
H1Gdev: SVGViewer is currently Windows platform only. (Because System.Drawing is Windows platform only...)

Here only resolve warning.
If we run SVGViewer on non-Windows platforms, we need to consider drawing system first.

mrbean-bremen (approved):

PR #1048, opened 2023-03-12
Update from .NET 5 to .NET 6.

H1Gdev:
#### Reference Issue

Ref. PR #1045 

Change order of values in \<TargetFrameworks\>.

### DIFF
--- a/Samples/Entities/Entities.csproj
+++ b/Samples/Entities/Entities.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net462;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


mrbean-bremen:
Can you please explain why this is needed?

H1Gdev:
SVGViewer etc. will start using the leftmost framework.
I thought it would be better to start with the latest framework rather than net462 starting by default.

mrbean-bremen:
You mean if you have compiled several version and start them inside Visual Studio? Isn't the target shown and selectable in VS? (I'm usually using Rider instead of VS)

H1Gdev:
It is possible to choose which framework to run.
I think default framework is `.NET` rather than `.NET Framework`.

mrbean-bremen:
> I think default framework is .NET rather than .NET Framework.

In this case, what difference does the change make? I'm obviously missing something...

H1Gdev:
SVGViewer

master
master

PR #1048 
pr1048

mrbean-bremen:
Ok, I see, thank you. It's a bit of a cosmetic issue, but I agree that's a bit more convenient.

mrbean-bremen (approved):

PR #1053, opened 2023-05-10
Fix `DeepCopy` result miss the `_styles` issue

mzh3511:
Some svg elements have styles that may include `clip-path`, the `DeepCopy` method should handle the `_styles`

### DIFF
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -1044,6 +1044,10 @@ namespace Svg
             foreach (var node in Nodes)
                 newObj.Nodes.Add(node.DeepCopy());
 
+            foreach (var style in _styles)
+            foreach (var pair in style.Value)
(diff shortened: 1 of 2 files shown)


mrbean-bremen (commented):
Thanks! Can you please an entry into the [release notes](https://github.com/svg-net/SVG/blob/master/doc/ReleaseNotes.md)?

mrbean-bremen (approved):

Source/SvgElement.cs
```diff
@@ -1044,6 +1044,10 @@ public virtual SvgElement DeepCopy<T>() where T : SvgElement, new()
             foreach (var node in Nodes)
                 newObj.Nodes.Add(node.DeepCopy());
 
+            foreach (var style in _styles)
+            foreach (var pair in style.Value)
```
H1Gdev, line 1048: Please add indent.

PR #1055, opened 2023-05-18
Fix the color string format incompatible with the Edge/Chrome in case of no `System.Drawing.Common`

mzh3511:
Fix the color string format incompatible with the Edge/Chrome in case of no `System.Drawing.Common`

In case of no `System.Drawing.Common`, the `SvgColourConverter.ConvertTo` method returns `'Red'`(with extra quotation mark), or `255, 17, 0` this lead to the resaved svg file can not display correctly by the Edge/Chrome, and this is not fit the svg w3c standard

So here introduce `SvgColourConverter.ToHtml` to handle it

### DIFF
--- a/Source/Painting/SvgColourConverter.cs
+++ b/Source/Painting/SvgColourConverter.cs
@@ -204,7 +204,30 @@ namespace Svg
 #endif
             }
 
-            return base.ConvertTo(context, culture, value, destinationType);
+            return ToHtml((Color)value);
+        }
+
+        /// <summary>
+        /// Converts color to html string format.
+        /// Refer to https://source.dot.net/#System.Drawing.Primitives/System/Drawing/ColorTranslator.cs
+        /// </summary>
+        /// <param name="c"></param>
+        /// <returns></returns>
+        private static string ToHtml(Color c)
+        {
+            var colorString = string.Empty;
+            if (c.IsEmpty)
+                return colorString;
+
+            if (c.IsNamedColor)
+            {
+                colorString = c == Color.LightGray ? "LightGrey" : c.Name;
+                colorString = colorString.ToLowerInvariant();
+            }
+            else
+                colorString = $"#{c.R:X2}{c.G:X2}{c.B:X2}";
+
+            return colorString;
         }
 
         /// <summary>
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -8,8 +8,9 @@ The release versions are NuGet releases.
 
 ### Fixes
 * fixed build error in C# 11 (see [PR #1030](https://github.com/svg-net/SVG/pull/1030))
-* fixed out of memory exception on SVGs with gradients (see [PR #1038] (https://github.com/svg-net/SVG/pull/1038))
-* fixed missing styles when `DeepCopy` the `SvgElement` (see [PR #1053] (https://github.com/svg-net/SVG/pull/1053))
+* fixed out of memory exception on SVGs with gradients (see [PR #1038](https://github.com/svg-net/SVG/pull/1038))
+* fixed missing styles when `DeepCopy` the `SvgElement` (see [PR #1053](https://github.com/svg-net/SVG/pull/1053))
+* fix the color string format incompatible with the Edge/Chrome browsers in case of no System.Drawing.Common (see [PR #1055](https://github.com/svg-net/SVG/pull/1055))
 
 ## [Version 3.4.4](https://www.nuget.org/packages/Svg/3.4.4)  (2022-10-29)
 


mzh3511:
The change fixes the following case under the no `System.Drawing.Common` condition:

Here's a simple svg file, its content:
```
<svg width="100%" height="100%" viewBox="0 0 100 100"
     xmlns="http://www.w3.org/2000/svg">

  <circle cx="50" cy="50" r="40" style="fill:#f10;stroke:Blue;"/>
</svg>
```
Open the svg file and save it to another file, the codes are the following:
```
        _svgDocument = SvgDocument.Open(@"g:\demo2.svg");
        _svgDocument.Write(@"g:\demo2_1.svg");
```

Got the `demo2_2.svg`, and its content:
```
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xml="http://www.w3.org/XML/1998/namespace" width="100%" height="100%" viewBox="0, 0, 100, 100">
  <circle cx="50" cy="50" r="40" style="fill:255, 17, 0;stroke:'Blue';" />
</svg>
```

Open the resaved svg by Edge or Chrome, the svg can not display correctly, although it can display by my codes:
image
We can see the circle shape style in svg is not recognized.

I investigated the [wieslawsoltes/SVG](https://github.com/wieslawsoltes/SVG) and [svg-net/SVG](https://github.com/svg-net/SVG), and found that this is because of the conditional compilation symbol (`NO_SDC` )  in line200 in  `SvgColourConverter.cs`:
image

So, I created the pull request to handle the `SvgColourConverter.ConvertTo` to return the color  html string format in case of no `System.Drawing.Common`

Source/Painting/SvgColourConverter.cs
```diff
@@ -204,7 +204,30 @@ public override object ConvertTo(System.ComponentModel.ITypeDescriptorContext co
 #endif
             }
 
-            return base.ConvertTo(context, culture, value, destinationType);
+            return ToHtml((Color)value);
```
mrbean-bremen, line 207: Would it be better to fix `ConvertTo` to also handle the `NO_SDC` case?
This way it would could used elsewhere.

doc/ReleaseNotes.md
```diff
@@ -10,6 +10,7 @@ The release versions are NuGet releases.
 * fixed build error in C# 11 (see [PR #1030](https://github.com/svg-net/SVG/pull/1030))
 * fixed out of memory exception on SVGs with gradients (see [PR #1038] (https://github.com/svg-net/SVG/pull/1038))
 * fixed missing styles when `DeepCopy` the `SvgElement` (see [PR #1053] (https://github.com/svg-net/SVG/pull/1053))
+* fix the color string format incompatible with the Edge/Chrome in case of no System.Drawing.Common (see [PR #1055] (https://github.com/svg-net/SVG/pull/1055))
```
H1Gdev, line 13: Please remove space...
Link does not work.

Source/Painting/SvgColourConverter.cs
mzh3511: Are you mean you want to change the `ConvertTo` method as follows?
```
        public override object ConvertTo(System.ComponentModel.ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
        {
            return ToHtml((Color)value);
        }
```

But the `ToHtml` method can not handle those `KnownColor`
For example, an `activeborder` color (comes from `Color.FromKnownColor(KnownColor.ActiveBorder)`), the `base.ConvertTo` and `ColorTranslator.ToHtml` return `activeborder`, `ToHtml` method will return a string start with `#`

Please confirm, thanks

doc/ReleaseNotes.md
mzh3511: Will fix it together with other comments, thanks for the mention

mzh3511: I will fix line11, 12, 13 in all

Source/Painting/SvgColourConverter.cs
mrbean-bremen: No, strike that, sorry - it should be ok as is.

doc/ReleaseNotes.md
```diff
@@ -8,8 +8,9 @@ The release versions are NuGet releases.
 
 ### Fixes
 * fixed build error in C# 11 (see [PR #1030](https://github.com/svg-net/SVG/pull/1030))
-* fixed out of memory exception on SVGs with gradients (see [PR #1038] (https://github.com/svg-net/SVG/pull/1038))
-* fixed missing styles when `DeepCopy` the `SvgElement` (see [PR #1053] (https://github.com/svg-net/SVG/pull/1053))
+* fixed out of memory exception on SVGs with gradients (see [PR #1038](https://github.com/svg-net/SVG/pull/1038))
+* fixed missing styles when `DeepCopy` the `SvgElement` (see [PR #1053](https://github.com/svg-net/SVG/pull/1053))
+* fix the color string format incompatible with the Edge/Chrome in case of no System.Drawing.Common (see [PR #1055](https://github.com/svg-net/SVG/pull/1055))
```
mrbean-bremen, line 13: minor nitpicking: either "with Edge/Chrome" or "with the Edge/Chrome browsers".

mzh3511: "with the Edge/Chrome browsers" used in new update

H1Gdev:
Please add tests for what this PR fixes.

mzh3511:
>

mzh3511:
> Please add tests for what this PR fixes.

Sorry, but this statement is not reached, because our repository does not add the `NO_SDC`  symbol, so I can not covered the statement by testing `SvgColourConverter.ConvertTo`

mzh3511:
Sorry, I clicked `Close with comment` by a mistake

mrbean-bremen (approved):

PR #1058, opened 2023-05-24
Fix: duplication of excessive SvgElement in DeepCopy

### LINKED ISSUE #1054: Problem writing a changed DeepCopied document
Hello! I have a problem writing a changed DeepCopied document.

#### Description

I have an original SVG file, that looks like that
image
I open it and make a deep copy of it:
```c#
var sampleDoc = SvgDocument.Open(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"../../../sample_closed.svg"));
var newDoc = (SvgDocument)sampleDoc.DeepCopy<SvgDocument>();
SvgPath newPath = (SvgPath)newDoc.Children[1].Children[1];
``` 

Then I make some changes to the path elements of the deep copied document. For example:
```c#
newPath.PathData[1] = new PointF(newPath.PathData[1].End.X - 1000, newPath.PathData[1].End.Y - 1000);
``` 

After that I call `Draw` method to get changed png file as a result:
```c#
newDoc.Draw().Save(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"../../../dataset/" + index + ".png"));
```  
And it works fine:
image

Then I try to save changed svg, calling `Write` method:
```c#
newDoc.Write(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, @"../../../dataset/" + index + ".svg"));
``` 
But I get a svg file without any change as a result

Is it a bug or I do something wrong?

#### Used Versions
Target framework: .net 6.0

H1Gdev:
#### Reference Issue

Fixes #1054

### DIFF
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -1045,8 +1045,8 @@ namespace Svg
                 newObj.Nodes.Add(node.DeepCopy());
 
             foreach (var style in _styles)
-            foreach (var pair in style.Value)
-                newObj.AddStyle(style.Key, pair.Value, pair.Key);
+                foreach (var pair in style.Value)
+                    newObj.AddStyle(style.Key, pair.Value, pair.Key);
 
             return newObj;
         }
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -1042,7 +1042,18 @@ namespace Svg
                 newObj.CustomAttributes.Add(attribute.Key, attribute.Value);
 
             foreach (var node in Nodes)
-                newObj.Nodes.Add(node is SvgElement ? newObj.Children[Children.IndexOf((SvgElement)node)] : node.DeepCopy());
+            {
+                if (node is SvgElement)
+                {
+                    var index = Children.IndexOf((SvgElement)node);
+                    if (index >= 0)
+                    {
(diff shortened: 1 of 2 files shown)


H1Gdev:
Please wait a moment as I will reconsider.

Source/SvgElement.cs
```diff
@@ -1042,7 +1042,18 @@ public virtual SvgElement DeepCopy<T>() where T : SvgElement, new()
...
+                {
+                    var index = Children.IndexOf((SvgElement)node);
+                    if (index >= 0)
+                    {
+                        newObj.Nodes.Add(newObj.Children[index]);
+                        continue;
+                    }
+                }
+                newObj.Nodes.Add(node.DeepCopy());
+            }
```
mrbean-bremen, line 1056: Is it guaranteed that at this point the elements have already been copied? Otherwise, we would need two passes, where first the objects are collected and copied in the second pass if not already copied.

H1Gdev: `DeepCopy` of `Children` is done just before.(line 986)

mrbean-bremen: Ah, right of course. Thanks!

mrbean-bremen (approved):

PR #1066, opened 2023-07-16
Simplify target frameworks

charlesroddie:

- Make the library netstandard2.0 only. This supports all current platforms so specifically building for other target frameworks is not necessary. Note: currently there is a difference in StringParser.cs which on netstandard2.1 parses directly from a span. To restore this performance improvement we can add netstandard2.1 as an additional target framework.
- Make the example projects and test projects net7.0 or net7.0-windows. Since the code in the main library isn't platform-specific there isn't a need to illustrate it or test it on multiple platforms.

See discussion in https://github.com/svg-net/SVG/pull/1037 .

Notes:
- NO_SDC was previously unused
- System DPI was previous used to form a default for PointsPerInch which made document generation system-dependent. This was replaced with a constant 72 which appears to be the most common.
- Code that conditioned on !USE_SOURCE_GENERATORS was removed since source generators are now always used, a significant benefit since this also removes bad code like including code using reflection.

To resolve:
- Re-add netstandard2.1 to get access to faster span parsers.
- A few other remaining `#if`s


mrbean-bremen:
> Make the library netstandard2.0 only. ... To restore this performance improvement we can add netstandard2.1 as an additional target framework.

Agreed. These should be sufficient.

> Make the example projects and test projects net7.0 or net7.0-windows. 

Agreed about the example projects, but I would retain tests for .NET framework (as it uses NetStandard 2.0). There may be more improvements like `span` that are not compatible with that, so we want to test both versions.

> NO_SDC was previously unused

It is not used _inside_ SVG, but outside to replace SDC with [Skia](https://github.com/wieslawsoltes/Svg.Skia). Removing it would break `Svg.Skia`. It is a bit unfortunate that this is not visible in SVG.net, maybe we should at least add some documentation and / or tests (CC @wieslawsoltes).

> System DPI was previous used to form a default for PointsPerInch which made document generation system-dependent. This was replaced with a constant 72 which appears to be the most common.

This was a fix that was requested - it had been 72 before, which led to problems with high-res displays. As high-res displays are quite common now, I wouldn't say that 72 is (or will be) the most common.

> Code that conditioned on !USE_SOURCE_GENERATORS was removed since source generators are now always used, a significant benefit since this also removes bad code like including code using reflection.

Agreed. This was added at a time where .NET 4.52 was still supported, that does not support source generators, so it can be removed now.

inforithmics:
Shouldn't the dpi be settable externally. Because when the svg is generated on a Server it would take the dpi from the server.

mrbean-bremen:
> Shouldn't the dpi be settable externally. Because when the svg is generated on a Server it would take the dpi from the server.

True, that would make sense. I think the default case should still be to use the current resolution, though.

inforithmics:
Yes, The logic shouldn't be changed in this Pull Request, It should be kept as is. Because if a different behavior is wanted, The value can already be set externally.

paulushub:
> Code that conditioned on !USE_SOURCE_GENERATORS was removed since source generators are now always used, 
> a significant benefit since this also removes bad code like including code using reflection.

@charlesroddie Please can you update your PR to include only the above?

charlesroddie:
The current code to get system dpi https://github.com/svg-net/SVG/blob/master/Source/SvgDocument.cs#L83 has the following issues:
- It only works on Windows and is incorrect elsewhere.
- Why a document would depend on the current system dpi is unexplained, and the system may not be windows, and may not have a gui.
- It introduces Windows dependencies in https://github.com/svg-net/SVG/blob/master/Source/SvgDocument.Win32.cs . The only purpose of this Windows dependency appears to be to get this dpi number that instead a user could pass in.

In order to make this library cross-platform surely this is worth breaking by requiring a consumer to pass in the dpi?

paulushub:
@charlesroddie  The acceptance of NO_SDC to me was a mistake. But I think it was done to accommodate `Svg.Skia`, which for some reasons never became part of this project.
My current plan is to clean up, clear the PRs, fix as many issues as possible and then start working on separating the rendering from the parsing of the SVG. Hope you will help with this efforts.

mrbean-bremen:
> The acceptance of NO_SDC to me was a mistake. But I think it was done to accommodate Svg.Skia, which for some reasons never became part of this project.

`NO_SDC` was a stop-gap measure to allow `Svg.Skia` to use `Svg` as a backend without the need for a fork - the long-term goal was the real separation of backend and frontend.

paulushub:
@charlesroddie Thanks for the understanding.
BTW, by your profile I guess you are doing a cross-platform. If so, can I ask for a another favor?
How is this library doing on MacOS (it only tested on Windows)?

With https://github.com/svg-net/SVG/pull/1116 merged, I will close this PR for now.
Please bring in the other issues separately for consideration.

charlesroddie:
> How is this library doing on MacOS (it only tested on Windows)?

We use this library via Svg.Skia and it works well across Windows (uwp), MacOS (net8.0-macos nativeaot), IOS (net8.0-ios nativeaot), Android (net8.0-android), and WASM (blazor).

The bits that are not cross-platform-friendly are not invoked in this path, although the compiler may not know that. It's best to have a guarantee of cross-platform and trimming-friendliness via removing platform-specific code from the parsing part of this library which I gather is your plan.

It does seem to result in a large deployment size and I noticed 1MB on WASM related to Svg.Skia and this must have something to do with trimming. So I think this will be greatly improved when the plan above is implemented.

PR #1071, opened 2023-08-26
Nuget dependency Cleanup

inforithmics:
don't refererence System.Memory and System.ObjectModel in Target Frameworks that don't need them

I saw that the nuget Package references System.Memory when it targets .NET 6, .NET Standard 2.1 this isn't necessary because the functionality is integrated in the .NET 6 Framework already.

only the Targets .NET Standard 2.0 and .NET 4.6.2 need it

I removed System.ObjectModel altogether because it is included in .NET Standard 2.0


inforithmics:
Seems I cannot change in this pull request the yml file.

mrbean-bremen (approved):
Looks good to me!

PR #1079, opened 2023-08-28
Reduce pressure on SOH in hot path

AntonyCorbett:
This line is on a hot path and string interopolation reduces the number of memory allocations for strings.


mrbean-bremen:
@wieslawsoltes - please have a look at these performance related PRs, as you already have experience there.

PR #1083, opened 2023-08-29
Update ex css to 4.2.1 (In preparation for using it to evaluate css selector)

inforithmics:
#### Reference Issue
https://github.com/svg-net/SVG/discussions/1073
Updates ExCss to the newest Version and fixes Unit Tests for that.

### DIFF
--- a/Source/SvgDocument.cs
+++ b/Source/SvgDocument.cs
@@ -444,7 +444,7 @@ namespace Svg
             if (styles.Any())
             {
                 var cssTotal = string.Join(Environment.NewLine, styles.Select(s => s.Content).ToArray());
-                var stylesheetParser = new StylesheetParser(true, true);
+                var stylesheetParser = new StylesheetParser(true, true, tolerateInvalidValues: true);
                 var stylesheet = stylesheetParser.Parse(cssTotal);
 
                 foreach (var rule in stylesheet.StyleRules)
--- a/Tests/Svg.UnitTests/EntitiesTests.cs
+++ b/Tests/Svg.UnitTests/EntitiesTests.cs
@@ -26,8 +26,8 @@ namespace Svg.UnitTests
 
             var svgPath = Path.Combine(TestsRootPath, EntitiesSampleSvgPath);
             var doc = SvgDocument.Open<SvgDocument>(svgPath, entities);
-            Assert.That(doc.Children[0].Children[0].Fill, Is.EqualTo(new SvgColourServer(Color.Red)));
-            Assert.That(doc.Children[0].Children[1].Fill, Is.EqualTo(new SvgColourServer(Color.Yellow)));
+            Assert.That(doc.Children[0].Children[0].Fill, Is.EqualTo(new SvgColourServer(Color.FromArgb(255, 0 ,0))));
+            Assert.That(doc.Children[0].Children[1].Fill, Is.EqualTo(new SvgColourServer(Color.FromArgb(255, 255 ,0))));
         }
 
         [Test]
--- a/Tests/Svg.UnitTests/LexerIssueTests.cs
+++ b/Tests/Svg.UnitTests/LexerIssueTests.cs
@@ -55,8 +55,8 @@ namespace Svg.UnitTests
             // valid colors
             var doc = GenerateLexerTestFile("fill: #ff0000; stroke: #ffff00");
             var path = doc.GetElementById<SvgPath>("path1");
-            Assert.AreEqual(System.Drawing.Color.Red, ((SvgColourServer)path.Fill).Colour);
-            Assert.AreEqual(System.Drawing.Color.Yellow, ((SvgColourServer)path.Stroke).Colour);
+            Assert.AreEqual(System.Drawing.Color.FromArgb(255,0,0), ((SvgColourServer)path.Fill).Colour);
+            Assert.AreEqual(System.Drawing.Color.FromArgb(255,255, 0), ((SvgColourServer)path.Stroke).Colour);
         }
 
         [Test]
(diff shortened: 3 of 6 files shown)


Source/SvgDocument.cs
```diff
@@ -444,7 +444,7 @@ private static T Create<T>(XmlReader reader) where T : SvgDocument, new()
             if (styles.Any())
             {
                 var cssTotal = string.Join(Environment.NewLine, styles.Select(s => s.Content).ToArray());
-                var stylesheetParser = new StylesheetParser(true, true);
+                var stylesheetParser = new StylesheetParser(true, true, tolerateInvalidValues: true);
```
inforithmics, line 447: TolarateInvalidValues is needed so that Invalid Property Values are still read. Could be Too that ExCss is not correctly annotated yet.

This fixed a Bug: But now we have to set tolarateInvalidValues to still get all values
https://github.com/TylerBrinks/ExCSS/commit/cea80ccf5c392e2bc5cd0e853c94167c811d315e

Tests/Svg.UnitTests/EntitiesTests.cs
```diff
@@ -26,8 +26,8 @@ public void TestOpenWithEntities()
 
             var svgPath = Path.Combine(TestsRootPath, EntitiesSampleSvgPath);
             var doc = SvgDocument.Open<SvgDocument>(svgPath, entities);
-            Assert.That(doc.Children[0].Children[0].Fill, Is.EqualTo(new SvgColourServer(Color.Red)));
-            Assert.That(doc.Children[0].Children[1].Fill, Is.EqualTo(new SvgColourServer(Color.Yellow)));
+            Assert.That(doc.Children[0].Children[0].Fill, Is.EqualTo(new SvgColourServer(Color.FromArgb(255, 0 ,0))));
```
inforithmics, line 29: Colors are now in ARGB values so I have to use this to Correctly Compare them

Tests/Svg.UnitTests/LexerIssueTests.cs
```diff
@@ -66,16 +66,16 @@ public void Lexer_FileWithInvalidHex_ColorTagOnlyReverstToDefaultForTheInvalidTa
             var doc = GenerateLexerTestFile("fill: #ff00; stroke: #00ff00");
             var path = doc.GetElementById<SvgPath>("path1");
             // default fill color is Black
-            Assert.AreEqual(System.Drawing.Color.Black, ((SvgColourServer)path.Fill).Colour);
-            Assert.AreEqual(System.Drawing.Color.Lime, ((SvgColourServer)path.Stroke).Colour);
+            Assert.AreEqual(System.Drawing.Color.FromArgb(0,255,255,0), ((SvgColourServer)path.Fill).Colour);
```
inforithmics, line 69: 4Hex values are now valid colors

https://github.com/TylerBrinks/ExCSS/commit/cb19e8f9d26860887142a1ad3ee75fbb93fbc599

inforithmics (commented):
I made a self review to explain the changes.

mrbean-bremen (approved):
Thanks! Can you also an entry in the [release notes](https://github.com/svg-net/SVG/blob/master/doc/ReleaseNotes.md)?

H1Gdev:
Input values for tests have changed.
Does this mean that the current input does not work ?

inforithmics:
I think only the representation has changed
Red == System.Drawing.Color.FromArgb(255,0,0) Because it is now parsed as a color before it was parsed as Any (which probably parsed it as string) to Red, Now it seems to parse it as color and setting the Colors Directly.

var doc = GenerateLexerTestFile("fill: #ff00; stroke: #00ff00");
Because it can now parse the Fill Color as Color.FromArgb(255,255,0)

H1Gdev:
I think this response will have a large impact on users.
In tests, the value obtained from the Fill property is changing.

By the way, if you update to 4.2.0, there is no need to change the test. Do I have to update now ?
I am currently in the process of considering the latest support.

mrbean-bremen:
> I think this change will have a large impact on users.

True, I haven't considered this. I think we still want to have this change, as it _is_ fixing stuff, but communicate the breaking change clearly to the user, and maybe think about a major release because of it (I'm unsure about this).

> By the way, if you update to 4.2.0, there is no need to change the test. Do I have to update now ?

It is a bit unfortunate that the [color fix](https://github.com/TylerBrinks/ExCSS/pull/156) in ExCSS was made in a minor patch release and went a bit unnoticed, but as I said - it fixes color handling, so this is something we want to have.

One possibility is to update to 4.2.0, make a patch release, and later update to 4.2.1 in preparation for a major (or minor?) release.

H1Gdev:
I think purpose is to remove `Fizzler` dependency from `SVG.NET`.
(I agree that there will be less dependence.)

I think v4.2.0 can also achieve this purpose.

If we use 4.2.1 as in this PR, I think you need to review handling of colors in SVG.NET (file output, etc.).

inforithmics:
@H1Gdev updating to v4.2.0 as an intermediatary step would work, but finally I need to a yet to be realeased Version Greater than 4.2.1 Because of this.

https://github.com/TylerBrinks/ExCSS/pull/160

Because without that I cannot evalute the Selector. (Theoretically I could I would to Parse the ComplexSelector.

inforithmics:
I think the issue that's need investigating is.

var doc = GenerateLexerTestFile("fill: #ff00; stroke: #00ff00");

The other color parsing changes simply Set the RGB Colors instead of the KnownColor name
ARBG(0,255,0,0) instead of red, which shouldn't cause problems.

I think I will debug through what happens here.
var doc = GenerateLexerTestFile("fill: #ff00; stroke: #00ff00");

inforithmics:
I debug it through and there are really 4 hex Css Colors (8Bit per color) so 4 Hex values is ARGB 8Bit Color, So the behavior is correct.

https://www.quackit.com/css/color/values/css_hex_color_notation_4_digits.cfm

inforithmics:
Closing this in favor of this
https://github.com/svg-net/SVG/pull/1091

PR #1087, opened 2023-09-02
Add Svg.Custom Project so that NO_SDC PreCompiler Constants is recogn…

inforithmics:
…ized as used

#### Reference Issue
Because here the NO_SDC Compiler directive was removed because it was not seen as used.
https://github.com/svg-net/SVG/pull/1066

Adds the Svg.Custom Project from Skia.Svg so that the NO_SDC Compiler directive does not regress.

### DIFF
--- a/Source/Svg.sln
+++ b/Source/Svg.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.8.34004.107
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29215.179
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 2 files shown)


Source/Svg.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29215.179
+# Visual Studio Version 17
```
mrbean-bremen, line 3: Not sure if we shall change the version - it means it won't be opened with versions < 2022.

mrbean-bremen (commented):
Good idea! @wieslawsoltes - can you have a look please?

Source/Svg.sln
inforithmics: I'll revert that

mrbean-bremen (approved):

PR #1091, opened 2023-09-06
Update ExCSS to 4.2.2.

H1Gdev:

- Update ExCSS to 4.2.2.
- Corresponds to TylerBrinks/ExCSS#152.

### DIFF
--- a/Source/SvgElementFactory.cs
+++ b/Source/SvgElementFactory.cs
@@ -52,7 +52,7 @@ namespace Svg
         private static readonly Dictionary<string, ElementInfo> availableElementsWithoutSvg;
         private static readonly List<ElementInfo> availableElements;
 #endif
-        private readonly StylesheetParser stylesheetParser = new StylesheetParser(true, true);
+        private readonly StylesheetParser stylesheetParser = new StylesheetParser(true, true, tolerateInvalidValues: true);
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


H1Gdev:
@mrbean-bremen 

https://github.com/svg-net/SVG/pull/1083#issuecomment-1699983474

Update ExCSS to latest without impacting users.

mrbean-bremen:
@inforithmics - please have a look

Source/SvgElementFactory.cs
```diff
@@ -186,7 +186,7 @@ private void SetAttributes(SvgElement element, XmlReader reader, SvgDocument doc
                         var inlineSheet = stylesheetParser.Parse("#a{" + reader.Value + "}");
                         foreach (var rule in inlineSheet.StyleRules)
                             foreach (var declaration in rule.Style)
-                                element.AddStyle(declaration.Name, declaration.Value, SvgElement.StyleSpecificity_InlineStyle);
+                                element.AddStyle(declaration.Name, declaration.Original, SvgElement.StyleSpecificity_InlineStyle);
```
inforithmics, line 189: Yes this is the solution, so that the values don't change because before the update .Value was the same as .Original

inforithmics (approved):
This is the better solution for updating the ExCSS Library less changes

PR #1094, opened 2023-10-03
The SvgDocument.Open method got an additional parameter for external CSS

RomanKalachik:
… CSS

SvgDocument has CSS support which is cool. But sometimes you may want to specify an external CSS for an SVG document. This MR adds one more parameter to the SvgDocument.Open method for external CSS.


mrbean-bremen (approved):
Looks good to me - @H1Gdev, please have another look!

PR #1106, opened 2023-12-13
Fix declaring node also allows to use prefix for namespace

Fruchtzwerg94:
#### Issue / problem
Writing a SVG file where the namespace of a node is declared as an attribute in this node fails to resolve the namespace prefix.

As a minimal example, assume we want to add some additional meta data to a SVG file:

```
SvgDocument svg = SvgDocument.Open("MyImageWithMetaData.svg");

const string NAMESPACE_RDF = "http://www.w3.org/1999/02/22-rdf-syntax-ns#";
const string NAMESPACE_DC = "http://purl.org/dc/elements/1.1/";

SvgUnknownElement metadata = new SvgUnknownElement("metadata");
NonSvgElement rdfMetadata = new NonSvgElement("RDF", NAMESPACE_RDF);
rdfMetadata.Namespaces["rdf"] = NAMESPACE_RDF;

NonSvgElement rdfMetadataDescription = new NonSvgElement("Description", NAMESPACE_RDF);
rdfMetadataDescription.Children.Add(new NonSvgElement("creator", NAMESPACE_DC) { Content = "Hello World" });
rdfMetadata.Children.Add(rdfMetadataDescription);
metadata.Children.Add(rdfMetadata);
svg.Children.Add(metadata);

svg.Write("MyImageWithMetaData.svg");
```

The additional meta data will look like this:
```
<RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
  <rdf:Description />
</RDF>
```
Obviously the namespace here is mentioned twice which is not necessary since (see e.g. [Oracle](https://www.oracle.com/technical-resources/articles/srivastava-namespaces.html) or [w3resource](https://www.w3resource.com/index.php))

> The scope of a declared namespace begins at the element where it is declared and applies to the entire content of that element

which means namespace prefixes can be already used in the node they are declared.

```
<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
  <rdf:Description />
</rdf:RDF>
```

This PR also takes care of this and looks up the namespaces declared in the nodes themselfes.

### DIFF
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -516,7 +516,7 @@ namespace Svg
                     {
                         foreach (var kvp in this.Namespaces)
                         {
-                            if (kvp.Value == this.ElementNamespace && !string.IsNullOrEmpty(kvp.Key))
+                            if (kvp.Value.Equals(this.ElementNamespace) && !string.IsNullOrEmpty(kvp.Key))
                             {
                                 prefix = kvp.Key;
                                 break;
--- a/Source/SvgElement.cs
+++ b/Source/SvgElement.cs
@@ -512,6 +512,17 @@ namespace Svg
                 else
                 {
                     var prefix = writer.LookupPrefix(this.ElementNamespace);
+                    if (prefix == null && !this.ElementNamespace.Equals(SvgNamespaces.SvgNamespace))
+                    {
+                        foreach (var kvp in this.Namespaces)
+                        {
+                            if (kvp.Value == this.ElementNamespace && !string.IsNullOrEmpty(kvp.Key))
+                            {
+                                prefix = kvp.Key;
+                                break;
+                            }
+                        }
+                    }
                     if (prefix == null)
                         writer.WriteStartElement(this.ElementName, this.ElementNamespace);
                     else
--- a/doc/ReleaseNotes.md
+++ b/doc/ReleaseNotes.md
@@ -1,6 +1,11 @@
 # SVG.NET Release Notes
 The release versions are NuGet releases.
 
+## [Version X.X.X](https://www.nuget.org/packages/Svg/X.X.X)  (20XX-XX-XX)
+
+### Fixes
+* fixed XML namespace prefixes are also applied for nodes declaring them (see [PR #1106](https://github.com/svg-net/SVG/pull/1106))
+
 ## [Version 3.4.6](https://www.nuget.org/packages/Svg/3.4.6)  (2023-11-16)
 
 ### Fixes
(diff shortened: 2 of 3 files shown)


mrbean-bremen (commented):
Thanks, looks good! Please also add an entry in the [release notes](https://github.com/svg-net/SVG/blob/master/doc/ReleaseNotes.md), and if possible, a test.

Fruchtzwerg94:
> Thanks, looks good! Please also add an entry in the [release notes](https://github.com/svg-net/SVG/blob/master/doc/ReleaseNotes.md), and if possible, a test.

Test and changelog added @mrbean-bremen . Please keep in mind that the version and date in the changelog is just a placeholder _XX_ yet and needs to be set accordlingly at release.

Source/SvgElement.cs
```diff
@@ -512,6 +512,17 @@ protected virtual void WriteStartElement(XmlWriter writer)
                 else
                 {
                     var prefix = writer.LookupPrefix(this.ElementNamespace);
+                    if (prefix == null && !this.ElementNamespace.Equals(SvgNamespaces.SvgNamespace))
+                    {
+                        foreach (var kvp in this.Namespaces)
+                        {
+                            if (kvp.Value == this.ElementNamespace && !string.IsNullOrEmpty(kvp.Key))
```
H1Gdev, line 519:
```cs
if (!string.IsNullOrEmpty(kvp.Key) && kvp.Value == this.ElementNamespace)
```

Source/SvgElement.cs
```diff
@@ -512,6 +512,17 @@ protected virtual void WriteStartElement(XmlWriter writer)
                 else
                 {
                     var prefix = writer.LookupPrefix(this.ElementNamespace);
+                    if (prefix == null && !this.ElementNamespace.Equals(SvgNamespaces.SvgNamespace))
```
H1Gdev, line 515: Is there a reason it's `Equals` here and `==` on line 519 ?

doc/ReleaseNotes.md
```diff
@@ -1,6 +1,11 @@
 # SVG.NET Release Notes
 The release versions are NuGet releases.
 
+## [Version X.X.X](https://www.nuget.org/packages/Svg/X.X.X)  (20XX-XX-XX)
```
mrbean-bremen, line 4: Thanks - we usually just use `## Unreleased` here to avoid dead links.

Fruchtzwerg94: Updated

Source/SvgElement.cs
Fruchtzwerg94: I agree, one way to do it for both comparisons would be more clean. I've updated it in my latest commit. Good catch.

mrbean-bremen (approved):
Thanks!

PR #1107, opened 2023-12-13
Fixed Conflicts of Pull Request https://github.com/svg-net/SVG/pull/1094

inforithmics:
#### Reference Issue
Fixes the Conflicts of this pull request
https://github.com/svg-net/SVG/pull/1094

SvgDocument has CSS support which is cool. But sometimes you may want to specify an external CSS for an SVG document. This MR adds one more parameter to the SvgDocument.Open method for external CSS.

#### Any other comments?

### DIFF
--- a/Source/.cr/personal/FavoritesList/List.xml
+++ b/Source/.cr/personal/FavoritesList/List.xml
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Root Type="DevExpress.CodeRush.Foundation.CodePlaces.Options.FavoritesListContainer">
+  <Options Language="Neutral">
+    <Groups />
+  </Options>
+</Root>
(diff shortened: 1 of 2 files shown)


H1Gdev:
@inforithmics 
Please add release notes and tests.

Why do you only provide this feature for some `Open` methods?
For example, is it not needed [here](https://github.com/svg-net/SVG/blob/4cd055a23f1297edace3661692a993cd44ca7398/Source/SvgDocument.cs#L227) ?


In addition, I have a negative opinion about the way this functionality is provided from maintainability standpoint.
(It seems to be the same as [entities](https://github.com/svg-net/SVG/blob/4cd055a23f1297edace3661692a993cd44ca7398/Source/SvgDocument.cs#L285)...)

In my opinion, I think you should just add `CSS` to created `SvgDocument` instance.

inforithmics:
I'll make a follow up pull request.
I discovered another problem this change is not binary compatible. So I will create the overloaded functions to make it binary compatible.

PR #1110, opened 2023-12-17
Css parameter follow up

inforithmics:
#### Reference Issue
Follow Up of following Pull request
https://github.com/svg-net/SVG/pull/1107

Made the pull request binary compatible.
Added Release notes


inforithmics:
The question is if there should be made an Open overload with an options object like this.
`public static T Open<T>(string path, SvgOptions options) where T : SvgDocument, new()`
With this additional settings won't break the binary compatibility and wouldn't create more and more overloads.

mrbean-bremen:
@H1Gdev - please have a look!

H1Gdev:
As mentioned in #1107, I am negative to add more arguments to `Open` method.
This PR has also add a further variation `Open` method...(I think it will cause confusion to users.)

@inforithmics 
If this feature is needed, how about providing methods that can set external CSS ?

inforithmics:
I thought about it and I think it is the better solution providing an external function that applies a css. So I can apply an css without again loading the document I will change this pull request accordingly

paulushub:
> The question is if there should be made an Open overload with an options object like this. `public static T Open<T>(string path, SvgOptions options) where T : SvgDocument, new()` `public static T Open<T>(Stream stream, SvgOptions options) where T : SvgDocument, new()` With this additional settings won't break the binary compatibility and wouldn't create more and more overloads.

The `SvgOptions` (or `SvgSettings`) class will be a good idea. Most of the current `Open/Create` extra arguments could be added as fields.
Also, it can be added through an additional `SvgDocument` constructor. The default construction will create the default `SvgOptions` class with default settings.

paulushub:
The thought on options is to include properties
```cs
using System;
using System.Collections;
using System.Collections.Generic;
#if NETCORE
using System.Diagnostics.CodeAnalysis;
#endif

namespace Svg
{
    public class SvgOptions : IDictionary<string, string>, ICloneable
    {
        private readonly IDictionary<string, string> _properties;

        public SvgOptions()
        {
            _properties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
        }

        public string this[string key] 
        {
            get
            {
                if (_properties.TryGetValue(key, out string value))
                    return value;
                return string.Empty;
            }
            set
            {
                if (key != null)
                {
                    _properties[key] = value;
                }
            }
        }

        public ICollection<string> Keys => _properties.Keys;

        public ICollection<string> Values => _properties.Values;

        public int Count => _properties.Count;

        public bool IsReadOnly => _properties.IsReadOnly;

        public void Add(string key, string value)
        {
            if (key != null)
            {
                _properties.Add(key, value);
            }
        }

        public void Add(KeyValuePair<string, string> item)
        {
            this.Add(item.Key, item.Value);
        }

        public void Clear()
        {
            _properties.Clear();
        }

        public bool Contains(KeyValuePair<string, string> item)
        {
            return _properties.Contains(item);
        }

        public bool ContainsKey(string key)
        {
            return _properties.ContainsKey(key);
        }

        public void CopyTo(KeyValuePair<string, string>[] array, int arrayIndex)
        {
            _properties.CopyTo(array, arrayIndex);
        }

        public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
        {
            return _properties.GetEnumerator();
        }

        public bool Remove(string key)
        {
            return _properties.Remove(key);
        }

        public bool Remove(KeyValuePair<string, string> item)
        {
            return _properties.Remove(item);
        }

#if NETCORE
        public bool TryGetValue(string key, [MaybeNullWhen(false)] out string value)
        {
            return _properties.TryGetValue(key, out value);
        }
#else
        public bool TryGetValue(string key, out string value)
        {
            return _properties.TryGetValue(key, out value);
        }
#endif

        IEnumerator IEnumerable.GetEnumerator()
        {
            return ((IEnumerable)_properties).GetEnumerator();
        }

        object ICloneable.Clone()
        {
            throw new NotImplementedException();
        }

        public SvgOptions Clone()
        {
            SvgOptions options = new SvgOptions();
            foreach (KeyValuePair<string, string> item in _properties)
            {
                options.Add(item);
            }

            return options;
        }
    }
}
```

mrbean-bremen (approved):

paulushub:
@inforithmics For new public methods/properties please use interfaces `IDictionary<T, T>` instead of the class `Dictionary<T, T>` where possible.

Issue #900, opened 2021-09-28
Private font incorrectly displays italic style


huigy:
#### Description
I am using SVG.NET to convert SVG to PNG images, while some parts of the text need to use our private font file:  Jyeoohai.ttf  and jyeoohai-italic.ttf.
I use svgfontManager.privateFontPathList.add (fonts) to load the above two font files, the resulting image text of normal form is correct, but the italic text is invalid.
If I install the font file on Windows, the font styles are correct even without calling svgfontManager.privateFontPathList.add.  However, software is going to be deployed to dozens of servers , so  I don't want to install fonts on each one.  Can SVG.NET dynamically load these character files?

#### Example data
```
<svg xmlns="http://www.w3.org/2000/svg">
  <text y="20" x="5" font-size="32px"
      font-style="normal" style="font-family: Jyeoohai">
    Normal font style
  </text>

  <text y="55" x="5" font-size="32px"
      font-style="italic" style="font-family: Jyeoohai">
    Italic font style
  </text>
</svg>
```

#### Used Versions
SVG version 3.3.0.0

mrbean-bremen:
Can you show what do you mean by "the italic text is invalid"? I guess it is not possible to upload your private font...

H1Gdev:
@huigy 

I used [Open Sans](https://fonts.google.com/specimen/Open+Sans) to check how [PrivateFontCollection](https://docs.microsoft.com/dotnet/api/system.drawing.text.privatefontcollection) behaves.

If font name is the same, font added earlier will be applied.
1. OpenSans-Regular.ttf <- Applies as `Open Sans`
2. OpenSans-Italic.ttf

I have never created font files, is it possible to combine all styles into one file ?

PR #1118, opened 2023-12-30
Remove net462 TargetFramework as net462 is supported by netstandard2.0

charlesroddie:
The net462 TargetFramework in svg.csproj is unnecessary as net462 is supported by netstandard.

This allows cleaning up some code using `#if` annotations for `Net462` and `NETFULL`.

Some code is removed which only works on `Net462`: `RegisterEvents` involving mice, and `Web/SvgHandler.cs`. It is assumed that these are legacy methods which are no longer needed as they work only on Net462.

Tests and samples involving net462 are unchanged


paulushub:
@charlesroddie Thanks for the contribution. Please tiny up but keep the support for .NET 4.6.2. 
It is a multiple framework support and having .NET 4.6.2 does not affect anyone.
See the packaging of [SkiSharp](https://www.nuget.org/packages/SkiaSharp/) by Microsoft, it supports
- .NET Core 3.1
- .NET Framework 4.6.2
- .NET Standard 1.3
- .NET Standard 2.0
- .NET Standard 2.1
- .NET 6.0
- etc.

Libraries like `Svg.Skia`, which depends on `SkiaSharp` can actually also support these frameworks, if they wish.
Industrial platform like MATLAB support .NET Framework Version 4.0 run time and higher, all the samples are for .NET 4.0.

charlesroddie:
> @charlesroddie Thanks for the contribution. Please tiny up but keep the support for .NET 4.6.2. It is a multiple framework support and having .NET 4.6.2 does not affect anyone.

This PR doesn't remove support for .NET 4.6.2 since it retains netstandard2.0 which supports .Net 4.6.2.

> Libraries like `Svg.Skia`, which depends on `SkiaSharp` can actually also support these frameworks, if they wish. Industrial platform like MATLAB support .NET Framework Version 4.0 run time and higher, all the samples are for .NET 4.0.

Svg.Skia targets `netstandard2.0`, `net6.0`, and `net8.0` (https://github.com/wieslawsoltes/Svg.Skia/blob/master/src/Svg.Skia/Svg.Skia.csproj#L5) and supports .Net 4.6.2 via `netstandard2.0`. This is a reasonable combination.

> See the packaging of [SkiSharp](https://www.nuget.org/packages/SkiaSharp/) by Microsoft

I am not competent to understand how SkiaSharp uses the dotnet project system to deploy to all platforms. Presumably it has explicit framework support because of all the platform-specific native bindings it needs to support?

mrbean-bremen:
Looks like the reason for the failing tests is missing support for private fonts. From the test code:
```cs
#if NETSTANDARD
        // Private font does not work if .NET Standard.
        protected override int ExpectedSize { get { return 3000; } } // 3155
#else
        protected override int ExpectedSize { get { return 3200; } } // 3512
#endif
```
Not sure why `NETSTANDARD` is not defined in these tests. I also don't know how important it is to have private font support for 4.6.2.

paulushub:
> This PR doesn't remove support for .NET 4.6.2 since it retains netstandard2.0 which supports .Net 4.6.2.

It works but packaging `netstandard2.0` assemblies with final build application is not what everybody want to do.

> I am not competent to understand how SkiaSharp uses the dotnet project system to deploy to all platforms. 

Easy way, compare the output directory (`bin`) of each SVG-NET platform build. The presence of the .net standard 2.0 assemblies in the .net462 directory is due to `ExCSS` excluding .net462 support.

paulushub:
> I also don't know how important it is to have private font support for 4.6.2.

In SVG, I do not think it is possible to avoid private fonts, since loading any font file not installed on the system is private.
We are actually going to do more going forward to resolve some of the existing issues like #900.

paulushub:
@charlesroddie Thanks for the contribution. For now, we will keep net462 TargetFramework target. 
Any removal will be considered after we complete the cleanup and resolve at least some of the pending issues.

PR #1019, opened 2022-10-19
Added net6.0-windows to TFMs.

AraHaan:
Due to the fact that in .NET 6 S.D.C is Windows only, I decided to make a Windows only target for .NET 6 which removes a few annoying package refs that get placed into my build outputs in my .NET 6 WinForms applications that use Svg. If the application is windows only, I see no need to include the ``osx`` System.Drawing runtime reference either.

#### Reference Issue

Contributes to: https://github.com/svg-net/SVG/issues/939

#### What does this implement/fix? Explain your changes.

This adds a ``net6.0-windows`` target to Svg which allows it to be used in .NET 6 WindowsDesktop applications without annoyingly unneeded dependencies polluting the output directory in them.

#### Any other comments?

None.

PR #1121, opened 2024-01-08
fixed CS8352 in net8

RomanKalachik:
if you try to compile this project in net8, it will show some 

SvgPointCollection.cs(66,89,66,94): error CS8352: Cannot use variable 'state' in this context because it may expose referenced variables outside of their declaration scope

This MR fixes that errors.

### DIFF
--- a/Source/DataTypes/SvgPointCollection.cs
+++ b/Source/DataTypes/SvgPointCollection.cs
@@ -61,7 +61,7 @@ namespace Svg
             if (value is string s)
             {
                 var coords = s.AsSpan().Trim();
-                var state = new CoordinateParserState(ref coords);
+                var state = new CoordinateParserState(coords);
                 var result = new SvgPointCollection();
                 while (CoordinateParser.TryGetFloat(out var pointValue, ref coords, ref state))
                 {
--- a/Source/Paths/CoordinateParser.cs
+++ b/Source/Paths/CoordinateParser.cs
@@ -24,7 +24,7 @@ namespace Svg
         public int Position;
         public bool HasMore;
 
-        public CoordinateParserState(ref ReadOnlySpan<char> chars)
+        public CoordinateParserState(ReadOnlySpan<char> chars)
         {
             CurrNumState = NumState.Separator;
             NewNumState = NumState.Separator;
--- a/Source/Paths/SvgPathBuilder.cs
+++ b/Source/Paths/SvgPathBuilder.cs
@@ -50,14 +50,14 @@ namespace Svg
                         if (command.Length > 0)
                         {
                             var commandSetTrimmed = pathTrimmed.Slice(start, length).Trim();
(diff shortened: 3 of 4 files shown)


paulushub:
@RomanKalachik Thanks for the contribution. Still do not understand why `ref` was needed, but will look into it.
How about just adding the .NET 8 support as well? (was hoping #1019 will add it but still no progress there).

RomanKalachik:
> @RomanKalachik Thanks for the contribution. Still do not understand why `ref` was needed, but will look into it. How about just adding the .NET 8 support as well? (was hoping #1019 will add it but still no progress there).

Net8.0 support in this whole project is a long story because of System.Drawing. 
I believe that `ref` is not required in that case and removed it in this MR

paulushub:
> Net8.0 support in this whole project is a long story because of System.Drawing.

Not tried it yet, but was thinking it should succeed with `System.Drawing.Primitives`, which is available as Nuget package.

paulushub:
@RomanKalachik Please can you share the issues you have with the .NET 8 support?
Currently, the support for `System.Drawing` is fixed at version `5.0.3` in the SVG-NET. 
I downloaded your project, added the .NET 8, it builds without any issue.
image

I tested all on command-lines:
```bash
dotnet test --framework net8.0 Tests/Svg.UnitTests/Svg.UnitTests.csproj
dotnet test --framework net6.0 Tests/Svg.UnitTests/Svg.UnitTests.csproj
dotnet test --framework netcoreapp3.1 Tests/Svg.UnitTests/Svg.UnitTests.csproj
dotnet test --framework net462 Tests/Svg.UnitTests/Svg.UnitTests.csproj
```
All tests succeeded except 1, and it is due to difference in line ending. Still do not know why this fails with your source code (I downloaded it, not a git clone).
Untitled1

paulushub:
Why does your PR includes cases as shown below for the `ReadOnlySpan<char>` (the `CoordinateParserState` is modified so `ref` is expected).
#### Svg.Helpers.StringParser
```cs
	public static float ToFloat(ref ReadOnlySpan<char> value)
	{
#if NETSTANDARD2_1 || NETCOREAPP2_1_OR_GREATER
		return float.Parse(value, NumberStyles.Float, Format);
#else
		return float.Parse(value.ToString(), NumberStyles.Float, Format);
#endif
	}
```

#### Svg.CoordinateParser
```cs
public static bool TryGetFloat(out float result, ref ReadOnlySpan<char> chars, ref CoordinateParserState state)
{
	var charsLength = chars.Length;

	while (state.CharsPosition < charsLength && state.HasMore)
	{
		var currentChar = chars[state.CharsPosition];

		// NOTE: Deleted for simplicity
	}
}
```

mrbean-bremen (commented):
Can you please add an entry in the release notes?

paulushub:
Issues raised here are fixed in #1137

PR #1141, opened 2024-02-03
Refactoring of the Source Code Generator

paulushub:
### Description
- Moved the `Svg_Model.cs` file as `SvgModel.cs` into the main source.
- Restore the standard naming of the generated code files with `*.g.cs` suffix.
- Set the output directory of the generated codes to `Generated`.
- The generated source codes can now be viewed when you turn on `Show All Files`.
- Added the Generated directory to the main source codes.
- Moved the `Svg.Custom` project to the `Tests` directory.

### Type of change
- [x]  Refactoring of generator code, no change in main source codes (non-breaking change which fixes an issue)
- [x] This change does not require a documentation update
- [x] `SvgModel.cs`, originally part of the code generator, includes attribute classes.
  - The attributes are "fixed" internal classes, does not require reflection and there is no need to generate it.
  - The use of generators to create the attributes affects the use of latest `docfx` for documentation.
  - Reported `docfx` issue with attributes: https://github.com/dotnet/docfx/issues/9619
  - `docfx` currently provides no solution other than a warning.
- [ ] Resolved review issues.

### How Has This Been Tested?
- [x] SvgW3CTestRunner is tested for each supported platform
- [x] Svg.UnitTests tested on command line for each supported platform
- [x] Svg.Benchmarks tested on command line for each supported platform.


Generators/AvailableElementsGenerator.cs
```diff
@@ -109,17 +36,19 @@ public void Initialize(GeneratorInitializationContext context)
...
             // Check is we have our SyntaxReceiver object used to filter compiled code.
             if (!(context.SyntaxReceiver is SyntaxReceiver receiver))
             {
                 return;
             }
 
-            var options = (context.Compilation as CSharpCompilation)?.SyntaxTrees[0].Options as CSharpParseOptions;
-            var compilation = context.Compilation.AddSyntaxTrees(CSharpSyntaxTree.ParseText(SourceText.From(ModelText, Encoding.UTF8), options));
+            //var options = (context.Compilation as CSharpCompilation)?.SyntaxTrees[0].Options as CSharpParseOptions;
+            //var compilation = context.Compilation.AddSyntaxTrees(CSharpSyntaxTree.ParseText(
```
mrbean-bremen, line 49: just remove the unneeded code

Generators/AvailableElementsGenerator.cs
```diff
@@ -109,17 +36,19 @@ public void Initialize(GeneratorInitializationContext context)
         /// <inheritdoc/>
         public void Execute(GeneratorExecutionContext context)
         {
-            // Add the ElementFactory model source to compilation. 
-            context.AddSource("Svg_Model", SourceText.From(ModelText, Encoding.UTF8));
+            //// Add the ElementFactory model source to compilation. 
+            //context.AddSource("Svg.Model.g.cs", SourceText.From(SourceTexts.ModelText, Encoding.UTF8));
```
mrbean-bremen, line 40: Is this not needed?

paulushub: Moved to `Source/Svg.Model.cs`. Initiallly moved the class to template `SourceTexts`, before deleting it from the Generator.

mrbean-bremen: Ah. missed this.

mrbean-bremen (approved):

PR #1184, opened 2025-07-21
Add IsTrimmanle/IsAotCompatible attributes

maxkatz6:

This PR fixes several warnigns/erros that are reproducible during PublishAOT build with this library referenced:
```
    /_/externals/SVG/Source/DataTypes/SvgAspectRatio.cs(59): Trim analysis error IL2026: Svg.SvgAspectRatio.ToString(): Using member 'System.ComponentModel.TypeDescriptor.GetConverter(Type)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.
    /_/externals/SVG/Source/DataTypes/SvgAspectRatio.cs(59): AOT analysis error IL3050: Svg.SvgAspectRatio.ToString(): Using member 'System.Enum.GetValues(Type)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. It might not be possible to create an array of the enum type at runtime. Use the GetValues<TEnum> overload or the GetValuesAsUnderlyingType method instead.
```

These special attributes fix the problem by ensuring attributes are statically linked, and not removed from the assembly. Without changing any logic.

#### Any other comments?

It would be possible to avoid DynamicDependency attributes, by refactoring code to not depend on TypeDescriptors, but that would be much more involving change, and in my opinion unnecessary.

### DIFF
--- a/Docs/articles/ReleaseNotes.md
+++ b/Docs/articles/ReleaseNotes.md
@@ -5,6 +5,7 @@ The release versions are NuGet releases.
 
 ### Enhancements
 * Added support for `Switch` element with "systemLanguage" selector attribute (see [#1176](https://github.com/svg-net/SVG/issues/1176)).
+* Enabled `IsTrimmable` and `IsAotCompatible` attributes on the assembly (see [#1184](https://github.com/svg-net/SVG/pull/1184)).
 
 ### Fixes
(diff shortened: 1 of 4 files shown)


mrbean-bremen (commented):
Thanks! Can you please add an entry to the [release notes](https://github.com/svg-net/SVG/blob/master/Docs/articles/ReleaseNotes.md)?

maxkatz6:
@mrbean-bremen done

PR #1199, opened 2025-12-18
Add support for dominant-baseline

gdube-nrc:
- Add support for parsing `dominant-baseline`.

#### Any other comments?
Used mdn for reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/dominant-baseline


gdube-nrc:
Okay, I'll check it out

PR #1201, opened 2026-02-05
SvgElementCollection tests added; Insert now handles case when index==Count

### LINKED ISSUE #1200: SvgElement.Children.Insert fails to add an element to the end
`SvgElementCollection.Insert(index, element)` throws an index-out-of-range exception when `index==Children.Count`. Of course, `Add(element)` can be used in this case, but it would be cleaner if `Insert `worked according to the `IList<>` spec, as `SvgElementCollection` does implement that interface.

See the Remarks in the [IList documentation](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.system-collections-ilist-insert?view=net-9.0):

> If index equals the number of items in the [IList](https://learn.microsoft.com/en-us/dotnet/api/system.collections.ilist?view=net-9.0), then item is appended to the end.

jlidbeck:
Fixes #1200: SvgElementCollection.Insert throws exception when index==count.

According to [IList documentation](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.system-collections-ilist-insert?view=net-9.0), in this case the expected behavior is that the element is added to the end of the list. A minor fix was made to the SvgElementCollection internals to match this behavior.

Unit tests were added to verify insert/remove behavior.

### DIFF
--- a/Docs/articles/ReleaseNotes.md
+++ b/Docs/articles/ReleaseNotes.md
@@ -13,6 +13,7 @@ The release versions are NuGet releases.
 
 ### Fixes
 * Fixed: `<CompilerGeneratedFilesOutputPath>` in `Svg.Custom` to generate `Generated` directory under project. (see [PR #1153](https://github.com/svg-net/SVG/pull/1153))
+* Fixed `SvgElementCollection.Insert()` to allow child elements to be added to end of list (see [PR #1200](https://github.com/svg-net/SVG/issues/1101))
 
 ## [Version 3.4.7](https://www.nuget.org/packages/Svg/3.4.7)  (2024-02-22)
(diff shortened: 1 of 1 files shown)


mrbean-bremen (commented):
Thanks! Can you please add an entry to the [release notes](https://github.com/svg-net/SVG/blob/master/Docs/articles/ReleaseNotes.md)?

znakeeye:
`Insert(0, item)` should **not** throw if collection is empty. Version 3.4.7 suffers from this bug.

```
var document = new SvgDocument();
document.Children.Insert(0, new SvgGroup()); // Throws :(
```

PR #1211, opened 2026-05-12
Remove DynamicallyAccessedMemberTypes to improve trim/aot compatibility

charlesroddie:
Remove DynamicallyAccessedMemberTypes to improve trim/aot compatibility

These attributes were added as part of trim annotations in https://github.com/svg-net/SVG/pull/1184 

This PR simplifies the code to remove the need for these attrbutes.

@maxkatz6 am I missing something here? You mentioned in the above PR that "It would be possible to avoid DynamicDependency attributes, by refactoring code to not depend on TypeDescriptors" but the attributes were only on two little types and the fixes seem very straightforward.

### DIFF
--- a/Source/DataTypes/SvgAspectRatio.cs
+++ b/Source/DataTypes/SvgAspectRatio.cs
@@ -1,6 +1,5 @@
 using System;
 using System.ComponentModel;
-using System.Diagnostics.CodeAnalysis;
 using Svg.DataTypes;
 
 namespace Svg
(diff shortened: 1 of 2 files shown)


maxkatz6:
@charlesroddie yes, this PR seems to be correct. You essentially removed dependency on type descriptors. It was much simpler than I previously thought.

mrbean-bremen (approved):
Thanks!

charlesroddie:
@wieslawsoltes I think this may allow getting rid of this attribute in Svg.Skia:
```
Using member 'Svg.Skia.SKSvg.FromSvg(String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. SVG document parsing may use trim-unsafe runtime discovery paths.
```

